PR #1, opened 2026-03-31
Add test-conventions skill

jcdendrite:
## Summary
- Adds a new auto-triggered Claude Code skill (`user-invocable: false`) that provides language-agnostic testing conventions and principles
- Triggers when discussing test strategy, planning tests, reviewing coverage, or writing test infrastructure
- Covers: test pyramid (unit/integration/contract/smoke), design for testability, test isolation (DB, parallel execution), naming conventions, test data management, coverage judgment, mock design, flaky test diagnosis, and 16 anti-patterns
- Reviewed by Staff Backend Engineer and Senior SDET agents — all HIGH and MEDIUM findings addressed

## Key sections
1. **Test pyramid** — includes contract testing layer, unit test speed thresholds, and "signs of an inverted pyramid" diagnostics
2. **Design for testability** — DI, language-aware error handling advice, test double taxonomy (stub/mock/fake/spy)
3. **Test isolation** — global state, DB transaction rollback, parallel-safe test guidance
4. **Naming and structure** — scenario+outcome naming, Arrange/Act/Assert, regression intent comments
5. **Test data** — factory/builder patterns, magic value avoidance, shared DB collision prevention
6. **Coverage judgment** — split HTTP-level vs service-level integration tests, concurrency coverage, security allow+deny paths
7. **Mock design** — fidelity, mutation recording, tautological mock detection with concrete example
8. **Flaky tests** — expanded root-cause enumeration (TZ, locale, floats, async, iteration order), in-test vs CI-level retry distinction
9. **Anti-patterns** — 16 entries including assertion-free tests, sleep-based sync, over-mocking, test interdependence

## Test plan
- [ ] Verify skill triggers correctly in a test strategy conversation
- [ ] Verify skill does NOT trigger when a project-level test skill is loaded
- [ ] Verify skill does NOT trigger for mechanical tasks (running tests, fixing assertions)
- [ ] Review advice accuracy across Python, Go, TypeScript, and Java contexts
- [ ] Confirm 225 lines is within acceptable context budget (~3-5K tokens on trigger)

### DIFF
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -6,7 +6,6 @@
         "hooks": [
           {
             "type": "command",
-            "if": "Bash(git commit*)",
             "command": "~/.claude/hooks/require-code-review.sh",
             "statusMessage": "Checking code review..."
           }
@@ -18,5 +17,8 @@
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
   },
-  "effortLevel": "high"
+  "effortLevel": "high",
+  "enabledPlugins": {
+    "atlassian@claude-plugins-official": true
+  }
 }
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -1,4 +1,5 @@
 {
+  "model": "Opus",
   "hooks": {
     "PreToolUse": [
       {
@@ -17,8 +18,5 @@
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
   },
-  "effortLevel": "high",
-  "enabledPlugins": {
-    "atlassian@claude-plugins-official": true
-  }
+  "effortLevel": "high"
 }
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -1,5 +1,4 @@
 {
-  "model": "Opus",
   "hooks": {
     "PreToolUse": [
       {
@@ -17,6 +16,5 @@
   "statusLine": {
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
-  },
-  "effortLevel": "high"
+  }
 }
--- a/claude/.claude/skills/test-conventions/SKILL.md
+++ b/claude/.claude/skills/test-conventions/SKILL.md
@@ -100,7 +100,7 @@ When tests run in parallel (pytest-xdist, Jest workers, Go's `t.Parallel()`, JUn
 
 Test names should describe the **scenario** and **expected outcome**, not just the function name:
 - Good: `rejects_expired_token_with_401`, `returns_empty_list_when_no_matches`
-- Bad: `test_refreshToken`, `test_search`, `it_works`
+- Bad: `test_refreshToken`, `test_search`, `it_works` (language-required prefixes like `test_` or `Test` are fine — the problem is having no scenario or expected outcome after the prefix)
 
 A reader should understand what the test verifies without reading its body.
 
--- a/claude/.claude/skills/test-conventions/SKILL.md
+++ b/claude/.claude/skills/test-conventions/SKILL.md
@@ -25,8 +25,8 @@ user-invocable: false
 | **Smoke** (post-deploy health) | Seconds | Production/staging infra | Critical paths are up after deployment | 1-2 per service, run on deploy only |
 
 **Contract tests** verify that service-to-service API schemas stay compatible without requiring a running instance of the other service. Use them at any service boundary where teams deploy independently.
-
-**E2E tests** verify complete user journeys across the full stack. **Smoke tests** are lightweight post-deploy checks that verify critical paths are functional — they answer "is it up?" not "does every flow work?"
+**E2E tests** verify complete user journeys across the full stack. 
+**Smoke tests** are lightweight post-deploy checks that verify critical paths are functional — they answer "is it up?" not "does every flow work?"
 
 ## 2. Design for testability
 
@@ -76,7 +76,7 @@ When replacing globals (HTTP client, fetch function, clock), always restore the 
 
 ### Parallel-safe tests
 Whether running locally or in CI, parallel test runners (pytest-xdist, Jest workers, Go's `t.Parallel()`, JUnit parallel mode) introduce isolation requirements:
-- Never bind to hardcoded ports; use port 0 or dynamic allocation
+- In tests that may run concurrently, never bind to hardcoded ports; use port 0 or dynamic allocation
 - Use unique temp directories per test (e.g., `mkdtemp`, test-scoped `tmp_path`)
 - When sharing a test database, use per-test schemas or transaction rollback isolation
 - If a test mutates module-level state, it cannot safely run in parallel — document this constraint explicitly
@@ -128,6 +128,7 @@ For tests that guard against a specific past bug, include a comment or docstring
 **Contract tests:**
 - Consumer expectations match the provider's actual API schema — the agreed-upon interface between services, independent of either side's implementation
 - Distinct from integration response shape tests: contract tests verify the *schema agreement* between services; integration tests verify your *code's actual output*
+- **When to use contract vs integration for response shape:** Use integration tests when you own both sides (or there's a single consumer) and can run the provider locally. Use contract tests when provider and consumer are owned by different teams, deploy independently, or when spinning up the provider in tests is expensive
 - Run when either side changes; no need for a live instance of the other service
 
 **E2E tests:**
--- a/claude/.claude/skills/test-conventions/SKILL.md
+++ b/claude/.claude/skills/test-conventions/SKILL.md
@@ -1,0 +1,225 @@
+---
+name: test-conventions
+description: >
+  Testing conventions and principles for any codebase: test pyramid, test isolation,
+  mock design, coverage judgment, and common anti-patterns.
+  TRIGGER when: discussing test strategy, planning tests for new code, reviewing
+  test coverage, when someone proposes a testing approach that seems wrong-layered,
+  or when writing test infrastructure (mocks, helpers, fixtures).
+  DO NOT TRIGGER when: a project-level test skill is already loaded, or the task
+  is purely mechanical (running tests, fixing a single assertion, updating a snapshot).
+user-invocable: false
+---
+
+# Testing Conventions
+
+## 1. Test pyramid
+
+| Layer | Speed | External deps | What it covers | Volume |
+|-------|-------|---------------|----------------|--------|
+| **Unit** (stubbed deps) | Milliseconds | None | Branch logic, error paths, edge cases, data transformations | Many — every code path |
+| **Integration** (real local services) | Hundreds of ms | Local DB, local services | Auth boundaries, response contracts, wiring between layers | Few — happy path + key error paths per endpoint or boundary |
+| **Contract** (schema verification) | Milliseconds–seconds | None (consumer) or local (provider) | API schemas between services match consumer expectations | One per service boundary |
+| **Smoke / E2E** (real external APIs) | Seconds | Third-party APIs, real infra | Full flow works end-to-end | 1-2 per feature, run on deploy not on every commit |
+
+**Contract tests** verify that service-to-service API schemas stay compatible without requiring a running instance of the other service. Use them at any service boundary where teams deploy independently.
+
+### Signs of an inverted pyramid
+- Most tests call real services (DB, APIs) to test internal branching logic
+- Tests are slow (>1s each) because of network round-trips
+- Individual unit tests taking >50-100ms, indicating a hidden real dependency or overly complex setup
+- External API rate limits constrain how often you can run the test suite
+- Tests are flaky because of network timeouts or service availability
+- You can't run the test suite offline or in CI without credentials
+- Adding a new code branch requires setting up complex test state in a real database
+
+### When to move a test down the pyramid
+
+**Integration → unit when:**
+- The test sets up complex DB state to exercise one `if` branch
+- The test hits an external API just to test error handling
+- The test is slow (>500ms) and tests pure logic, not wiring
+- You're hitting rate limits on external APIs
+
+**Smoke → integration when:**
+- The test calls a real third-party API to verify your code handles errors
+- The test is flaky because of network conditions
+- The test creates real resources (emails, contacts) as a side effect
+
+## 2. Design for testability
+
+### Dependency injection over internal construction
+Functions should accept their dependencies as parameters, not create them internally. A function that constructs its own client internally cannot be tested without hitting the real service. Accepting the client as a parameter lets callers pass a test double.
+
+### Make expected failure paths easy to assert on
+In languages with explicit error types (Go, Rust, functional patterns), returning errors as values simplifies test assertions. In exception-oriented languages (Python, Java, C#, Ruby), use the framework's exception-assertion utilities (e.g., `pytest.raises`, JUnit's `assertThrows`). The key principle: expected failure paths should be as easy to test as success paths, using whatever the language's idiomatic mechanism is.
+
+### Choose the right test double
+
+| Double | Behavior | Use when |
+|--------|----------|----------|
+| **Stub** | Returns canned data, no call tracking | Testing your code's behavior given specific inputs |
+| **Mock** | Records calls for assertion | Verifying your code called a dependency with the right arguments |
+| **Fake** | Lightweight real implementation (in-memory DB, local HTTP server) | Testing without real infrastructure; also useful in unit tests when stubs are too complex (e.g., an in-memory repository that maintains state across calls) |
+| **Spy** | Wraps the real implementation, records calls | You need real behavior but want to verify interaction |
+
+Use the narrowest double that covers the test's intent. Prefer stubs for unit tests of pure logic; use mocks only when verifying interaction is the point of the test.
+
+### Test double seams by dependency type
+- **Database calls:** Stub/fake the client object, or use transaction rollback isolation (see section 3)
+- **External HTTP APIs:** Intercept by URL pattern or use a fake HTTP server
+- **Env vars / config:** Set and restore in setup/teardown blocks
+- **Time:** Inject timestamps as parameters, or set up data relative to "now"
+
+## 3. Test isolation
+
+### Tests must be independent
+- Never rely on test execution order
+- Each test creates its own data and cleans up in teardown
+- Use unique identifiers (timestamps, counters) to prevent cross-test collision
+
+### Global state must be saved and restored
+When tests modify global state (env vars, global functions, singletons), always save the original and restore in a guaranteed cleanup block (teardown, `finally`, `defer`, etc.). **Never** unconditionally delete or overwrite global state — a test that removes a value without saving it first will break every subsequent test that needs it.
+
+### Test double cleanup must be guaranteed
+When replacing globals (HTTP client, fetch function, clock), always restore the original in a guaranteed cleanup block, even if the test fails or throws.
+
+### Database test isolation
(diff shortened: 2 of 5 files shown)


claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+**Contract tests** verify that service-to-service API schemas stay compatible without requiring a running instance of the other service. Use them at any service boundary where teams deploy independently.
+
+### Signs of an inverted pyramid
+- Most tests call real services (DB, APIs) to test internal branching logic
+- Tests are slow (>1s each) because of network round-trips
+- Individual unit tests taking >50-100ms, indicating a hidden real dependency or overly complex setup
+- External API rate limits constrain how often you can run the test suite
+- Tests are flaky because of network timeouts or service availability
+- You can't run the test suite offline or in CI without credentials
+- Adding a new code branch requires setting up complex test state in a real database
```
jcdendrite: This might be a useful skill when debugging why tests are slow or when evaluating a repo, but I don't think it's helpful to reference all the time when writing tests. Thoughts?

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+**Integration → unit when:**
+- The test sets up complex DB state to exercise one `if` branch
+- The test hits an external API just to test error handling
+- The test is slow (>500ms) and tests pure logic, not wiring
+- You're hitting rate limits on external APIs
+
+**Smoke → integration when:**
+- The test calls a real third-party API to verify your code handles errors
+- The test is flaky because of network conditions
+- The test creates real resources (emails, contacts) as a side effect
```
jcdendrite: This shouldn't be stated as an absolute, no-exceptions rule. This is more "when to consider moving a test down the pyramid".

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+| **Fake** | Lightweight real implementation (in-memory DB, local HTTP server) | Testing without real infrastructure; also useful in unit tests when stubs are too complex (e.g., an in-memory repository that maintains state across calls) |
+| **Spy** | Wraps the real implementation, records calls | You need real behavior but want to verify interaction |
+
+Use the narrowest double that covers the test's intent. Prefer stubs for unit tests of pure logic; use mocks only when verifying interaction is the point of the test.
+
+### Test double seams by dependency type
+- **Database calls:** Stub/fake the client object, or use transaction rollback isolation (see section 3)
+- **External HTTP APIs:** Intercept by URL pattern or use a fake HTTP server
+- **Env vars / config:** Set and restore in setup/teardown blocks
+- **Time:** Inject timestamps as parameters, or set up data relative to "now"
```
jcdendrite: Using data relative to "now" creates flaky tests. I say avoid unless absolutely necessary.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+
+## 3. Test isolation
+
+### Tests must be independent
+- Never rely on test execution order
+- Each test creates its own data and cleans up in teardown
+- Use unique identifiers (timestamps, counters) to prevent cross-test collision
+
+### Global state must be saved and restored
```
jcdendrite: And avoid tests that require global state if there's an alternative that doesn't require it.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+When replacing globals (HTTP client, fetch function, clock), always restore the original in a guaranteed cleanup block, even if the test fails or throws.
+
+### Database test isolation
+- **Transaction rollback pattern:** Wrap each test in a transaction and roll back at the end. Standard in Django, Rails, Spring, and most ORMs — more efficient than truncation.
+- **Test database lifecycle:** Use a dedicated test database, run migrations before the suite, reset state between tests.
+- **In-memory vs. real engine:** In-memory databases (e.g., SQLite in-memory mode) are fast but have dialect differences (JSON columns, CTEs, locking behavior). When dialect fidelity matters, use the same engine as production.
+
+### Parallel-safe tests
+When tests run in parallel (pytest-xdist, Jest workers, Go's `t.Parallel()`, JUnit parallel mode):
+- Never bind to hardcoded ports; use port 0 or dynamic allocation
```
jcdendrite: What about for running tests locally? Consider both local and CI testing.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+### What each test layer should verify
+
+**Unit tests:**
+- Every `if/else` branch in the function
+- Error return paths (missing input, API failure, invalid state)
+- Edge cases (zero values, null, empty strings, boundary conditions)
+- Side effects via mocks (did it call `.update()` with the right payload?)
+
+**Integration tests:**
+- **HTTP-level:** Auth rejection (401/403), authorized success (2xx), response shape (expected fields and types)
```
jcdendrite: There's some overlap with contract tests when you talk about response shape. Should be more specific here.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+# Testing Conventions
+
+## 1. Test pyramid
+
+| Layer | Speed | External deps | What it covers | Volume |
+|-------|-------|---------------|----------------|--------|
+| **Unit** (stubbed deps) | Milliseconds | None | Branch logic, error paths, edge cases, data transformations | Many — every code path |
+| **Integration** (real local services) | Hundreds of ms | Local DB, local services | Auth boundaries, response contracts, wiring between layers | Few — happy path + key error paths per endpoint or boundary |
+| **Contract** (schema verification) | Milliseconds–seconds | None (consumer) or local (provider) | API schemas between services match consumer expectations | One per service boundary |
+| **Smoke / E2E** (real external APIs) | Seconds | Third-party APIs, real infra | Full flow works end-to-end | 1-2 per feature, run on deploy not on every commit |
```
jcdendrite: Smoke tests and e2e tests are different entities for different purposes. They do involve actual real APIs, but I recommend not merging them.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+   - Locale-dependent formatting (number/date formats vary by system locale)
+   - Floating point comparison (use approximate equality, not exact)
+   - Async timing (asserting on async state without proper synchronization)
+   - Non-deterministic iteration order (hash maps/sets with no guaranteed order)
+   - Port or resource conflicts in parallel execution
+2. **Fix the layer** — if the test is flaky because it hits a real service to test logic, push it down the pyramid (see section 1)
+3. **Quarantine in-test retries** — retry loops inside test code mask the underlying issue. If a fix isn't immediate, quarantine the test (skip with a tracking issue) rather than letting it erode trust in the suite. CI-level retry policies (rerun failed tests once before failing the build) are a separate, reasonable practice for transient infrastructure issues — but track retry rates and investigate if a test needs retries frequently
+4. **Never delete without replacement** — a flaky integration test that covers an auth boundary still represents needed coverage. Replace it with a reliable test at the right layer before removing it
```
jcdendrite: Might be good to put this in a test debugging or evaluation section or separate skill, like with the section about whether a pyramid is inverted. I'm not convinced that this list will actually help with writing and planning the tests in the first place.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -0,0 +1,225 @@
...
+| Unconditional global state deletion | Breaks subsequent tests that need the value | Save and restore in guaranteed cleanup |
+| In-test retry loops for flaky tests | Masks root cause, inflates suite duration | Root-cause and fix or quarantine |
+| Test interdependence | Test B depends on state from Test A; reorder breaks both | Each test sets up and tears down its own state |
+| Assertion-free tests | Test runs code but never asserts; passes as long as nothing throws | Every test must assert on a specific expected outcome |
+| Sleep-based synchronization | `sleep(2)` to wait for async work; slow and still flaky | Use polling with timeout, await, or synchronization primitives |
+| Hardcoded colliding test data | Every test uses `id=1` or `email=test@example.com`; parallel runs collide | Generate unique identifiers per test |
+| Over-mocking | So many mocks that the test encodes the implementation, not behavior; brittle to refactors | Mock only direct dependencies; let integration tests cover wiring |
+| Mocking third-party internals | Mocking a library's internal API rather than its public interface | Use the library's test utilities or mock at your own abstraction boundary |
```
jcdendrite: Same comment as above. Better for test evaluation, like a code review skill or reference specific for test code.

claude/.claude/settings.json
```diff
@@ -18,5 +17,8 @@
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
   },
-  "effortLevel": "high"
+  "effortLevel": "high",
+  "enabledPlugins": {
```
jcdendrite: Don't commit this part about enabledPlugins. Can you put it in settings.local.json or does it need to be installed at the user level because it applies to about 7-10 repos for one of the projects I'm working on?

jcdendrite (commented):
Comments for Claude Code

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Agree — the inverted pyramid diagnostics are more useful when evaluating an existing suite than when writing new tests. Will move this and the related evaluation sections (flaky test diagnosis, anti-patterns) to a separate test-evaluation skill, keeping this skill focused on writing conventions.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** You're right — these are heuristics, not rules. Will change the heading to "When to consider moving a test down the pyramid" and soften the language to signal guidance rather than mandates.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Agreed — "relative to now" is a flaky test source (midnight crossings, TZ differences, DST transitions). Will remove it as a recommendation and keep only "inject timestamps as parameters."

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Good point — the section jumps straight to managing global state without first saying to avoid it. Will add a preamble: prefer designs that don't require global state in tests; when unavoidable, save and restore.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Fair — the parallel-safe guidance applies equally to local development. Will rephrase the intro to cover both local and CI contexts.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Good catch. The distinction: integration tests verify your code'\''s actual response shape (the real output from your endpoint). Contract tests verify schema compatibility between services without calling the other service. Will add a clarifying sentence to draw that boundary.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Agreed — smoke tests and E2E tests serve different purposes. Smoke tests are lightweight post-deploy health checks ("is the critical path up?"). E2E tests verify complete user flows end-to-end. Will split them into separate pyramid rows.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Agree — will move the flaky test diagnosis list to the evaluation skill. Though a few preventive items ("don'\''t use sleep-based sync", "don'\''t hardcode colliding data") are worth keeping in the writing skill since they prevent mistakes during authoring rather than diagnosing them after.

claude/.claude/skills/test-conventions/SKILL.md
jcdendrite: **[Claude Code]** Same plan — will move the full anti-patterns table to the evaluation skill. Will keep a short list of the most common authoring-time mistakes (assertion-free tests, sleep-based sync, hardcoded colliding data, over-mocking) inline in the writing skill.

claude/.claude/settings.json
jcdendrite: **[Claude Code]** This file shouldn'\''t be in the PR — the branch inherited 2 prior commits from main that were ahead of origin/main. Will clean up the PR scope.

For the plugin question: `enabledPlugins` works at all config levels (user, project, local). Since this is a dotfiles repo that deploys to `~/.claude/`, the cleanest option is to keep `enabledPlugins` in `~/.claude/settings.json` but exclude it from the committed dotfiles version. You could either:
1. Add it to a `settings.local.json` overlay that gets applied at deploy time but isn'\''t committed
2. Configure it directly in `~/.claude/settings.json` on each machine after dotfiles deploy

Option 1 is cleaner for a dotfiles workflow. Either way, the plugin will be available across all repos without committing it to the public repo.

jcdendrite: Let's go with Option #1. settings.local.json shouldn't be committed anyways

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -90,7 +75,7 @@ When replacing globals (HTTP client, fetch function, clock), always restore the
 - **In-memory vs. real engine:** In-memory databases (e.g., SQLite in-memory mode) are fast but have dialect differences (JSON columns, CTEs, locking behavior). When dialect fidelity matters, use the same engine as production.
 
 ### Parallel-safe tests
-When tests run in parallel (pytest-xdist, Jest workers, Go's `t.Parallel()`, JUnit parallel mode):
+Whether running locally or in CI, parallel test runners (pytest-xdist, Jest workers, Go's `t.Parallel()`, JUnit parallel mode) introduce isolation requirements:
 - Never bind to hardcoded ports; use port 0 or dynamic allocation
```
jcdendrite: Sorry, I meant this instruction about binding to hardcoded ports might be different for local setups versus CI...

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -20,31 +21,12 @@ user-invocable: false
...
-- The test sets up complex DB state to exercise one `if` branch
-- The test hits an external API just to test error handling
-- The test is slow (>500ms) and tests pure logic, not wiring
-- You're hitting rate limits on external APIs
-
-**Smoke → integration when:**
-- The test calls a real third-party API to verify your code handles errors
-- The test is flaky because of network conditions
-- The test creates real resources (emails, contacts) as a side effect
+**E2E tests** verify complete user journeys across the full stack. **Smoke tests** are lightweight post-deploy checks that verify critical paths are functional — they answer "is it up?" not "does every flow work?"
```
jcdendrite: Formatting inconsistency. These got pulled onto the same line, and there is a newline in between E2E and contract tests. I'm committing a fix.

claude/.claude/skills/test-conventions/SKILL.md
```diff
@@ -135,20 +120,24 @@ For tests that guard against a specific past bug, include a comment or docstring
...
-- **HTTP-level:** Auth rejection (401/403), authorized success (2xx), response shape (expected fields and types)
+- **HTTP-level:** Auth rejection (401/403), authorized success (2xx), response shape — the actual output your endpoint produces (fields, types, status codes)
 - **Service/module-level:** Wiring between layers maps results correctly, shared modules are called with expected arguments
 - Not every integration test needs to go through HTTP — service-level integration tests are cheaper when you're verifying wiring, not auth or response shape
 - Don't re-test every branch — that's the unit tests' job
 
 **Contract tests:**
-- Consumer expectations match the provider's actual schema
+- Consumer expectations match the provider's actual API schema — the agreed-upon interface between services, independent of either side's implementation
+- Distinct from integration response shape tests: contract tests verify the *schema agreement* between services; integration tests verify your *code's actual output*
```
jcdendrite, line 130: You may want to give more guidance as to when to make a response shape-based Integration test rather than a response shape-based contract test.

claude/.claude/skills/test-evaluation/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+
+These are heuristics, not rules — evaluate each case against the specific codebase's constraints.
+
+**Integration → unit when:**
+- The test sets up complex DB state to exercise one `if` branch
+- The test hits an external API just to test error handling
+- The test is slow (>500ms) and tests pure logic, not wiring
+- You're hitting rate limits on external APIs
+
+**E2E/Smoke → integration when:**
```
jcdendrite: E2E - > integration makes sense. Smoke tests have a totally different purpose - I don't think you can provide guidance to convert a smoke test to an integration test...

claude/.claude/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "model": "Opus",
```
jcdendrite: let's not commit the model

claude/.claude/settings.json
```diff
@@ -17,8 +18,5 @@
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
   },
-  "effortLevel": "high",
-  "enabledPlugins": {
-    "atlassian@claude-plugins-official": true
-  }
+  "effortLevel": "high"
```
jcdendrite: also shouldn't commit effort level

claude/.claude/skills/test-conventions/SKILL.md
jared-yetimyfriend: **[Claude Code]** Good point — locally you might intentionally use a fixed port for convenience (e.g., localhost:3000 for a dev server). The "never hardcode" advice is specifically about tests that may run in parallel. Will qualify this to distinguish test port binding from dev server conventions.

claude/.claude/skills/test-conventions/SKILL.md
jared-yetimyfriend: **[Claude Code]** Noted — your formatting fix is on disk, will include it in the next commit.

claude/.claude/skills/test-conventions/SKILL.md
jared-yetimyfriend: **[Claude Code]** Will add decision criteria. The key distinction: use integration response shape tests when you own both sides (or there'\''s only one consumer) and can test against a running instance. Use contract tests when provider and consumer are owned by different teams, deploy independently, or when spinning up the provider in tests is expensive.

claude/.claude/skills/test-evaluation/SKILL.md
jared-yetimyfriend: **[Claude Code]** You'\''re right — smoke tests verify deployment health, not behavior. You wouldn'\''t "move a smoke test down" because it'\''s not testing the same thing as integration tests. Will change this to only cover E2E → integration.

claude/.claude/settings.json
jared-yetimyfriend: **[Claude Code]** Will move both model and effortLevel to settings.local.json — these are personal preferences, not shared config.

claude/.claude/settings.json
jared-yetimyfriend: **[Claude Code]** Agreed — moving to settings.local.json with model.

PR #10, opened 2026-04-11
Automated gate hooks for code-review, respond-pr, and settings edits

jcdendrite:
## Summary

Replaces the advisory "ask" pattern with **deny-redirect** hooks that force Claude to run the relevant review skill before taking gated actions. Solves the silent-failure problem where Claude might skip `/code-review` and fabricate having run it.

- **`require-code-review.sh`** — blocks `git commit` (including chained forms like `git add . && git commit`) until the staged diff's sha256 matches a marker at `~/.claude/review-markers/<repo-hash>`. Marker is written by `/code-review` on clean review and auto-invalidates the moment the staging area changes. Marker lives under `$HOME`, not inside the repo, so it never pollutes `git status`.
- **`require-respond-pr.sh`** — blocks PR comment reads/posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, which fetches **all three** comment types (inline, top-level review bodies, issue-level — Claude habitually fetches only the first and misses real feedback) and enforces the `**[Claude Code]**` attribution prefix on replies. Honors a 60-minute bypass marker at `~/.claude/.respond-pr-active` that the skill sets on entry and removes on exit, with staleness cutoff to auto-recover from orphaned markers.
- **`ask-review-permissions.sh`** — asks before `Edit`/`Write`/`MultiEdit` to `.claude/settings*.json`, nudging toward `/review-permissions` if the edit touches `permissions.allow`. Kept in `ask` mode because a precise "does this edit touch permissions.allow" heuristic is fuzzy.
- **`hooks/tests/test_hooks.py`** — 37 end-to-end pytest cases covering allow, deny, and ask paths for all three hooks. Uses a temp git repo fixture for realistic `git diff --cached` state and sandboxes `$HOME` per test via `monkeypatch`, so the suite never touches the user's real `~/.claude/` state. Parametrized decision tables for the stateless cases; explicit tests for marker/bypass state transitions.
- **`.github/workflows/hooks.yml`** — runs the pytest suite on PRs and main pushes that touch `claude/.claude/hooks/**` or the workflow file itself. `paths:` filter on both triggers keeps unrelated changes (CLAUDE.md, skills, starship, etc.) from burning GHA minutes. `permissions: contents: read` for least-privilege, `timeout-minutes: 2` (3× headroom over p99 setup) to cap any future runaway hook subprocess, and `concurrency` group with `cancel-in-progress` to kill superseded runs on rapid pushes. Pinned to `ubuntu-24.04` and `pytest==8.*` for reproducibility.
- **`.gitignore`** — ignores `__pycache__/` and `.pytest_cache/` so the new pytest suite doesn't leave artifacts in `git status`.

### Notable during development
A trailing-newline bug surfaced mid-commit: the hook computes the repo hash via `printf '%s' "$REPO_ROOT"` (no newline), while the original SKILL.md command piped `git rev-parse --show-toplevel | sha256sum` directly (hashing the newline), so the marker landed at a path the hook never checked. Fixed by adding `tr -d '\n'` to the skill command plus an end-to-end regression test (`test_skill_marker_write_command_matches_hook_path`) that runs the exact command from SKILL.md and verifies the hook accepts the resulting marker — so any future drift between the two sides will fail.

### Testing framework choice
Tests were originally written in bash (`hook-tests.sh`) since the SUT is bash. Ported to pytest on review: parametrize collapses the decision tables, `monkeypatch`/`tmp_path` sandbox `$HOME` cleanly (no backup/restore of live state), JSON assertions are clearer than `jq` piped inside a bash helper, and `python3` + `pytest` are available on every CI runner without a separate install step. Hooks themselves stay in bash — PreToolUse cold start matters, and `python3 -c '...'` is ~10× slower than bash for this workload.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 37/37 passing locally
- [x] CI workflow triggers on hook path changes only (verified via `paths:` filter on both push and pull_request)
- [x] Verified `git commit` blocked on no-marker, wrong-hash, and stale-marker states
- [x] Verified `git add file && git commit -m foo` (chained form) blocked when marker stale
- [x] Verified `git commit-tree` not a false positive
- [x] Verified `gh pr view`, `gh pr list`, `gh api user` allowed (non-matching)
- [x] Verified bypass marker allows skill's own `gh api` calls when fresh, denies when stale
- [x] Verified end-to-end: skill's marker-write command produces a path the hook accepts


claude/.claude/hooks/tests/test_hooks.py
```diff
@@ -0,0 +1,315 @@
...
+# ---------------------------------------------------------------------------
+
+
+class TestAskReviewPermissions:
+    @pytest.mark.parametrize(
+        "tool_input",
+        [
+            edit_input("/home/jared/foo/.claude/settings.json"),
+            edit_input("/home/jared/foo/.claude/settings.local.json"),
+            write_input("/home/jared/foo/.claude/settings.json"),
```
jcdendrite: Shouldn't hard-code my personal dir

claude/.claude/settings.json
```diff
@@ -1,14 +1,30 @@
 {
+  "model": "opus",
```
jcdendrite: Remove - shouldn't be hard-coded

claude/.claude/settings.json
```diff
@@ -1,14 +1,30 @@
...
   "hooks": {
     "PreToolUse": [
       {
         "matcher": "Bash",
         "hooks": [
           {
             "type": "command",
-            "if": "Bash(git commit *)",
             "command": "~/.claude/hooks/require-code-review.sh",
+            "if": "Bash(git commit *)",
```
jcdendrite, line 9: Nit - this was confusing - you moved the order of the json fields but didn't change the field value

claude/.claude/settings.json
```diff
@@ -17,5 +33,9 @@
   "statusLine": {
     "type": "command",
     "command": "bash ~/.claude/statusline-command.sh"
-  }
+  },
+  "enabledPlugins": {
+    "atlassian@claude-plugins-official": true
+  },
+  "effortLevel": "high"
```
jcdendrite: Shouldn't be hard-coded

claude/.claude/hooks/tests/test_hooks.py
jcdendrite: **[Claude Code]** Fixed — replaced all `/home/jared/foo` and `/home/jared/.claude/` references with `/some/project/` in the parametrized cases.

claude/.claude/settings.json
jcdendrite: **[Claude Code]** Removed `"model": "opus"` in commit 615d053.

claude/.claude/settings.json
jcdendrite: **[Claude Code]** Good catch — restored the original `"if"` before `"command"` order in commit 615d053 so the existing hook entry shows as unchanged in the diff. The swap was a prior-session artifact with no semantic purpose.

claude/.claude/settings.json
jcdendrite: **[Claude Code]** Removed `"effortLevel": "high"` (and `"enabledPlugins"` two lines up) in commit 615d053.

PR #16, opened 2026-04-21
Add query-conventions skill

jcdendrite:
## Summary
- New skill `query-conventions` — loads when writing or modifying database queries (SQL, PostgREST/Supabase `.from(...)`, ORMs, raw drivers).
- Three rules targeting AI anti-patterns that scale badly under growth: explicit limits, no N+1 from async loops, explicit column selection. Plus an adjacent anti-patterns section (check-then-insert, `COUNT(*)` for existence, unindexed filters, compute-in-app-instead-of-SQL).
- Parallels `test-conventions` in shape and scope — patterns for *writing*, not for reviewing or running.

## Why this exists
Surfaced during the DAY-484 (daytag) four-specialist review of Lovable guardrails — staff backend and staff data engineer both flagged unbounded queries and N+1 iteration as high-incidence AI anti-patterns. Keeping the global CLAUDE.md minimal (per DeployHQ / Termdock guidance and our own length research) meant the right home was a trigger-loaded skill, not always-on prose.

## Test plan
- [x] Skill format matches existing `test-conventions` / `config-environments` structure
- [ ] After merge, trigger a test query-writing task in any repo and verify the skill loads


claude/.claude/skills/query-conventions/SKILL.md
```diff
@@ -0,0 +1,76 @@
+---
+name: query-conventions
```
jcdendrite: This sklll is specific to sql queries so should probably be named as such. Currently it doesn't cover conventions for NoSQL setups at all (like DynamoDB or Mongo)

claude/.claude/skills/query-conventions/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+common AI anti-pattern in data-access code.
+
+```
+# WRONG
+for id in ids:
+    row = await db.from('table').select().eq('id', id)
+    results.append(row)
+```
+
+Fix by batching:
```
jcdendrite: Do we need to add more guidance around appropriate batch size?

jcdendrite (commented):
Make sure to have a staff analytics engineer with two decades of SQL expertise do another pass.

claude/.claude/skills/query-conventions/SKILL.md
jcdendrite: **[Claude Code]** Agreed — renamed the skill to `sql-query-conventions`, tightened the TRIGGER to SQL / PostgREST / SQL-shaped ORMs, and added an explicit DO-NOT-TRIGGER exclusion for document/KV stores (DynamoDB, Mongo, Cassandra). A future `nosql-query-conventions` skill is the right home for the other half. Commit 5f756e9.

claude/.claude/skills/query-conventions/SKILL.md
jcdendrite: **[Claude Code]** Yes — qualitative-only let agents ship `.in("id", ten_thousand_ids)` that blows up in prod. Added a concrete ceilings table covering Postgres (~1k–5k, with `= ANY($1::int[])` as the scaling alternative), Supabase `.in()` (~few hundred, URL-length bound), BigQuery (~1k, switch to `UNNEST` or temp-table JOIN), Snowflake (16,384 IN expressions hard limit). Commit 5f756e9, rule 2.

jcdendrite:
**[Claude Code]** Consulted a staff analytics engineer with ~20 years SQL experience (Postgres, MySQL, BigQuery, Snowflake, Redshift, DuckDB). Findings applied in commit 5f756e9:

- Rename to `sql-query-conventions` (rule set is relational-specific; NoSQL belongs in a separate skill)
- **Batch-size ceilings** — per-backend table as described in the line-50 reply
- Rule 1: split `.single()` (errors on non-exact-one) vs `.maybeSingle()` / `LIMIT 1` (zero-or-one silent); conflating them teaches a bug either way
- Rule 1: **stable pagination** requires a total-ordered `ORDER BY` with a deterministic tiebreaker
- Rule 1: **keyset pagination** recommended over offset for large page numbers (offset is O(N))
- Anti-patterns: added `SELECT ... FOR UPDATE` for the legitimate read-then-write case, adjacent to check-then-insert
- Anti-patterns: tightened `COUNT(*)` — use `EXISTS` / `LIMIT 1` for existence checks; `COUNT(*)` is fine for actual counts
- Anti-patterns: strengthened compute-in-DB with explicit window-function callouts (`ROW_NUMBER` / `RANK` / `FILTER`) since agents do not reach for them by default
- New **Verification** section: `EXPLAIN` / `EXPLAIN ANALYZE` guidance — check planned index usage, join order, row-estimate accuracy

jcdendrite (commented):
lgtm

PR #19, opened 2026-04-22
Add git-feature-branch-sync skill

jcdendrite:
## Summary

- Adds a global auto-triggering skill (`user-invocable: false`) that gives Claude a decision framework for keeping a feature branch current with `main`: when to rebase-and-force-push vs merge-main-in, and how to force-push safely via `--force-with-lease --force-if-includes`.
- Deliberately scoped as the *global framework*. Per-client policy — squash-merge vs merge-commit, rebase-after-review rules, protected-branch specifics — lives in each repo's CLAUDE.md or a client-scoped skill, not in this skill.
- Cross-references the existing `git-state-safety` skill for bad-merge recovery.

## Review process

Initial draft was reviewed via `/code-review` + two specialist agents (staff DevOps engineer, security engineer). Their findings drove the second commit on this branch, which:

- Expands the never-force-push list to include tags and environment branches (`staging`, `production`, `qa`, `uat`), and clarifies that the *remote's* default branch (`git symbolic-ref refs/remotes/origin/HEAD`) is authoritative, not `init.defaultBranch`.
- Warns that server-side branch protection may be absent (GitHub Free private repos, or simply unconfigured) and must not be the only guardrail.
- Adds a signed-commits check to the pre-flight (rebase drops signatures unless `commit.gpgsign=true`).
- Adds an independent-verification step for "personal branch" claims (`git log --format=%ae`), so the agent doesn't rely solely on an unverified user claim.
- Fixes the recovery recipe: pin the recovered SHA in a backup ref *before* anything else, fetch and merge reconciliation commits *before* force-pushing, and corrects the reflog unreachable-expiry default (90 days, not 30).
- Notes that `--force-if-includes` silently no-ops when reflogs are disabled (`core.logAllRefUpdates=false`).
- Reframes "rebase default" as opinionated rather than universal.

A verification pass by the same two specialists confirmed the fixes were adequate and flagged no regressions.

## Test plan

- [ ] Review the skill's frontmatter TRIGGER / DO NOT TRIGGER conditions — does the scope feel right (fires on rebase/merge/force-push decisions, defers to `git-state-safety` for mid-merge state)?
- [ ] Sanity-check the recovery recipe ordering: find SHA → pin → fetch + reconcile → restore → coordinate.
- [ ] Spot-check `git symbolic-ref refs/remotes/origin/HEAD` behavior on a repo where local `init.defaultBranch` differs from the remote default, to confirm the guidance is accurate.

### DIFF
--- a/claude/.claude/skills/git-feature-branch-sync/SKILL.md
+++ b/claude/.claude/skills/git-feature-branch-sync/SKILL.md
@@ -31,9 +31,11 @@ Keeping a feature branch current with `main` has two shapes:
 - **Merge** — `git merge origin/main`. Creates a merge commit on the
   feature branch. Preserves SHAs. No force-push needed.
 
-**Default: rebase on personal feature branches.** Switch to merge when any
+**This skill's default is rebase on personal feature branches** — a
+reasonable and common choice, but not universal (many shops default to
+merge even on personal branches). Switch to merge when any
 *merge-required condition* below applies. Always defer to per-client
-policy if it contradicts this default.
+policy when it contradicts this default.
 
 ## Merge-required conditions
 
@@ -72,19 +74,33 @@ git push --force-with-lease --force-if-includes
 `--force-with-lease` alone has a known gap: a stray `git fetch` updates
 the remote-tracking ref, which makes the lease look held even though you
 haven't integrated the remote's new commits. `--force-if-includes` closes
-that gap. Use both.
+that gap by checking your local reflog against the remote-tracking ref.
+Use both. Note: `--force-if-includes` silently no-ops if reflogs are
+disabled (`core.logAllRefUpdates=false`) — uncommon but seen in some CI
+images.
 
 ## Never-force-push targets
 
 Refuse to force-push to any of these regardless of flavor:
 
-- `main`, `master`, `trunk`, `develop`, or any default branch
-- `release/*`, `hotfix/*`, or any other protected / shared branch
+- `main`, `master`, `trunk`, `develop`, or **the remote's default branch**
+  (check `git symbolic-ref refs/remotes/origin/HEAD` —
+  `init.defaultBranch` is the local default, not authoritative)
+- `release/*`, `hotfix/*`, `staging`, `production`, `qa`, `uat`, or any
+  other shared / environment-tracking branch
+- **Tags** (`refs/tags/*`) — never force-push. Tags are the
+  supply-chain anchor for releases and SBOMs; signed tags especially
+  must never be rewritten.
 - Any branch the user has not explicitly said is personal
 
 For those branches, rebasing the branch itself is also off the table —
 they keep merge history. Bringing `main` into them (when it even makes
 sense) is a merge, not a rebase.
+
+**Do not rely on server-side branch protection** to catch mistakes. On
+GitHub Free private repos the branch-protection API is inaccessible; on
+other repos the rules may simply not be configured. Treat the list above
+as the last line of defense, not the only one.
 
 ## Interaction with PR merge strategy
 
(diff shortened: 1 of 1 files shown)


claude/.claude/skills/git-feature-branch-sync/SKILL.md
```diff
@@ -0,0 +1,173 @@
+---
+name: git-feature-branch-sync
+description: >
+  Decision framework for keeping a feature branch up to date with main:
+  when to rebase-and-force-push vs merge-main-in, how to force-push
+  safely (`--force-with-lease` / `--force-if-includes` / plain `--force`),
+  and when the "never force-push" instinct is correct vs overcautious.
+  This is the global framework — per-client policy (squash-merge vs
```
jcdendrite: Use "per project" rather than "per client" given that this is an open-source repo and not specific to consultants. "Project" is more universal, as it includes professional arrangements of a consultant, FTE work, and personal projects.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+  mid-merge / mid-rebase / mid-cherry-pick state (use `git-state-safety`
+  instead).
+user-invocable: false
+---
+
+# Feature-Branch Sync: Rebase, Merge, and Force-Push
+
+## The decision
+
+Keeping a feature branch current with `main` has two shapes:
```
jcdendrite: Keep in mind `main` might not be the actual default branch for repos. Be more inclusive of setups here. It's fine to use `main` as an example.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+
+- **Another active committer** on the branch. Rebasing rewrites history
+  they have based work on; they will need a recovery dance.
+- **Another branch is stacked on this one.** Rebase invalidates the
+  child's base. If the stack is intentional and you must rebase, use
+  `git rebase --update-refs` and coordinate.
+- **Review is already in flight** in a culture that anchors comments to
+  specific commit SHAs. Rebasing detaches comments from their anchors.
+  Some teams forbid rebase post-review; prefer "rework" commits.
+- **Per-client policy says merge-only.** Check the repo's CLAUDE.md /
```
jcdendrite: Not sure what you mean by this statement. Is this an example of a project-specific policy? Or are you pulling in a policy from a client?

claude/.claude/skills/git-feature-branch-sync/SKILL.md
```diff
@@ -0,0 +1,173 @@
...
+that gap by checking your local reflog against the remote-tracking ref.
+Use both. Note: `--force-if-includes` silently no-ops if reflogs are
+disabled (`core.logAllRefUpdates=false`) — uncommon but seen in some CI
+images.
+
+## Never-force-push targets
+
+Refuse to force-push to any of these regardless of flavor:
+
+- `main`, `master`, `trunk`, `develop`, or **the remote's default branch**
```
jcdendrite: I had a comment above which references the assumption of `main` as default branch. This statement is closer to what I'm looking for in terms of  being more inclusive to setups.

jcdendrite (commented):
Also make sure to get reviews from two agents each with two decades of Git experience, a staff backend engineer and a staff DevOps engineer.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
jcdendrite: **[Claude Code]** Agreed — replaced "per-client" with "per-project" throughout (description, merge-required conditions, PR-merge-strategy section, rule of thumb). Open-source, FTE, and personal-project contexts all included. Commit a5113b6.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
jcdendrite: **[Claude Code]** Added a preamble that defines `main` as a placeholder and tells the reader to substitute per-repo via `git symbolic-ref refs/remotes/origin/HEAD`. The never-force-push list now leads with "the remote's default branch" rather than listing `main` first, and the surrounding prose says "the default branch" rather than "`main`". Where `main` still appears in command examples it is explicitly an example. Commit a5113b6.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
jcdendrite: **[Claude Code]** Good catch — reworded to be explicit. It is the per-project policy layer: the repo's own CLAUDE.md (or a project-scoped skill) may say "squash-merge PRs and use merge-main to stay current," and this skill defers to that. Not a policy pulled in from an external client. Commit a5113b6.

claude/.claude/skills/git-feature-branch-sync/SKILL.md
jcdendrite: **[Claude Code]** Thanks — took that phrasing as the model and applied the same pattern to the description, the preamble, and line 26. Commit a5113b6.

jcdendrite:
**[Claude Code]** Consulted staff backend + staff DevOps engineers (both with ~20 years of Git experience). Findings applied in commit `a5113b6`:

**Critical CI gotcha:** `--force-with-lease` without an explicit expected value degrades silently to `--force` semantics on shallow / fresh / missing-tracking-ref clones. Documented the explicit `--force-with-lease=<branch>:$(git rev-parse origin/<branch>)` form and required it in automation.

**Rebase-vs-merge precision:**
- `git merge --no-ff origin/main` when a merge commit is the goal (plain `git merge` can fast-forward)
- `git pull --rebase` = `fetch + rebase`; prefer the two-step form so the fetch is auditable

**Stacked-PR expansion:** `git rebase --update-refs` + `rebase.updateRefs=true`; force-push stacks leaf-to-root (a failed lease mid-stack leaves the stack half-rewritten).

**Long-lived branches:** `rerere.enabled=true` — highest-ROI configuration for a branch that rebases main repeatedly.

**Pre-flight hardening:**
- Use `%ae %ce` (not just `%ae`) in the personal-branch verification so committers aren't hidden by rebases/cherry-picks
- `git fetch --unshallow` before trusting history in CI
- `commit.gpgsign=true` requires the signing key to be available **non-interactively** (critical on headless runners — silent unsigned commits otherwise)
- Explicit fail-closed step for ambiguous automation cases

**Protected-branch interactions beyond "may not exist":**
- Required linear history (rejects merge-main)
- Required signed commits
- Required status checks (force-push re-pends the PR and can knock it out of auto-merge)
- Actor-specific rules (bots excluded even when humans are allowed)
- Added API-query guidance with fail-closed default

**Post-force-push communication:** After a legitimate force-push on a branch with review activity, post a PR comment with pre/post SHAs and `git range-diff` output so reviewers don't lose their comment anchors. Senior-engineer move that costs one comment.

**Deferred as out-of-scope:** Gerrit / Azure / GHES / Bitbucket platform-specific force-push quirks (would belong in a platform-notes skill); `git rebase --exec` checkpointing; worktree-based rebase isolation.

The sibling session independently reviewed this PR before yours and flagged no blockers; their validation against primary Git docs (git-scm.com/docs/git-push) confirmed the `--force-with-lease` / `--force-if-includes` gap claims.

jcdendrite (commented):
lgtm

PR #24, opened 2026-04-23
Add worktree-enforcement hook with per-repo opt-in sentinel

jcdendrite:
## Summary

- Adds a new PreToolUse Bash hook, `require-worktree-for-git-writes.sh`, that denies non-read-only git operations in the main working tree of repos that commit a `.claude/worktree-required` sentinel at the repo root. Operations inside a linked git worktree pass through cleanly.
- Opt-in is per repo via the committed sentinel — repos without the file are unaffected. Ships the sentinel in this repo so enforcement activates on `claude-config` itself after the next `git pull`.
- 40 new pytest cases bring the suite to 117 total.

## Motivation

Concurrent Claude Code sessions on the same working tree can race: one session's `git reset --hard`, `git stash`, or `git checkout` silently wipes another session's uncommitted edits. See [anthropics/claude-code#34327](https://github.com/anthropics/claude-code/issues/34327) for examples in the wild. Worktree isolation is the established fix; this hook makes it enforceable rather than hopeful.

## Design decisions

- **Committed sentinel (vs. local git config)**: the sentinel travels with the repo, so all clones enforce consistently — no per-machine, per-clone setup that can be forgotten. `git config --local` was ruled out because the "did I remember?" surface area compounds across developers, repos, and machines; for a safety hook, forgetting puts you back in the race.
- **`git-dir != git-common-dir` check (vs. path-pattern match)**: semantically correct detection of linked worktrees. Resistant to `GIT_DIR` env-var spoofing and to false matches on repos literally located at paths containing `/worktrees/`. The hook also defensively unsets `GIT_DIR`/`GIT_WORK_TREE`.
- **`worktree` on the allowlist**: the remediation tells users to run `git worktree add`; denying that command on the main tree would strand them.
- **`config` NOT on the allowlist**: `git config alias.x '!sh -c "..."'`, `core.pager`, `credential.helper`, `core.sshCommand` all fire on the next git invocation. Treating `git config` as "read-only because it doesn't mutate the tree" was wrong — it mutates a file that becomes executable code.
- **Fail-closed everywhere**: malformed stdin JSON, unparseable git subcommands, and globbing inside the parser all deny rather than silently allow.

## Review

Three specialist reviewers (Senior Bash/Security, Claude Code config, SDET) surfaced issues around `git config` in the allowlist (ACE risk), `GIT_DIR` env-var bypass, globbing inside `extract_git_subcmd`, and JSON parse-failure fail-open. All addressed before commit.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 117 passed
- [ ] After merging, `git pull` in the local `claude-config` checkout. Stow already folded `~/.claude/hooks` as a symlink to the repo's hooks dir, so the new script is visible immediately through the existing symlink — no re-stow needed.
- [ ] Confirm enforcement activates: `cd ~/MyCode/claude-config && git commit --allow-empty -m test` should deny (main tree, sentinel committed)
- [ ] Confirm worktree path works: `git worktree add .claude/worktrees/smoke -b smoke && cd .claude/worktrees/smoke && git commit --allow-empty -m test` should pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
 - **`require-code-review.sh`** — blocks `git commit` (including chained forms like `git add . && git commit`) until `/code-review` has run on the current staged state. Verified via sha256 marker in `~/.claude/review-markers/<repo-hash>`, which auto-invalidates the moment the staging area changes.
 - **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a 60-minute bypass marker at `~/.claude/.respond-pr-active` that the skill sets on entry and removes on exit.
 - **`ask-review-permissions.sh`** — asks before `Edit`/`Write`/`MultiEdit` to `.claude/settings*.json`, nudging toward `/review-permissions` when the edit touches `permissions.allow`.
+- **`require-worktree-for-git-writes.sh`** — opt-in per repo. When active, denies non-read-only git operations unless the session runs in a linked git worktree. Prevents concurrent Claude Code sessions from racing on the same working tree. See [Worktree enforcement](#worktree-enforcement) below for opt-in instructions.
 
 ### Skills (slash commands)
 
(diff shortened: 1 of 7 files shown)


claude/.claude/CLAUDE.md
```diff
@@ -31,8 +31,11 @@
 
 - After writing or modifying an implementation plan, run `/plan-review` before presenting the plan to the user. If the review finds issues, address them first, then present the final version.
 
+## Worktree enforcement
+
+- If the worktree-enforcement hook denies a git command, remediate by `cd`-ing into (or creating) a worktree under `.claude/worktrees/` and retrying — full remediation is in the deny message.
+
```
jcdendrite: This level of prescriptiveness is excessive. Claude should be able to figure this out.

jcdendrite: Unless you really think it won't and it will end up saving a memory, which defeats the point of this.

README.md
```diff
@@ -41,9 +42,47 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
...
+Concurrent Claude Code sessions that share a working tree can race: one session's `git reset --hard`, `git stash`, or `git checkout` silently wipes another session's uncommitted edits. See [Claude Code issue #34327](https://github.com/anthropics/claude-code/issues/34327) for examples of this failure mode in the wild.
+
+`require-worktree-for-git-writes.sh` mitigates by denying non-read-only git operations (`commit`, `push`, `rebase`, `reset`, `merge`, `checkout`, etc.) unless the session runs inside a linked git worktree. Read-only commands (`status`, `log`, `diff`, `fetch`, `show`, `blame`, etc.) are always allowed. The hook is opt-in per repo via a committed sentinel file.
+
+### Activating enforcement on a repo
+
+```bash
+cd path/to/your/repo
+
+mkdir -p .claude
```
jcdendrite, line 58: What if the repo already has a .claude directory

README.md
```diff
@@ -41,9 +42,47 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
...
+
+echo '.claude/worktrees/' >> .gitignore
+
+git add .claude/worktree-required .gitignore
+git commit -m "Activate Claude Code worktree enforcement"
+```
+
+### Working inside a worktree
+
+With enforcement active, start sessions for non-trivial work in a worktree instead of the main tree:
```
jcdendrite, line 75: Might be helpful to explain what a worktree is because this is the README

claude/.claude/CLAUDE.md
jcdendrite: **[Claude Code]** Agreed — removed it in 6157460. The hook’s deny message already spells out remediation (`cd` into a worktree, create one with `git worktree add`, or spawn with `isolation: worktree`). On the memory-save concern: if Claude fails to act on the deny directly and ends up saving a memory, that’s a signal to make the deny message more explicit — not to re-add CLAUDE.md duplication.

README.md
jcdendrite: **[Claude Code]** Added a coexistence note in 6157460 before the activation snippet: `mkdir -p` is a no-op if `.claude/` already exists, and the sentinel is an inert marker file alongside any existing content (plans/, settings.local.json, worktrees/, etc.). Also made the `.gitignore` append idempotent (`grep -qxF ... || echo >>`) so rerunning the steps on a repo that already has the line does not duplicate it.

README.md
jcdendrite: **[Claude Code]** Added a one-paragraph primer in 6157460 right before the worktree-usage snippet: explains that a git worktree is a linked working directory on a separate branch sharing the repo’s `.git` object storage, created via `git worktree add <path> -b <branch>`. Links to the git-worktree man page for readers who want more.

jcdendrite (commented):
lgtm

PR #31, opened 2026-04-23
Fix worktree-enforcement hook word-boundary false positive

jcdendrite:
## Summary

`require-worktree-for-git-writes.sh` used `*git*` substring checks that matched `.github`, `.gitignore`, `github.com`, and similar, then fed fragments like `ls .github/workflows/` into `extract_git_subcmd`, which returned `hub/workflows/` as the alleged subcommand and denied the read. Harmless reads of this repo's own `.github/` and `.gitignore` were being blocked.

- Top-level fast-path replaced with a word-boundary regex: `(^|[^[:alnum:]])git([^[:alnum:]]|$)`. Rejects `.github`, `.gitignore`, `longitude`, `gitk`; accepts `git`, `/usr/bin/git`, `FOO=1 git`.
- New `fragment_invokes_git` helper scans each fragment's words and only returns success when some word is exactly `git` or ends in `/git`. Replaces the per-fragment `*git*` check. `extract_git_subcmd` is unchanged and runs only after the helper validates a real git invocation.

Pre-existing edge case left alone: `A=foo.git git log` still extracts `git` as the subcommand and denies instead of resolving to `log`. Not covered by any existing test and outside the scope of this fix.

## Test plan

- [x] 9 new regression tests under `TestRequireWorktreeForGitWrites`:
  - Parametrized: `ls .github/workflows/`, `cat .gitignore`, `grep -r github.com /src`, `find . -name '*.git'`, `./git-foo`, `gitk master` → allow
  - `ls .github/workflows/ && git log --oneline` → allow (chained read-only git after `.github` fragment)
  - `cat .gitignore && git commit -m x` → deny (real write still caught after `.gitignore` fragment)
  - `git log -- .github/workflows/hooks.yml` → allow (positive: real git command whose arg contains `.github`)
- [x] Full suite: 125 passed (117 pre-existing + 8 new — the parametrized case counts once in the name but fires 6 subtests, making the total 125)
- [x] Reviewed by bash/shell-safety specialist and SDET specialist under `/code-review`; SDET's `gitk`-style boundary test suggestion incorporated

### DIFF
--- a/claude/.claude/hooks/require-worktree-for-git-writes.sh
+++ b/claude/.claude/hooks/require-worktree-for-git-writes.sh
@@ -38,10 +38,12 @@ if [ "$JQ_EXIT" -ne 0 ]; then
   exit 0
 fi
 
-# Fast-path: commands that don't mention git are not our concern. Use bash
-# pattern matching instead of grep to avoid locale-dependent behavior on
-# commands containing non-UTF-8 bytes.
-if [[ "$COMMAND" != *git* ]]; then
+# Fast-path: commands that don't mention `git` as a word are not our
+# concern. A plain `*git*` substring check false-positives on `.github`,
+# `.gitignore`, `github.com`, `longitude`, and similar, blocking harmless
+# reads like `ls .github/workflows/`. Require a non-alnum boundary (or
+# string edge) on both sides so `git` fires only as a command word.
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #35, opened 2026-04-23
Normalize settings.json field order to match CLI auto-format

jcdendrite:
## Summary

Pure reordering of `claude/.claude/settings.json`, no semantic change:

- Inside hook entries: `"command"` before `"if"` (was `"if"` first).
- At top level: `"enabledPlugins"` before `"extraKnownMarketplaces"` (was after).

## Why

The Claude CLI normalizes to this shape whenever it touches the file. That means every session starts with the same dirty-tree diff that either gets ignored (and reopens next session) or accidentally bundled into an unrelated PR. Committing the normalized form stops the churn.

Confirmed `jq .` parses the result — JSON is still valid, and field order has no semantic effect on settings evaluation.

## Test plan

- [ ] Merge, pull, and confirm `git status` stays clean on main after `claude` does normal housekeeping touches.

### DIFF
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -6,14 +6,14 @@
         "hooks": [
           {
             "type": "command",
-            "if": "Bash(git commit *)",
             "command": "~/.claude/hooks/require-code-review.sh",
+            "if": "Bash(git commit *)",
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #36, opened 2026-04-23
Scope deny-client-refs.sh to claude-config and exempt its own tests

jcdendrite:
## Summary

- **Fix the root cause:** the redaction gate was registered globally with no repo-identity check, so it fired on every `git commit` in every repo where the config was installed — silently blocking legitimate commits that reference the host repo's own tracker IDs. Add an early-exit when `origin.url` doesn't contain `claude-config`, so the gate only runs in the repo whose public-leak concern motivated it. Substring match also covers forks, which inherit the same redaction discipline.
- **Fix the adjacent paper cut:** exempt the hook's own test directory from the staged-diff scan. Tests for this hook need synthetic tracker tokens as test data (the file's header lists them as invented prefixes), and the pre-fix behavior would block any commit that added a new test case. The exemption is narrow — `:(top,exclude)claude/.claude/hooks/tests/**` — and a suspect token in a non-test file staged alongside a test-dir change still blocks.
- **Tests:** migrate the existing `TestDenyClientRefs` cases to a new `claude_config_repo` fixture so they continue to exercise the deny path under scoping. Add seven new tests: five for scoping boundaries (unrelated remote, no remote, fork origin, preserved reason message, staged-diff scoping) and two for the test-dir exemption (allow path and over-broad-pathspec guard).

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 133 passed
- [x] Manual: invoked worktree hook directly in an unrelated repo with a suspect token in the commit message → exit 0 (allow)
- [x] Manual: invoked worktree hook directly in claude-config with a suspect token → deny with expected reason message
- [x] `:(top,exclude)` pathspec sanity-checked against a temporary test repo with both test-dir and root-level changes
- [ ] After merge, `git pull` in the main tree so the stow-linked live hook picks up the fix

### DIFF
--- a/claude/.claude/hooks/deny-client-refs.sh
+++ b/claude/.claude/hooks/deny-client-refs.sh
@@ -33,9 +33,28 @@ if [ -z "$REPO_ROOT" ]; then
   exit 0
 fi
 
-STAGED_DIFF=$(git diff --cached 2>/dev/null)
+# Scope: this redaction gate exists to protect the claude-config repo,
+# where accidental references to consulting clients would leak publicly.
+# Other repos legitimately reference their own tracker IDs. Short-circuit
+# unless origin.url looks like claude-config. `git config --get` returns
+# empty (not an error exit) when the remote is missing, so the substring
+# check safely handles the no-remote case too.
+REMOTE_URL=$(git config --get remote.origin.url 2>/dev/null)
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #34, opened 2026-04-23
Add seven reviewer subagent personas and wire into plan/code review

jcdendrite:
## Summary

- **Eight** stack-agnostic reviewer subagent personas in `claude/.claude/agents/`, with a new `staff-analytics-engineer` covering warehouse-side modeling and source-schema ELT-readiness review.
- `plan-review` and `code-review` skills carry **Item ownership** tables as the single source of truth for routing items to reviewers; agent files no longer enumerate owned items by number (dependency inversion). Renumbering items in a skill cannot drift the agent files.
- Schema-change diffs route three ways: `staff-backend-engineer` designs, `staff-data-engineer` flags pipeline impact and DDL execution shape, `staff-analytics-engineer` reviews ELT-readiness from a data-contract consumer perspective. Trigger discipline in the dispatchers prevents three-persona fire on trivial additive diffs.
- Latest revision (commit `644ed74`) is the third round, addressing 30 inline PR comments across two rounds of substantive review plus a follow-on cross-persona check against modern industry framings (dbt Labs / Beauchemin / Sanderson on data contracts / Locally Optimistic) using web research.

## Reviewer roster

| Agent | Role focus |
|---|---|
| `ciso-reviewer` | Threat modeling, auth boundaries (authN / authZ), privilege escalation, OWASP baseline, multi-tenant isolation, TOCTOU, audit log integrity |
| `staff-backend-engineer` | API contracts, idempotency, retry semantics, timeouts / cancellation, observability contract, queue / webhook semantics; AND application data-store schema design (relational + NoSQL): partition keys, GSI / LSI, document shape, single-table vs multi-table, index coverage for app queries |
| `staff-frontend-engineer` | Cache consistency, state management, routing, forms, a11y, Web Vitals, SSR / hydration, client-side analytics SDK emission, internationalization |
| `staff-data-engineer` | Operational data infrastructure across all stores: migration pipeline impact, DDL execution shape, CDC / change-stream config, ETL/ELT pipelines, warehouse ingestion transport, schema-drift detection, catalog / lineage tracking (conditional) |
| `staff-analytics-engineer` (NEW) | Warehouse-side modeling (fact / dim, SCD, partitioning, materialization), transformation correctness, source-schema review for ELT-readiness as a data-contract consumer |
| `staff-platform-engineer` | CI/CD, IaC, shell, deployment ordering + observability coverage, alerting, SLO impact, runbook linkage, load, cost; deploy-window ordering and lock-budget on migrations |
[...]

PR #37, opened 2026-04-24
Expand redaction-gate hook to gh pr; rename to deny-private-project-refs

jcdendrite:
## Summary

Two changes bundled:

1. **Expand the redaction-gate hook to cover `gh pr create` and `gh pr edit`.** The hook previously gated only `git commit`. PR bodies uploaded via `gh pr create --body-file` bypassed the redaction scan entirely — a prior PR in this repo (#34) shipped tracker-ID content in its description because of this gap.

2. **Rename the hook to `deny-private-project-refs.sh` and update the terminology throughout `CLAUDE.md` and `README.md`.** The redaction rule protects against leaking any private project the repo owner works on (client consulting, internal repos, private forks — not only consulting engagements). "Private project" is the neutral umbrella term; "client" narrows the scope unnecessarily.

## Hook changes (`deny-private-project-refs.sh`)

- Extends the fragment-start regex to match `gh pr create` and `gh pr edit` alongside `git commit`.
- Scans `--body` / `--title` inline content, plus `--body-file` and `--template` file contents — including short forms `-F` / `-T` and `=`-delimited variants.
- Fail-closes on unreadable body-source paths.
- Fail-closes on pseudo-file paths (`/dev/stdin`, `/dev/fd/*`, `/proc/*/fd/*`, bare `-`) whose contents the hook cannot statically verify — these would otherwise let gh read different content than the hook scanned.
- Fail-closes on `jq` parse failure, matching the posture of `require-worktree-for-git-writes.sh`.
- Extracts a shared `emit_deny` helper.
- Documents deferred coverage gaps (`--fill` / `-f` / `--fill-verbose` body-from-commits, shell command substitution in inline body/title, and `git commit -F` as a parallel pre-existing gap) in the hook header so the scope boundary is visible to future contributors.

## Rename + terminology

- File rename via `git mv` (blame preserved): `deny-client-refs.sh` → `deny-private-project-refs.sh`.
- `settings.json`: three `if`-dispatch entries now point at the new path, plus matching entries added for `Bash(gh pr create *)` and `Bash(gh pr edit *)`.
- `test_hooks.py`: constant `DENY_CLIENT_REFS_HOOK` → `DENY_PRIVATE_PROJECT_REFS_HOOK`, class `TestDenyClientRefs` → `TestDenyPrivateProjectRefs`, synthetic test prefixes `NULLCLIENT` → `NULLPROJ` and `FAKECLIENT` → `FAKEPROJ`.
- `CLAUDE.md`: section heading `Redact client-identifying content` → `Redact private-project-identifying content`; prose throughout the section, "When a skill is surfaced", "AI agents", and Enforcement sections.
- `README.md`: hook description updated (including the expanded gh pr coverage).
- Skill files that use "client" in the HTTP/API sense (`OAUTH_CLIENT_ID`, client-side rendering, etc.) are unchanged.

## Why no private-project-name blocklist

A committed list of private-project names in this public repo *would itself be the leak* — it would hardcode in cleartext the exact strings the rule is trying to prevent from shipping. A local blocklist file is technically viable but adds a maintenance surface and a second source of truth that can drift. Private-project-name defense stays with the `Redact private-project-identifying content` rule in `CLAUDE.md` and reviewer discipline on all three surfaces (file content, commit message, PR title/description).

This decision is documented in the hook header so future contributors (human or agent) don't re-propose a blocklist without reading the reasoning.

## Note for existing checkouts

After pulling this change, run `stow -R claude` from the repo root to remove the stale `~/.claude/hooks/deny-client-refs.sh` symlink and create the new `~/.claude/hooks/deny-private-project-refs.sh` link. GNU stow doesn't automatically garbage-collect symlinks when their source files are renamed — a plain `git pull` leaves a broken symlink pointing at nothing.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 166 passing (up from 148)
- [x] End-to-end: `gh pr create --body-file <file>` with a synthetic tracker ID in the file → hook denies with the expected message
- [x] End-to-end: each prior-review bypass (`-F`, `--template`, `/dev/stdin`, malformed JSON) → hook denies
- [x] Dry-run the hook against this PR's own staged state → allows (no tracker IDs outside the excluded test directory)
- [ ] `hook-tests` CI passes on this branch

New test cases cover:
- Inline `--body` / `--title` tracker (parametrized across `gh pr create` and `gh pr edit`)
- `--body-file` file content (space and equals forms)
- `-F` short flag (space and equals forms)
- `--template` / `-T` template file content (four form variants)
- `/dev/stdin`, `/dev/fd/*`, `/proc/*/fd/*`, bare `-` pseudo-files → fail-closed
- Nonexistent body-file path → fail-closed
- Allowlisted-only refs in inline body or body-file → pass
- Non-gated `gh` subcommands (`gh pr comment`) pass through
- Scoping: `gh pr` in an unrelated repo is not gated
- Regression lock-ins: amend-message-only allow, equals-form allow, test-dir pathspec exclusion allow
- `jq` fail-closed on malformed JSON input


jcdendrite (commented):
lgtm

PR #38, opened 2026-04-26
Add user-local private-projects blocklist to deny-private-project-refs

jcdendrite:
## Summary

Adds the second mechanical defense to `deny-private-project-refs.sh`: a user-local opt-in blocklist of project names. PR #37 closed the tracker-ID surface across all three write paths (`git commit`, `gh pr create`, `gh pr edit`); this PR catches the named-project category that the tracker-ID regex doesn't.

## Design

The hook reads `~/.claude/private-projects.md` at runtime — a user-local file outside any repo, per-machine, never in `git add`. Each non-comment, non-blank line becomes a case-insensitive literal substring blocklist entry against the same `SCAN_TARGET` the tracker-ID scan inspects (commit message, staged diff added lines, gh pr body/title, body-source file contents).

Tracker-ID matches take priority. Fail-open if the file is absent or unreadable — opt-in protection, not a mandatory checkpoint.

## Why a user-local file (not committed)

PR #37's description rejected a committed blocklist on the grounds that "a list of project names in this public repo would itself be the leak." That objection is correct — and is about *location*, not behavior. The user-local file lives at `~/.claude/private-projects.md`, outside any repo, so the same behavior (literal-substring scan) doesn't reintroduce the leak.

## No-echo invariant (load-bearing)

The blocklist deny message deliberately does NOT name the matched entry. Echoing a name the user explicitly flagged as sensitive would re-expose it in terminal output, screenshots, CI logs, and Claude's conversation context — exactly the surfaces the gate exists to protect. The tracker-ID scan reports matched tokens (defensible — they're mechanical regex matches against the user's own draft), but blocklist entries are user-flagged secrets and stay private even at deny time. A dedicated test locks this in.

## Other changes

- `.gitignore`: belt-and-suspenders for the file landing inside the stowed tree by mistake.
- `README.md`: new "Private-project redaction" section with opt-in instructions, file format, and a worked example explaining what to blocklist (distinctive project names) vs. what NOT to (tracker prefixes — already covered by the regex, prone to false positives on common substrings).
- `install.sh`: tri-state check (absent => tip, empty => warning, populated => silent) using POSIX `[[:space:]]` for macOS portability.
- `CLAUDE.md`: Enforcement section updated to describe both mechanical defenses.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 180 passing (up from 166)
- [x] No-echo invariant: deny message contains neither the matched entry nor any case variant; static "deliberately does not name which entry matched" phrase is asserted so a future refactor that drops the explanation fails fast
- [x] Fail-open: hook allows when blocklist file is absent
- [x] Scoping: blocklist match in unrelated repo short-circuits to allow (same `origin.url` check the tracker-ID scan respects)
- [x] Removing a blocklisted name in the staged diff is allowed (deletion lines aren't scanned, mirroring the tracker-ID `removing_a_tracker_id_is_allowed` test)
- [x] Self-test: the hook caught a tracker-shaped placeholder in this branch's own README worked example before commit. Fixed in the same diff. Satisfying meta-validation that the gate works end-to-end.
- [x] `hook-tests` CI passes on this branch

## New test cases (14)

Block paths: commit message match, gh pr inline body match, gh pr `--body-file` content match, staged-diff added-line match, multi-word entry, lowercase variant, comments-and-blanks-mixed file, whitespace-trimmed entry.

Allow paths: file absent, only-comments-and-blanks file, no-match populated file, unrelated remote scoping, removed-in-diff.

Invariant lock-in: deny message contains no case variant of matched entry AND contains the static explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,13 +103,21 @@ Two scans run, in order:
 ### Opt-in: enable the blocklist
 
 ```bash
-touch ~/.claude/private-projects.md
+# Create the file with a header explaining its purpose (the hook
+# ignores `#` lines, so headers don't affect matching):
+cat > ~/.claude/private-projects.md <<'EOF'
+# Project names listed below are blocked from appearing in any commit
+# message, PR description, or staged content in this repo. Read by
+# ~/.claude/hooks/deny-private-project-refs.sh as a literal blocklist.
+
+EOF
+
+# Append your project names, one per line:
 echo "Acme Corp" >> ~/.claude/private-projects.md
 echo "Project Bluebird" >> ~/.claude/private-projects.md
-
-# Same source of truth for Claude's context — so it self-scrubs while drafting:
-echo "@private-projects.md" >> ~/.claude/CLAUDE.md
 ```
+
+The hook is the load-bearing defense — when it blocks, you (or Claude) sees the deny message and revises before retrying. There's no need to also import the file into `~/.claude/CLAUDE.md`: that would write through the stow symlink to the tracked repo file, which is a footgun without a real defensive payoff.
 
 ### File format
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
 ### Hooks (PreToolUse gates)
 
 - **`require-code-review.sh`** — blocks `git commit` (including chained forms like `git add . && git commit`) until `/code-review` has run on the current staged state. Verified via sha256 marker in `~/.claude/review-markers/<repo-hash>`, which auto-invalidates the moment the staging area changes.
-- **`deny-private-project-refs.sh`** — blocks `git commit`, `gh pr create`, and `gh pr edit` when the staged diff, commit message, or PR title/body/body-source-file contains tracker-ID tokens (`[A-Z]{2,}-\d+`) outside an OSS-prefix allowlist (`CVE-`, `RFC-`, `GH-`, and similar — see the script for the full list). Enforces the tracker-ID category of the repo-root `CLAUDE.md` redaction rule; other categories (private-project names, internal tool names, structural fingerprints) still require review discipline.
+- **`deny-private-project-refs.sh`** — blocks `git commit`, `gh pr create`, and `gh pr edit` when the staged diff, commit message, or PR title/body/body-source-file contains either (a) tracker-ID tokens (`[A-Z]{2,}-\d+`) outside an OSS-prefix allowlist (`CVE-`, `RFC-`, `GH-`, and similar — see the script for the full list), or (b) a literal substring match against entries in the user's opt-in `~/.claude/private-projects.md` blocklist. Enforces the mechanical categories of the repo-root `CLAUDE.md` redaction rule; structural fingerprints still require review discipline. See [Private-project redaction](#private-project-redaction) below.
 - **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a 60-minute bypass marker at `~/.claude/.respond-pr-active` that the skill sets on entry and removes on exit.
 - **`ask-review-permissions.sh`** — asks before `Edit`/`Write`/`MultiEdit` to `.claude/settings*.json`, nudging toward `/review-permissions` when the edit touches `permissions.allow`.
 - **`require-worktree-for-git-writes.sh`** — opt-in per repo. When active, denies non-read-only git operations unless the session runs in a linked git worktree. Prevents concurrent Claude Code sessions from racing on the same working tree. See [Worktree enforcement](#worktree-enforcement) below for opt-in instructions.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -91,6 +91,68 @@ Agents spawned with `isolation: worktree` create their own worktrees under `.cla
...
+
+### Opt-in: enable the blocklist
+
+```bash
+touch ~/.claude/private-projects.md
+echo "Acme Corp" >> ~/.claude/private-projects.md
+echo "Project Bluebird" >> ~/.claude/private-projects.md
+
+# Same source of truth for Claude's context — so it self-scrubs while drafting:
+echo "@private-projects.md" >> ~/.claude/CLAUDE.md
```
jcdendrite: Is this safe to do? Also, won't Claude not have the context of what to do with these keywords with them sitting randomly in CLAUDE.md?

README.md
```diff
@@ -91,6 +91,68 @@ Agents spawned with `isolation: worktree` create their own worktrees under `.cla
...
+❌ **Don't add `ACME`** — the tracker-ID regex already catches `ACME-<digits>`. Adding bare `ACME` to the blocklist would false-positive on the substring `acme` inside English words (`acme of clarity`, `Acme Anvil` references in documentation, etc.). For shorter prefixes like `DAY` the false-positive surface is much worse (`today`, `holiday`, `daylight`).
+
+**Rule of thumb:**
+
+- **Tracker prefixes** (`[A-Z]{2,}` + dash + digits): trust the tracker-ID regex; don't blocklist the bare prefix.
+- **Distinctive project names** (full names, codenames): blocklist them.
+- **Generic-word codenames** (`Pulse`, `Atlas`, `Echo`): blocklist with caution — same false-positive risk as bare prefixes. If the name is a common English word, reviewer discipline may be a better tier than mechanical match.
+
+**Multi-word forms** reduce false-positive risk. If your project is referred to as "AC platform" internally and `AC` alone would false-positive everywhere, blocklist `AC platform` instead — the multi-word phrase is much more selective.
+
```
jcdendrite, line 148: Lots of exposition here. Maybe we should remove this as an issue and instead have the block keywords be case-sensitive and only match on word boundary. I kind of like that idea. Requires much less caveats which can be forgotten and be counter-intuitive

README.md
jcdendrite: **[Claude Code]** Both concerns are valid, and the second one points at a deeper issue I missed.

**Safety:** the `echo >> ~/.claude/CLAUDE.md` is actually broken in this repo's setup. `~/.claude/CLAUDE.md` is a symlink to the stowed `claude/.claude/CLAUDE.md` (the `claude/` stow root), so writing through it modifies the tracked repo file. Anyone following the README would silently stage changes to the public CLAUDE.md.

**Lack of Claude context:** correct — even setting aside the symlink issue, dropping `@private-projects.md` into CLAUDE.md as a bare import gives Claude a context-orphaned list of names with no instruction for what to do with them.

**Fix (pushed in 885e1d2):** drop the @-import advice entirely. The hook is the load-bearing defense — when it blocks, Claude sees the deny message and revises. Pre-emptive self-scrubbing was a nice-to-have, not a primary defense, and it's not worth the symlink-write footgun + context-orphan problem. Replaced the bare `touch + echo` with a `cat > ... <<EOF` that seeds the file with a `#`-commented header explaining the file's purpose (the hook ignores `#` lines, so the header doesn't affect matching).

README.md
jcdendrite: **[Claude Code]** Three options exist, not two — your idea has a hidden cost worth surfacing before I implement.

| Option | Match | Catches | Misses |
|---|---|---|---|
| A. Current — case-insensitive substring | `-iF` | `AcmeCorp`, `acmecorp`, `AcmeCorpService`, `Acme Anvil`, `today` if entry is `DAY` | nothing |
| B. Case-insensitive whole-word | `-iwF` | `AcmeCorp`, `acmecorp`, `ACMECORP` | `AcmeCorpService` (concat), `today` (good — false-positive avoided) |
| C. Your idea — case-sensitive whole-word | `-wF` | `AcmeCorp` only | `acmecorp`, `ACMECORP`, `AcmeCorpService`, `today` (good) |

**Hidden cost in C:** project names appear in different casings depending on context — prose (`AcmeCorp`), code identifiers (`acmecorpClient`, `ACMECORP_API_URL`), kebab paths (`acme-corp`). Case-sensitive whole-word turns one entry into 3-4 entries the user has to maintain. The README's "fewer caveats" win is offset by a "more entries to maintain" cost.

**Hidden cost in B:** loses concatenation matching. `AcmeCorp` doesn't match `AcmeCorpService` because `S` is a word character.

**My recommendation: B (case-insensitive whole-word).** Reasoning:

1. The biggest leak surface is commit messages and PR bodies — free text where the name appears as a standalone word. B catches all of that.
2. The concatenation-miss is real but less common in practice; in *added* diff lines (what the hook scans), the project-name-in-identifier was usually written long ago and isn't being re-introduced.
3. Removes the substring-FP class entirely (today / acme / OPS) without C's multi-casing-entries burden.
4. README rule of thumb collapses to one line: "Add the project name; matched whole-word, case-insensitive."

Implementation cost if you pick B: a single flag in the hook (`-qi -F` → `-qiw -F`), all 14 existing tests still pass (verified by inspection — they all use whitespace boundaries), plus 2-3 new tests locking in the new semantics (`Pulse` doesn't match `impulse`; `AcmeCorp` doesn't match `AcmeCorpService` as documented behavior).

If you'd rather keep A and just fix the DAY example, I'll swap DAY for a clearly-fictitious common-word like `OPS` (matches `stops`, `ops`, `scoops`) or `ART` (matches `start`, `smart`, `Cartesian`) — same lesson, no proximity to anything real.

Which path do you want?

jcdendrite: Let's do Option B and swap DAY for ART

README.md
```diff
@@ -103,14 +103,22 @@ Two scans run, in order:
 ### Opt-in: enable the blocklist
 
 ```bash
-touch ~/.claude/private-projects.md
+# Create the file with a header explaining its purpose (the hook
+# ignores `#` lines, so headers don't affect matching):
+cat > ~/.claude/private-projects.md <<'EOF'
+# Project names listed below are blocked from appearing in any commit
+# message, PR description, or staged content in this repo. Read by
+# ~/.claude/hooks/deny-private-project-refs.sh as a literal blocklist.
```
jcdendrite: This is going to get out of date with the README, which indicates in more detail how the user should think about the keywords (case insensitive word boundary is not obvious). I think this comment should reference the README.

README.md
```diff
@@ -103,14 +103,22 @@ Two scans run, in order:
...
+
+# Append your project names, one per line:
 echo "Acme Corp" >> ~/.claude/private-projects.md
 echo "Project Bluebird" >> ~/.claude/private-projects.md
-
-# Same source of truth for Claude's context — so it self-scrubs while drafting:
-echo "@private-projects.md" >> ~/.claude/CLAUDE.md
 ```
 
+The hook is the load-bearing defense — when it blocks, you (or Claude) sees the deny message and revises before retrying. There's no need to also import the file into `~/.claude/CLAUDE.md`: that would write through the stow symlink to the tracked repo file, which is a footgun without a real defensive payoff.
```
jcdendrite: This is not useful for the README. Maybe it's useful for Claude so you don't recommend this idea again.

README.md
jcdendrite: **[Claude Code]** Done in 9d41495. Hook switched to `-qiw -F` (case-insensitive whole-word). All 14 prior blocklist tests still pass; added 3 new tests locking in B's semantics (substring-within-word does not match, concatenated identifier does not match, punctuation-bordered match still fires). README "What to put / what NOT to" rewritten with `AcmeCorporation` as the ✅ example and `ART` (replacing `DAY`) as the cautionary short-codename example. Total: 183/183 tests passing.

README.md
jcdendrite: **[Claude Code]** Good catch. The seeded file header now points at the README section instead of duplicating the matching rule:

```
# Project names blocked from commits / PR titles / PR bodies in
# claude-config (and forks). Match semantics + what to put in this
# file: see README.md "Private-project redaction" in the
# claude-config repo.
```

That keeps the file self-explanatory enough for someone reading it cold, but moves the durable "how matching works" content to one place that does not drift. Updated in 9d41495.

README.md
jcdendrite: **[Claude Code]** Right — removed from the README in 9d41495. Captured the lesson in user-level memory at `~/.claude/projects/-home-jared-MyCode-claude-config/memory/project_claude_md_is_stow_symlinked.md` (indexed in MEMORY.md) so future Claude knows that `~/.claude/CLAUDE.md` is symlinked to a tracked repo file in this setup, never recommends `>>` writes through it, and reaches for the committed CLAUDE.md or hook-based defense instead. The `@-import` pattern itself is fine in general — the trap is specifically the symlink in this stow setup, which the memory makes explicit.

jcdendrite: Yeah I don't think memory is the right place for it. I think it should go in the CLAUDE.md for this repo specifically. Check the ai-instructions skill

jcdendrite (commented):
lgtm

README.md
jcdendrite: **[Claude Code]** You're right — moved to the committed CLAUDE.md in 2179788 (under "Working in this repo", next to the existing stow note). Removed the user-memory entry I created earlier. CLAUDE.md is the correct surface per `ai-instruction-and-memory-files`: stable architectural facts that apply to any agent working in any clone of this repo belong in user-written instructions, not Claude-written ephemeral memory.

jcdendrite (commented):
lgtm

PR #42, opened 2026-04-27
respond-pr: add --paginate to comment-fetch commands

jcdendrite:
## Summary

- Add `--paginate` to all three `gh api` calls in the `respond-pr` skill (inline file comments, top-level review comments, issue-level comments). Without it, GitHub's REST API caps at 30 results per page and silently truncates.
- Expand the existing warning to name both failure modes Claude hits: (a) fetching only one of the three comment types; (b) fetching without `--paginate`. They're distinct and both produce reviews that look complete but miss real feedback.

Surfaced when an audit of a PR with 34 inline comments (PR #34) found that an un-paginated fetch had returned only 30, and 4 real review comments had been missed.

## Test plan

- [x] After merge, install via `git pull` (skill is stowed — no re-install needed)
- [ ] Spot-check on a public PR with >30 comments of one type, e.g. `gh api repos/cli/cli/pulls/<long-thread-PR>/comments --paginate | jq 'length'` — should exceed 30
- [ ] Next `respond-pr` invocation against a PR with many comments returns the full set


jcdendrite (commented):
lgtm

PR #43, opened 2026-04-27
ai-instruction-and-memory-files: drop .lovable/*.md auto-load implication

jcdendrite:
## Summary

The skill's TRIGGER list mentioned `.lovable/*.md` without qualification, implying that Lovable auto-loads those files as project knowledge. It does not — `.lovable/*.md` is repo-tracked human reference; knowledge becomes visible to Lovable per session only by adding it through the Lovable UI's project-knowledge editor.

In `claude/.claude/skills/ai-instruction-and-memory-files/SKILL.md`:

- **§2** adds a 2-line bold callout stating `.lovable/*.md` is NOT loaded by Lovable, with a back-reference to the source-1 entry in the §2 list (the UI editor).
- **§3 "Don't embed PR or ticket refs"** is tightened from 7 lines to 5 — same point, less padding, stronger imperative ("state the rule, not the precedent"). Bundled scope; the section is the closest existing demonstration of the very rule it embodies. Placeholder also swapped to `TICKET-NNN` so the example doesn't itself match the redaction hook's tracker-ID pattern when staged.

The corrected `.lovable/*.md` fact also lives in §2's existing four-source list (which already correctly excluded `.lovable/*.md`); this PR makes the exclusion explicit instead of leaving it as an inference for the reader.

## Test plan

- [ ] Read the diff against the skill's own length-target / prose-vs-structural rules (skill-on-own-diff check). 197 → 198 lines, under the 200 soft target.
- [ ] Verify the new §2 callout doesn't introduce a misconception (e.g., "project-knowledge UI field is durable") — earlier draft had this risk; current version drops the risky sentence entirely.
- [ ] Verify TRIGGER still fires for `.lovable/*.md` editing (the trigger phrase is unchanged).
- [ ] Verify §3 trim preserves all load-bearing claims (rule, rotting failure mode, where they belong, takeaway).

### DIFF
--- a/claude/.claude/skills/ai-instruction-and-memory-files/SKILL.md
+++ b/claude/.claude/skills/ai-instruction-and-memory-files/SKILL.md
@@ -7,9 +7,8 @@ description: >
   index + topic files): precedence, duplication rules, length targets,
   the `@AGENTS.md` import pattern Anthropic officially endorses, and the
   split between user-written instructions and Claude-written memory.
-  TRIGGER when: editing CLAUDE.md or AGENTS.md, editing `.lovable/*.md`
-  (NOT auto-loaded by Lovable — see §2), creating a new instruction file,
-  creating, editing, auditing, or
+  TRIGGER when: editing CLAUDE.md or AGENTS.md, editing `.lovable/*.md`,
+  creating a new instruction file, creating, editing, auditing, or
   pruning Claude Code auto-memory in `~/.claude/projects/*/memory/`,
   deciding whether a rule belongs in CLAUDE.md vs auto-memory,
   evaluating whether rules should be duplicated across files, or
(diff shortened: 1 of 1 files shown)


claude/.claude/skills/ai-instruction-and-memory-files/SKILL.md
```diff
@@ -83,6 +83,9 @@ All four are loaded every session. Lovable docs warn that in very long
 conversations instructions can drift; the "always read" guarantee for
 AGENTS.md is the defense-in-depth.
 
+**`.lovable/*.md` is NOT loaded by Lovable** — these are repo-tracked
+notes for humans; visible knowledge comes only via the UI editor (source 1 above).
```
jcdendrite: repo-tracked `copies` for humans or something to that effect. This more terse update doesn't indicate that the latest .md files in the repo are supposed to match exactly the ones saved in UI editor.

jcdendrite: **[Claude Code]** Good catch — "notes for humans" lost the mirror semantic. Updated in d872f9e to "these files are repo-tracked copies of what's set in the UI editor (source 1 above); only the UI version loads." Same 2 lines; file still at 198. Used "copies" (your word) to keep it explicit that the file should match the UI exactly, not be free-form.

jcdendrite (commented):
lgtm

PR #44, opened 2026-04-27
respond-pr: per-session bypass marker keyed by session_id

jcdendrite:
## Summary

The `respond-pr` skill's bypass marker was a singleton at `~/.claude/.respond-pr-active`. Two parallel-session failure modes:

1. **Cleanup thrash** — session A's `rm -f` deletes the marker session B is still using; B's subsequent `gh` calls get blocked mid-skill.
2. **Bypass leak (silent)** — while *any* session held the marker, every other session's matching `gh` calls bypassed the gate. Unrelated sessions silently skipped the skill, fetching only inline comments (missing top-level reviews and issue-level comments) and posting replies without the `[Claude Code]` attribution prefix.

This PR replaces the singleton with per-session keying tied to the hook payload's documented `session_id` field.

## Design

- **Marker path:** `~/.claude/.respond-pr-active.d/<session_id>` (one file per session).
- **Hook:** `require-respond-pr.sh` reads `session_id` from its JSON input, bypasses only if *this session's* marker exists and is fresh (<60 min). Refreshes the marker's mtime on each bypass so long skill runs don't hit the staleness cutoff mid-execution.
- **Skill:** Step 0 looks up the session's `session_id` and writes the per-session marker. Step 6 removes only its own marker.
- **Bootstrap:** new SessionStart hook `capture-session-id.sh` writes `~/.claude/sessions/<claude-pid>` at session start so the skill (running as Bash tool calls without access to the hook payload) can look up its own `session_id` via the bash tool's `$PPID` (which is the claude main process directly).

## Why session_id, not PPID

Initially considered a `$PPID`-based scheme. An empirical probe of the hook process tree disproved the assumption it relied on: the hook is invoked through a transient `sh` shim, so its `$PPID` is that shim, not `claude`. The bash tool's `$PPID` *is* `claude`. They are not the same process. `session_id` from the JSON payload is the documented harness contract and is schema-stable; the probe confirmed it's reliably populated.

## Test plan

192 tests, all passing. New coverage in `claude/.claude/hooks/tests/test_hooks.py`:

- [x] `test_other_sessions_marker_does_not_leak_bypass` — regression test for the silent-bypass-leak failure mode (session A's marker must not authorize session B's gated calls).
- [x] `test_no_session_id_in_input_denies` — empty `session_id` in payload → deny, even if marker files exist.
- [x] `test_marker_dir_missing_denies` — sessions dir absent → deny.
- [x] `test_bypass_refreshes_marker_mtime` — long-run mitigation: hook touches the marker on bypass.
- [x] `TestCaptureSessionId` — happy path, empty `session_id`, missing field, malformed JSON, empty stdin (all must exit 0; SessionStart must never block startup).

Manual smoke (post-merge, after `git pull` to pick up the SessionStart hook):
- [ ] Run `/respond-pr` against a real PR; verify it completes cleanly.
- [ ] Open two Claude sessions in parallel; confirm cleanup thrash is gone.
- [ ] Verify a third session's `gh api .../pulls/N/comments` is *blocked* while the other two have active markers (regression check for the leak).

## Migration

Existing users can `rm -f ~/.claude/.respond-pr-active` after this lands — the singleton path is no longer read. Harmless if left in place; just an orphan file.

## Out of scope (follow-ups)

The `code-review` marker at `~/.claude/review-markers/<repo-hash>` has its own narrower parallel-session concern (two sessions in the same worktree staging different diffs collide). Different shape from this fix; tracked for a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,11 +20,12 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
 
 ## What this installs
 
-### Hooks (PreToolUse gates)
+### Hooks
 
 - **`require-code-review.sh`** — blocks `git commit` (including chained forms like `git add . && git commit`) until `/code-review` has run on the current staged state. Verified via sha256 marker in `~/.claude/review-markers/<repo-hash>`, which auto-invalidates the moment the staging area changes.
 - **`deny-private-project-refs.sh`** — blocks `git commit`, `gh pr create`, and `gh pr edit` when the staged diff, commit message, or PR title/body/body-source-file contains either (a) tracker-ID tokens (`[A-Z]{2,}-\d+`) outside an OSS-prefix allowlist (`CVE-`, `RFC-`, `GH-`, and similar — see the script for the full list), or (b) a literal substring match against entries in the user's opt-in `~/.claude/private-projects.md` blocklist. Enforces the mechanical categories of the repo-root `CLAUDE.md` redaction rule; structural fingerprints still require review discipline. See [Private-project redaction](#private-project-redaction) below.
-- **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a 60-minute bypass marker at `~/.claude/.respond-pr-active` that the skill sets on entry and removes on exit.
+- **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a per-session bypass marker at `~/.claude/.respond-pr-active.d/<session_id>` that the skill sets on entry and removes on exit; the hook refreshes the marker's mtime on each bypass so long skill runs don't hit the 60-minute staleness cutoff. Per-session keying (rather than a singleton path) keeps parallel Claude sessions from thrashing on cleanup or leaking bypass to unrelated sessions.
+- **`capture-session-id.sh`** (SessionStart) — at session start, writes the session's `session_id` to `~/.claude/sessions/<claude-pid>` so skills running as Bash tool calls (which don't see the hook payload) can look up their own session id via the bash tool's `$PPID`. Used by `/respond-pr` to compute its bypass-marker filename.
 - **`ask-review-permissions.sh`** — asks before `Edit`/`Write`/`MultiEdit` to `.claude/settings*.json`, nudging toward `/review-permissions` when the edit touches `permissions.allow`.
 - **`require-worktree-for-git-writes.sh`** — opt-in per repo. When active, denies non-read-only git operations unless the session runs in a linked git worktree. Prevents concurrent Claude Code sessions from racing on the same working tree. See [Worktree enforcement](#worktree-enforcement) below for opt-in instructions.
 
--- a/claude/.claude/hooks/capture-session-id.sh
+++ b/claude/.claude/hooks/capture-session-id.sh
@@ -19,21 +19,40 @@
 #   that shim, not claude. claude is the shim's parent. POSIX
 #   `ps -o ppid= -p $PPID` walks up one level. Works on Linux, macOS, WSL.
 #
-# Failure mode: every step exits 0 silently. If session_id or claude_pid
-# can't be determined, the lookup file isn't written; the skill will then
-# fail loudly when it tries to read it. A SessionStart hook must never
-# block session startup, even on internal error.
+# Failure mode: every step exits 0 (a SessionStart hook that fails-closed
+# would block session startup, which is worse than a delayed Step 0
+# failure in the skill). Each failure path emits a one-line diagnostic to
+# stderr — visible in the user's terminal, not added to Claude's context
+# (which is stdout). When the lookup file isn't written, the /respond-pr
+# skill's Step 0 fails loudly with a clear message; the stderr trail here
+# is the upstream signal explaining why.
 
 INPUT=$(cat 2>/dev/null)
-[ -z "$INPUT" ] && exit 0
+if [ -z "$INPUT" ]; then
+  echo "[capture-session-id] empty stdin; lookup file not written; respond-pr skill will fail at Step 0" >&2
+  exit 0
+fi
 
 SESSION_ID=$(printf '%s\n' "$INPUT" | jq -r '.session_id // empty' 2>/dev/null)
-[ -z "$SESSION_ID" ] && exit 0
+if [ -z "$SESSION_ID" ]; then
+  echo "[capture-session-id] no session_id in payload; respond-pr skill will fail at Step 0" >&2
+  exit 0
+fi
 
 CLAUDE_PID=$(ps -o ppid= -p "$PPID" 2>/dev/null | tr -d ' ')
-[ -z "$CLAUDE_PID" ] && exit 0
+if [ -z "$CLAUDE_PID" ]; then
+  echo "[capture-session-id] could not resolve claude PID via 'ps -o ppid= -p $PPID'; respond-pr skill will fail at Step 0" >&2
+  exit 0
+fi
 
-mkdir -p "$HOME/.claude/sessions" 2>/dev/null || exit 0
-printf '%s\n' "$SESSION_ID" > "$HOME/.claude/sessions/$CLAUDE_PID" 2>/dev/null
+if ! mkdir -p "$HOME/.claude/sessions" 2>/dev/null; then
+  echo "[capture-session-id] could not create $HOME/.claude/sessions; respond-pr skill will fail at Step 0" >&2
+  exit 0
+fi
+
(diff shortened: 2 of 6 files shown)


claude/.claude/hooks/tests/test_hooks.py
```diff
@@ -366,6 +459,71 @@ def test_ssh_origin_same_repo_denied(self, isolated_home, tmp_path):
...
+            input="",
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        assert result.returncode == 0
+        assert self._sessions_files(isolated_home) == []
+
+    def test_malformed_json_does_not_block(self, isolated_home):
+        """SessionStart hook must never fail-closed on payload corruption —
```
jcdendrite, line 528: Is there a warning if the hook fails? Otherwise we have silent thrashing, which is exactly what this PR is trying to fix. Maybe it should fail-closed?

claude/.claude/hooks/capture-session-id.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+# Deriving claude_pid from inside this hook:
+#   This hook is invoked through a transient `sh` shim, so its $PPID is
+#   that shim, not claude. claude is the shim's parent. POSIX
+#   `ps -o ppid= -p $PPID` walks up one level. Works on Linux, macOS, WSL.
+#
+# Failure mode: every step exits 0 silently. If session_id or claude_pid
+# can't be determined, the lookup file isn't written; the skill will then
+# fail loudly when it tries to read it. A SessionStart hook must never
+# block session startup, even on internal error.
```
jcdendrite: Related to my comment earlier. If the skill fails loudly, won't Claude try to work around the failure and cause thrashing or bypass the marker entirely?

claude/.claude/hooks/require-respond-pr.sh
```diff
@@ -7,12 +7,19 @@
 # skill fetches all three AND enforces the [Claude Code] attribution prefix
 # on replies.
 #
-# Bypass: the /respond-pr skill touches ~/.claude/.respond-pr-active at its
-# start and removes it at the end. While the marker exists AND is fresh
-# (<60 min old), this hook lets gh commands through so the skill itself
-# doesn't recurse into its own gate. The 60-minute staleness cutoff
-# prevents an orphaned marker from permanently disabling the gate if the
-# skill errored out mid-execution.
+# Bypass: the /respond-pr skill writes a marker at
```
jcdendrite, line 10: Just realized - will this entire plan work when the claude session spawns agents, or will they all be thrashing on the same marker? Do spawned agents have different session Ids from the initializing json?

claude/.claude/skills/respond-pr/SKILL.md
```diff
@@ -23,7 +27,11 @@ Fetch all review comments on the current branch's open pull request and address
    - If it's a question or discussion point, draft a clear response
 4. Post replies using the GitHub API with `in_reply_to` (use `-F` for integer IDs)
 5. Commit and push any code changes in a single commit
-6. **Remove the hook bypass marker:** `rm -f ~/.claude/.respond-pr-active`. If the skill errors out before reaching this step, don't manually clean up — the hook's 60-minute staleness cutoff handles orphaned markers automatically.
+6. **Remove this session's hook bypass marker:**
+   ```
+   SESSION_ID=$(cat "$HOME/.claude/sessions/$PPID" 2>/dev/null) && [ -n "$SESSION_ID" ] && rm -f "$HOME/.claude/.respond-pr-active.d/$SESSION_ID"
+   ```
+   Removes only this session's file. If the skill errors out before reaching this step, don't manually clean up — the hook's 60-minute staleness cutoff handles the orphan automatically.
```
jcdendrite, line 34: A reviewer agent said the 60-minute staleness cutoff was quite long, if I recall correctly. Reasons why you kept it as is?

README.md
```diff
@@ -20,11 +20,12 @@ This symlinks `claude/.claude/` into `$HOME/.claude/`.
...
-- **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a 60-minute bypass marker at `~/.claude/.respond-pr-active` that the skill sets on entry and removes on exit.
+- **`require-respond-pr.sh`** — blocks PR comment reads and posts (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and redirects to `/respond-pr`, so all three comment types get fetched and replies carry the `[Claude Code]` attribution prefix. Honors a per-session bypass marker at `~/.claude/.respond-pr-active.d/<session_id>` that the skill sets on entry and removes on exit; the hook refreshes the marker's mtime on each bypass so long skill runs don't hit the 60-minute staleness cutoff. Per-session keying (rather than a singleton path) keeps parallel Claude sessions from thrashing on cleanup or leaking bypass to unrelated sessions.
```
jcdendrite, line 27: I just realized - why would this skill run for 60 min? It's just responding to a comment. Also, why do we have a marker whatsoever? Why don't we have PR responses always route to the respond-pr skill? What am I missing?

claude/.claude/hooks/tests/test_hooks.py
jcdendrite: **[Claude Code]** Fair — pushed fixup `1f229e3` adding stderr diagnostics on every failure path (empty stdin, missing `session_id`, `ps` walk-up failure, `mkdir`/write failure). Stderr keeps the trail out of Claude's context (stdout would pollute it) but visible in the user's terminal where hook debugging belongs. Still `exit 0` because fail-closing a SessionStart hook would block session startup, which is worse than a delayed-but-clear Step 0 failure.

claude/.claude/hooks/capture-session-id.sh
jcdendrite: **[Claude Code]** Step 0's error message names the missing piece explicitly ("session bootstrap missing — `capture-session-id.sh` did not run; do not proceed"). Claude has no bypass route that succeeds: the regression test confirms a different `session_id`'s marker can't authorize this session, and the gate's only valid bypass is a fresh marker keyed by THIS `session_id`, which only the SessionStart hook can bootstrap. So Claude either (a) gets the SessionStart hook fixed and re-runs, or (b) reports to user. No workaround thrash because no workaround works. With the new stderr diagnostics in `1f229e3`, the upstream signal arrives earlier.

claude/.claude/hooks/require-respond-pr.sh
jcdendrite: **[Claude Code]** Honest answer: I haven't empirically verified subagent `session_id` behavior. Top-level parallel sessions get distinct `session_id`s — confirmed during the design probe when a parallel session showed up in the hook log with its own id. Agents spawned via the Agent tool likely share the parent's `session_id` since they run inside the same `claude` process, but that's a guess. If shared: parent and subagents bypass via the same marker — fine. The corner case is **nested `/respond-pr`**: a subagent that itself invokes the skill writes the same marker as the parent, then Step 6 cleanup deletes it — the exact thrash this PR fixes for parallel sessions, recurring nested. In practice subagents rarely run `/respond-pr` (it's a top-level user-driven action). The robust fix is keying the marker by `tool_use_id` (per-invocation, also in the hook payload) instead of just `session_id` — worth a follow-up if subagent-driven `/respond-pr` ever becomes a real pattern. Out of scope for this PR.

claude/.claude/skills/respond-pr/SKILL.md
jcdendrite: **[Claude Code]** The 60-min number is doing two jobs after this PR:

1. **Long-skill mitigation.** The reviewer's "60-min cutoff is a silent footgun for long runs" concern is now handled by the hook touching the marker on every bypass — the marker stays fresh as long as gated `gh` calls keep happening, so a long `/respond-pr` won't hit the cutoff mid-run.
2. **Orphan recovery.** If the skill crashes before Step 6, the marker becomes a leak for up to 60 min. Per-session keying contains the leak to its own `session_id`; a session-resume after crash actually finds its own fresh marker and continues — useful, not a leak.

The number itself was inherited from the singleton design and not re-justified here. Could be tightened (30 min?) for faster orphan forgetting, but no concrete pressure given the long-run footgun was the actual motivating concern and is mitigated separately.

README.md
jcdendrite: **[Claude Code]** Three questions, all pointing at the same core mechanism:

**"Why don't PR responses always route to `/respond-pr`?"** That's exactly what `require-respond-pr.sh` does. It blocks every PR-comment op (`gh api .../pulls|issues/N/{comments,reviews}`, `gh pr comment`, `gh pr review`) and tells Claude to run the skill. The whole point of the gate is forced routing.

**"Why a marker at all?"** Because the skill itself needs to make those exact ops — it has to fetch the comments and post replies. Without a bypass, the skill would recurse into its own gate. The marker is the SKILL's signal saying "I'm running, my own `gh` calls are authorized."

**"Why would the skill run for 60 min?"** Heavy reviews. 30 substantive comments at ~30 sec each (read file/line, draft attribution-prefixed reply, post) is 15 min; 100+ comments with code-context lookups can hit an hour. The mtime refresh keeps the marker alive across long runs without needing a longer hard cutoff.

jcdendrite (commented):
lgtm

PR #45, opened 2026-04-27
Reserve PROJ- and TICKET- as placeholder tracker prefixes in redaction gate

jcdendrite:
## Summary

- Adds two prefixes to `deny-private-project-refs.sh`'s `OSS_ALLOWLIST` so tracker-ID-shaped placeholders in skill examples, docs, and commit messages don't have to be obfuscated (e.g. to all-letter forms) to defeat the redaction scan. Obfuscation defeats the placeholder's purpose — the example is supposed to show the *shape* of a real tracker ID.
- Adds positive tests (placeholder tokens are now allowed) and a negative regression test that locks in the `^` anchor's protection against prefix substrings — tokens that share a tail with an allowlisted prefix but don't start with it must still deny. Without this test, a future regex refactor that drops the anchor would pass CI silently.
- Updates the repo-root `CLAUDE.md` "Redact private-project-identifying content" section and the `README.md` "Private-project redaction" section to document the new placeholder convention.

## Tradeoff

A real private project that uses one of these reserved prefixes as its actual tracker prefix would no longer be caught by the mechanical scan. The user-local `~/.claude/private-projects.md` blocklist remains the fallback for that case — adding the project name to that file catches surrounding context. Both reserved prefixes are conventionally placeholder shapes and unlikely to be a real tracker key.

## Test plan

- [x] `python3 -m pytest test_hooks.py` — 189 passed (was 185 on main; +2 positive placeholder cases, +4 negative substring cases).
- [x] Specialist reviews (security + hook correctness) ran; one BLOCKER raised (missing negative test) and one comment-budget NIT — both addressed before this PR.
- [ ] After merge: PR #43's obfuscated placeholder can be swapped back to a real-shaped tracker ID, since the scan will accept it once this lands.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -52,9 +52,7 @@ specific private project, engagement, or private codebase. Categories:
   Default: if in doubt, strip it. When an example or commit message
   needs a tracker-ID-shaped placeholder, use `PROJ-<digits>` or
   `TICKET-<digits>` — both prefixes are reserved as placeholders
-  and pass the allowlist with any digit suffix. Don't obfuscate
-  the digits (e.g. to `NNN`); the placeholder is supposed to show
-  the shape of a real tracker ID, which obfuscation hides.
+  and pass the allowlist with any digit suffix.
 - **Internal URLs, hostnames, Slack channels, project domains**.
 - **Absolute filesystem paths** that embed project names
   (`~/Code/acme-platform/...`, `/home/foo/WorkForProject/...`).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -49,7 +49,12 @@ specific private project, engagement, or private codebase. Categories:
   `[A-Z]{2,}-\d+` that is *not* on this allowlist of standard
   open-source references: `CVE-`, `RFC-`, `PEP-`, `ISO-`, `GH-`,
   `BUG-` / bugzilla-style, and clearly-public project prefixes.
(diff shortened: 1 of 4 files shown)


CLAUDE.md
```diff
@@ -49,7 +49,12 @@ specific private project, engagement, or private codebase. Categories:
   `[A-Z]{2,}-\d+` that is *not* on this allowlist of standard
   open-source references: `CVE-`, `RFC-`, `PEP-`, `ISO-`, `GH-`,
   `BUG-` / bugzilla-style, and clearly-public project prefixes.
-  Default: if in doubt, strip it.
+  Default: if in doubt, strip it. When an example or commit message
+  needs a tracker-ID-shaped placeholder, use `PROJ-<digits>` or
+  `TICKET-<digits>` — both prefixes are reserved as placeholders
+  and pass the allowlist with any digit suffix. Don't obfuscate
```
jcdendrite: With the carveouts for PROJ-digits and TICKET-digits, I think we can trim this line "Don't obfuscate
  the digits (e.g. to `NNN`); the placeholder is supposed to show
  the shape of a real tracker ID, which obfuscation hides."

jcdendrite: **[Claude Code]** Trimmed in 6cda40b — agreed, the `<digits>` form already shows the rule.

jcdendrite (commented):
lgtm

PR #50, opened 2026-04-27
code-review: per-session marker keyed by repo-hash + session_id

jcdendrite:
## Summary

The `/code-review` marker at `~/.claude/review-markers/<repo-hash>` is keyed only by repo top-level path. Two parallel Claude sessions in the same worktree staging different diffs collide on this singleton: last writer wins, the other session's commit is denied at the gate. Worktrees of the same repo are already disambiguated by their distinct top-level paths, so this only bites when sessions actually share a worktree.

This was the second unsafe singleton flagged in the audit that surfaced from #44; per-session keying mirrors the pattern used there.

## Design

- **Marker path:** `~/.claude/review-markers/<repo-hash>.<session_id>` (flat namespace, single dot delimiter).
- **Hook:** `require-code-review.sh` reads `session_id` from its JSON payload and checks THIS session's marker. Empty `session_id` (older Claude Code or payload-schema drift) falls through to deny rather than silently bypassing.
- **Skill:** the "Record review completion" command in `code-review/SKILL.md` looks up its own `session_id` via the same `~/.claude/sessions/$PPID` lookup file that the SessionStart hook (`capture-session-id.sh`, shipped in #44) writes at startup. Chain fails loudly if the lookup file is missing.
- **Path format choice:** flat-with-dot (`<repo-hash>.<session_id>`) over nested dirs (`<repo-hash>/<session_id>`) to avoid `mkdir` collision with old-format singleton FILES at `<repo-hash>` — those orphan harmlessly under the new layout because the new filename pattern doesn't conflict.

## Migration

After this merges and you `git pull`, any in-flight `/code-review` marker recorded under the old singleton path becomes inert. The hook will deny on the next commit attempt; re-run `/code-review` once on the staged diff, and the new path populates.

Old singleton files orphan harmlessly. Optional cleanup:

```
find ~/.claude/review-markers -maxdepth 1 -type f ! -name '*.*' -delete
```

(Matches files without a dot in the name — i.e., the old `<repo-hash>` singletons. New files have `<repo-hash>.<session_id>` shape and are skipped.)

## Test plan

201 tests pass (was 199 before; added 2 new). Coverage:

- [x] **`test_other_sessions_marker_does_not_authorize_commit`** — load-bearing regression: session A writes a marker for hashA; session B stages a diff with the same hashA; B's commit must still deny because the gate is per-session, not per-diff.
- [x] **`test_no_session_id_in_input_denies`** — empty `session_id` in the hook payload → deny even when a marker file exists.
- [x] **`test_skill_marker_write_command_matches_hook_path`** — runs the SKILL.md "Record review completion" command verbatim (with the new `SESSION_ID` lookup), then verifies the hook accepts the result. Cross-file regression guard.
- [x] Existing marker-check tests threaded with `session_id` (no_marker, wrong_hash, correct_hash, chained_add_commit, restaging, refreshed). Early-exit tests (non-bash, non-commit, outside-repo, empty staged diff) deliberately don't pass `session_id` — the hook returns before reaching marker logic.

Manual smoke (post-merge):
- [ ] Run `/code-review` on a real change, verify it commits cleanly.
- [ ] Open two Claude sessions in the same worktree, stage different diffs, run `/code-review` in each, verify both can commit.

## Out of scope

- The `/code-review` skill body itself (the checklist content). Unchanged.
- The hash-of-staged-diff comparison logic. Unchanged.
- Garbage collection of stale per-session marker files. Negligible at realistic usage (same conclusion the senior sysadmin reviewer reached for the respond-pr marker dir in #44).


jcdendrite (commented):
lgtm

PR #51, opened 2026-04-27
skills + agents: reduce overlap; tighten shared concerns

jcdendrite:
## Summary

Follow-up to #34 (specifically Q4 of the third-round review:
"Code-review and plan-review skills overlap with each other and with
agents"). Addresses the two structural overlap sources #34 deferred.

**Plan-review-vs-code-review duplication.** Items that named the same
concern in plan-shape vs code-shape were near-verbatim restatements
under different ID prefixes (`D2` ≡ `18`, `D5` ≈ `19`, `D4` ≈ `21`,
`S3` ≈ `26`, etc.). Tightened to one-liners in both skills with
distinct framings: plan-review uses plan-shape questions ("Does the
plan describe..."), code-review uses code-shape facts ("Destructive
operations need a backup path"). The Item ownership table is the
cross-reference. Approach: Option C from the handoff — accept some
duplication, but minimize it.

**Skill-item-vs-agent-angle restatement.** Pruned each agent's "Core
review angles" section to keep only perspectives the LLM needs to
hook attention onto at spawn time. Dropped pure restatements;
retained one-line anchors for angles an agent might miss when invoked
outside the skill dispatcher.

**Specialist self-review:** ran `/code-review` with all 8 reviewer
subagents. Findings filed and addressed:

| Specialist | Finding | Disposition |
|------------|---------|-------------|
| ciso | Defense in depth dropped | Restored as one-liner |
| platform | concurrency / env-parity / secret-scoping / idempotency anchors | Restored as one-liners |
| backend | per-request work shape (N+1 / sync-in-handler) anchor | Restored as one-liner |
| frontend | async cancellation / `useEffect` race / `AbortController` (pre-existing gap) | New angle added |
| product | D3 user-cohort phrasing | Restored ("failures users hit mid-deploy") |
| data, analytics, sdet | No concerns | — |

**Preserved from #34:** 8 personas, three-way schema review, no
Trigger-discipline section, single-phrase doc-inclusion mention,
pre-emptive vs retrospective scope asymmetry, item-ownership table
shape.

**Net:** trims duplication while preserving every spawn-time
attention hook the specialists identified as needed; closes one
pre-existing gap (frontend async cancellation) surfaced during
self-review.

## Test plan

- [ ] Item ownership tables in both skills still route every item to
      a primary owner; no orphaned personas
- [ ] All checklist item titles in skill bodies still match their
      Item ownership table entries (post-tightening)
- [ ] Each agent's Core review angles still names enough perspectives
      for spawn-time attention even without the skill body loaded
- [ ] On a small staged diff, `/code-review` still fires specialists
      and they file findings using their own perspectives, not by
      walking every numbered item

### DIFF
--- a/claude/.claude/agents/ciso-reviewer.md
+++ b/claude/.claude/agents/ciso-reviewer.md
@@ -36,9 +36,7 @@ If the change is bounded to cosmetic-only edits (typo fixes, formatting, copy po
 
 **Secret lifecycle** — new tokens/keys: provisioning, rotation, revocation. Logged anywhere (full request bodies, error payloads, audit tables)?
 
-**Defense in depth** — more than one layer of enforcement. "The policy handles it" without an in-code check is single-layer.
-
-**Error messages** — leak internal IDs, stack traces, query text, existence-of-record signals an attacker could enumerate.
+**Defense in depth** — flag single-layer designs even when the one layer is correct. "RLS handles it" without an in-code check, or an in-code check without a policy backstop, is a finding regardless of whether the layer present is sound.
 
 **Audit log integrity** — for security-relevant events (privileged actions, access grants, secret changes): log entry written, immutable, tamper-evident.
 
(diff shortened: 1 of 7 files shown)


jcdendrite (commented):
lgtm

PR #54, opened 2026-04-28
tests: parse hook-alignment recipes from SKILL.md instead of hardcoding

jcdendrite:
## Summary

Closes the gap discussed in the prior session: the existing code-review hook ↔ skill alignment test embedded a hardcoded copy of the SKILL.md marker-write recipe inline. If SKILL.md drifted from what the hook expected, the test still passed — it was testing its own copy, not the file. Same shape applies to respond-pr's enable/disable bypass commands, which had no alignment test at all.

This PR makes SKILL.md the single source of truth for the alignment tests by tagging the fenced blocks with HTML-comment markers and reading them at test time.

## Changes

- **`extract_skill_command(skill_path, fixture_id)`** in `test_hooks.py`. Reads SKILL.md, finds the `<!-- HOOK_TEST_FIXTURE: <id> -->` marker, returns the contents of the immediately-following fenced block. Asserts each fixture id resolves to exactly one block (catches accidental copy-paste duplication).
- **Three fixture markers added:**
  - `code-review/SKILL.md` → `marker-write`
  - `respond-pr/SKILL.md` → `enable-bypass`, `disable-bypass`

  HTML comments are invisible in rendered markdown but searchable in source. Each marker has a self-documenting note explaining the test contract.
- **Existing code-review test refactored** to read from SKILL.md instead of holding an inline copy. Also adds a sanity assertion that the marker landed at the path the hook computes — failures now localize to skill-side (wrong write path) vs hook-side (wrong read path).
- **Two new respond-pr tests** that execute the enable / disable recipes from SKILL.md and verify (a) the marker file is created/removed at the expected path, and (b) the hook flips between deny and allow accordingly.

## Notes on the parser

The regex tolerates indented closing fences. The respond-pr fixtures sit inside numbered list items (`0. **Enable hook bypass.**`, `6. **Remove this session's hook bypass marker:**`), so their fences carry the list's 3-space indent. Initial cut of the regex required column-0 closing fences and ran greedily across half the file — fixed by making the closing fence tolerate `[ \t]*` indent.

## Drift scenarios this now catches

- SKILL.md path swap (e.g., `$REPO_HASH.$SESSION_ID` → `$SESSION_ID.$REPO_HASH`): the file-existence assertion fires before the hook check, so the failure message names the marker path mismatch directly.
- SKILL.md write-method drift (e.g., recipe uses `mkdir` and `touch` but the file lands somewhere unexpected because of a quoting bug): the hook check fires.
- Recipe deletion or renaming: the parser raises `AssertionError` naming the missing fixture id.
- Duplicate fixture id (copy-paste mistake): parser raises `AssertionError` naming the duplication.

## Out of scope

- A test that asserts the SKILL recipe's hash matches a Python re-derivation. The reviewer suggested this; I dropped it because the canonical digest IS `git diff --cached | sha256sum` — re-deriving in Python tests bash, not skill-vs-hook alignment.
- Backporting the parser to other skill ↔ hook pairs (none exist beyond these two).

## Test plan

- [x] `pytest claude/.claude/hooks/tests/test_hooks.py` → 203 passed
- [x] Manually confirmed the regex captures only the targeted fence and not subsequent content (the indented-fence trap).
- [ ] CI run on the PR.

### DIFF
--- a/claude/.claude/hooks/tests/test_hooks.py
+++ b/claude/.claude/hooks/tests/test_hooks.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import hashlib
 import json
 import os
+import re
 import subprocess
 import time
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #55, opened 2026-04-28
Add require-stow-reminder PreToolUse hook

jcdendrite:
## Summary

- Adds `require-stow-reminder.sh` PreToolUse hook that fires on `gh pr create` / `gh pr edit` in the claude-config repo.
- Detects when a PR adds a new immediate child of `claude/.claude/` (file or directory) by diffing `main...HEAD` against the stow source.
- Requires the PR body, a `--body-file`/`--template` file, or a `--fill`-sourced commit message to mention `install.sh` or `stow` (case-insensitive). Denies otherwise with a remediation message.
- Wired in `settings.json` under PreToolUse → Bash with `if`-dispatch on `Bash(gh pr create *)` and `Bash(gh pr edit *)`, parallel to `deny-private-project-refs.sh`.

## Why

GNU Stow links each top-level child of `claude/.claude/` individually into `~/.claude/`. A brand-new child only materializes after re-running stow — `git pull` alone doesn't create the symlink, and Claude Code reads from `~/.claude/<X>` which stays empty. The `agents/` folder added in #51 hit exactly this failure mode: agents present in the repo, invisible to sessions until someone manually re-ran the install step.

The gate fires at PR-creation time (not commit time) so the reminder lands in the artifact you read at merge time.

## Scope and posture

- Repo-scoped: only fires when `origin.url` contains `claude-config`. Other repos pass through.
- Internal regex filter inside the hook (defense-in-depth against settings.json `if`-pattern drift), parallel to existing gates.
- For `gh pr edit`: only enforces when the edit is changing the body. Title-only / label-only edits pass through, since the create-time gate guarantees the marker landed initially.
- Fail-closed on malformed JSON; fail-open when local `main` ref is missing (rare fresh-clone state). Both documented in the header.
- Known gaps documented inline (matches the `deny-private-project-refs.sh` known-gaps section): shell command-substitution inside `--body`/`--title`, body-file pseudo-paths.

## Test plan

- [x] 17 new cases under `TestRequireStowReminder` covering allow paths (no new top-level, marker in body, marker in body-file, marker in `--fill` commit, label-only edit), deny paths (new dir/file without marker, body-file/edit-body without marker, `--fill` without marker), and edge cases (unrelated repo, missing `main`, malformed JSON).
- [x] Full hook suite green: 218/218.
- [x] Self-test against this PR: this PR adds files inside the existing `hooks/` and `hooks/tests/` top-level entries (no NEW top-level child), so the gate would not fire on it — verifying the false-positive direction.

### DIFF
--- a/claude/.claude/hooks/require-stow-reminder.sh
+++ b/claude/.claude/hooks/require-stow-reminder.sh
@@ -1,0 +1,197 @@
+#!/bin/bash
+# Gate: when a PR being opened/edited against claude-config introduces a
+# new top-level entry under `claude/.claude/`, require the PR body (or a
+# referenced body-source file, or a referenced commit message via
+# `--fill`) to mention `install.sh` or `stow`. Reason: GNU Stow links
+# each child of `claude/.claude/` individually into `~/.claude/`, but a
+# *new* child only appears after re-running stow — `git pull` alone
+# doesn't create the symlink. Without a reminder in the PR body,
+# whoever merges and pulls won't know to re-run install.sh, and the new
+# folder/file silently fails to load (Claude Code reads from
+# ~/.claude/<X>, which is empty until stow links it).
+#
+# NOTE — `if`-dispatch is advisory; the real gate is the internal regex
+# below. settings.json wires two `if` entries (`Bash(gh pr create *)`,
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #57, opened 2026-04-28
README: document require-stow-reminder.sh in the Hooks section

jcdendrite:
## Summary

- Adds a one-bullet entry for `require-stow-reminder.sh` in the README's Hooks section, slotted next to `deny-private-project-refs.sh` since both gate `gh pr create` / `gh pr edit`.
- Closes the documentation gap from #55, which added the hook + tests + settings.json wiring but not the README entry.

## Test plan

- [x] Visual check: bullet density / wording matches the surrounding entries (scope, what it blocks, mechanism, rationale).
- [x] No code change; nothing for the hook test suite to verify.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #59, opened 2026-04-28
require-worktree hook: append cwd-anchor note on chained cd && git

jcdendrite:
## Summary

- Refines the deny message in `require-worktree-for-git-writes.sh` so that when the offending command shows a `cd ... <chain> ... git ...` pattern, the deny reason appends a self-correction note explaining the hook reads cwd from Claude Code's session-persisted bash state — not from the inline `cd` — and pointing at the fix (anchor cwd via a standalone `cd /path/to/worktree` Bash call before the git op).
- Conditional append: pattern requires `cd <word> <chain-op> ... git`. Plain `git commit` from the main tree gets the original message unchanged. Targeting keeps the common deny short.
- 5 new tests on `TestRequireWorktreeForGitWrites` covering each chain operator (`&&`, `;`, `||`), a plain-git negative case, and a reverse-order (`git ... && cd ...`) negative case. Adds `run_hook_reason()` helper for tests that need to assert on deny message contents.

## Why

This session, I tripped on the same gotcha twice while opening follow-up PRs in this repo: ran `cd /worktree && git add ...` chained into a single Bash call, expected the cd to land me in the worktree, got denied because the hook reads cwd from JSON `tool_input` (the bash tool's *persisted* cwd from prior calls), not from the inline cd in the current command.

Initially I reached for an auto-memory entry to capture the workaround. Audited that against the `ai-instruction-and-memory-files` skill and the answer was clear: structural enforcement at moment-of-need beats always-loaded prose. The hook is the natural place — it's already 100% effective at denying, and the deny reason is the artifact the agent reads when it self-corrects. Adding the explanation there delivers the lesson exactly when it's needed and costs zero session-context budget when irrelevant.

## Test plan

- [x] 5 new tests under `TestRequireWorktreeForGitWrites` covering both append paths and non-append paths.
- [x] Full suite 225/225 green (added 5; was 220 after #55 + #57 landed).
- [x] Manual trace: regex `(^|[[:space:]])cd[[:space:]].*(&&|\|\||;).*git` is ERE-compatible; `\|\|` correctly matches literal `||`. Word-anchored `cd` boundary prevents false positives on substring matches inside paths or other tokens.

### DIFF
--- a/claude/.claude/hooks/require-worktree-for-git-writes.sh
+++ b/claude/.claude/hooks/require-worktree-for-git-writes.sh
@@ -31,6 +31,23 @@ emit_deny() {
     "$reason_json"
 }
 
+# When the command chains `cd ... <op> ... git ...`, the agent likely
+# expected the inline `cd` to land inside a worktree. This hook reads
+# CWD from the JSON tool_input — which reflects Claude Code's bash
+# tool's *persisted* cwd from prior calls, not the cwd produced by the
+# inline `cd` in the current command. The hook fires before the
+# subshell runs, so the inline `cd` hasn't taken effect yet. Returns a
+# self-correction note for the agent when this pattern is detected;
+# empty otherwise. Pattern requires `cd` as a word, then a chain
+# operator, then `git` somewhere later — this targets the precise
+# failure mode (`cd /worktree && git ...`) and avoids false positives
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #60, opened 2026-04-28
require-code-review hook: append marker-chain note on chained marker-write && commit

jcdendrite:
## Summary

- Refines `require-code-review.sh`'s deny message: when the offending command shows a `>` redirect into a path containing `review-markers` chained before `git commit`, appends a self-correction note explaining that PreToolUse hooks evaluate at command-submission time (before the chained subshell runs), so the hook read the marker file from disk before the chained marker-write executed. Fix: submit the marker-seed as its own Bash call first, then run `git commit` separately.
- Mirrors the structural-enforcement pattern shipped for the worktree hook in #59. Same conditional-append philosophy, same helper-function shape, same parallel test structure.
- Restructures the deny path from a raw `echo` with single-quoted JSON to a `REASON` variable + `jq -Rs` encoding + `printf`. Necessary for interpolation; incidentally hardens against any future deny content with `"` or `\`.
- 6 new tests covering each chain operator (`&&`, `;`, `||`), a plain-commit negative, a literal-substring-in-message false-positive guard, and a realistic canonical-marker-recipe positive.

## Why

This session, I audited the existing `feedback_code_review_marker_chaining.md` auto-memory entry against the `ai-instruction-and-memory-files` skill. Same conclusion as the cwd-anchor case (PR #59): if a hook structurally enforces a rule, prose memory restating it is pure load — the deny message is the right place to teach the lesson, delivered exactly when the failure fires.

`/plan-review` flagged three changes against the initial plan, all addressed:
1. Note text now covers both stale-marker and missing-marker cases (`"Whether the marker was missing entirely or held a prior review's hash, the chained marker-write hadn't executed yet."`).
2. Added a 6th test using a near-canonical marker recipe shape — the realistic chain an agent following SKILL.md would produce, not just synthetic forms.
3. Multi-line-command limitation documented in the helper's header comment.

## Follow-up

After merge, deletion of `feedback_code_review_marker_chaining.md` and removal of its index line in `MEMORY.md` (the auto-memory location, not in this repo). I'll handle that as a session-local cleanup, no PR.

## Test plan

- [x] 6 new tests under `TestRequireCodeReview` covering both append paths and non-append paths (including the canonical-recipe realistic case).
- [x] Full suite 231/231 green (added 6; was 225 after #59).
- [x] Manual trace: regex `>[^&|;]*review-markers[^&|;]*(&&|\|\||;)[^&|;]*git[[:space:]]+commit` is ERE-compatible. False-positive guards verified for all three constructive negative cases.
- [x] Deny output JSON-shape equivalent to pre-change (refactored `echo` → `printf` with `jq -Rs`-encoded reason; round-trip safe).

### DIFF
--- a/claude/.claude/hooks/require-code-review.sh
+++ b/claude/.claude/hooks/require-code-review.sh
@@ -25,6 +25,29 @@
 # - The marker auto-invalidates as soon as the staging area changes, so
 #   re-staging after review correctly forces a re-review.
 
+# When the command shows a redirect (`>`) into a path containing
+# `review-markers` chained before `git commit`, the agent likely
+# chained the marker-seed with their commit in a single Bash call.
+# PreToolUse:Bash hooks evaluate at command-submission time, BEFORE
+# the chained subshell runs, so this hook reads the marker file from
+# disk before the chained marker-write executes — denying a chain
+# that would have worked if its commands ran in sequence outside the
+# hook. Returns a self-correction note for the agent when this pattern
+# is detected; empty otherwise.
+#
+# Pattern requires `>` redirect, then `review-markers` in the path,
+# then a chain operator, then `git commit`. Single-line commands only
+# — multi-line bash with `\` continuations between the redirect and
+# `git commit` won't match (the regex uses `[^&|;]*`, which excludes
+# newlines under default ERE behavior). In practice agents submit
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #61, opened 2026-04-28
plan-review + code-review: design-fitness gate, default-to-judgment, prior-round context

jcdendrite:
## Summary

Three connected changes targeting review-process drift surfaced by real-world use:

- **Design-fitness gate** (Step 2 in plan-review, Step 1 in code-review). Before any gap-finding, answer "is the design / implementation appropriately sized for the ticket scope?" The skill stops on over-elaboration and surfaces the simpler design as the primary review output. Lists five concrete markers of over-elaboration (dormant Phase-N logic, defensive paths against unobserved attacks, duplicate layers, granularity exceeding consumer needs, captured outputs with no reader). Distinguishes implementation choice (reviewer-tunable) from feature scope (ticket-fixed).

- **Default-to-judgment** in Reviewer roles (plan-review) and Ripple-effect triage (code-review). The dispatcher is the principal-engineer-generalist; specialist subagents have narrower lane depth, not broader context. Spawn only when an escalation criterion applies: below-staff-level domain depth, high-stakes domain boundary, holistic-reasoning overload, convergence-as-design-tell signal, or explicit user request. The always-spawn carve-out for `ciso-reviewer` on auth/secrets/etc. is preserved as a high-stakes-domain case. Spawn per question, not per file-path domain.

- **Prior-round context** mandated on re-spawn. Reviewers without prior context re-discover; that's wasted spawn. The skill now requires the dispatcher to pass prior findings + applied fixes when re-invoking on a re-review round.

- **Reconciliation section** added to both skills. When reviewer findings concentrate on a single surface, the dispatcher pauses and asks two questions, not one: "design-wrong-shape (replace and re-review) or prompt-overlap artifact (apply once, skip duplicates)?" Refines the convergence-as-design-tell signal without picking an arbitrary number threshold.

- B6 (Proportionality) in plan-review updated to defer whole-design proportionality to the gate; checklist focuses on local proportionality only.

## Why

Review rounds were producing more elaboration rather than less: each round's reviewers found gaps in the prior round's defensive infrastructure, the dispatcher patched them mechanically, and the design accumulated complexity until the wrong abstraction was visibly wrong. Three layered fixes target the actual root causes:
- The skills had no design-fitness gate before gap-finding, so over-elaborate designs accumulated patches indefinitely.
- The skills defaulted to spawning the full specialist panel on cross-boundary changes, regardless of whether the dispatcher's own judgment would have sufficed.
- Reviewer subagents had no memory across rounds, re-discovering concerns each round.

The over-elaboration markers in the gate are real patterns from recent debugging, not theoretical risks. The default-to-judgment change ships with explicit escalation criteria rather than a new principal-engineer-generalist persona — the main agent already inhabits that role; what was missing was clear escalation criteria.

## Test plan

- [ ] Read both skill bodies in full to check internal consistency (B6 reference to gate, Reconciliation reference from Reviewer roles, etc.)
- [ ] Apply the design-fitness gate to a sample plan that's over-elaborated; confirm the skill instructs the reviewer to stop and surface the simpler design rather than proceed to gap-finding.
- [ ] Apply the default-to-judgment change to a single-domain change; confirm the dispatcher forms its own review without spawning specialists, and that the output describes which boundaries were considered.
- [ ] Apply the prior-round context change on a re-review; confirm prior findings are passed to specialists.
- [ ] Confirm the always-spawn carve-out for `ciso-reviewer` still fires on auth/secrets changes despite the default-to-judgment reframe.

### DIFF
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -270,25 +270,27 @@ S6. **Secret lifecycle** — Does the plan describe provisioning, storage, rotat
 
 ## Reviewer roles
 
-You are the principal-engineer-generalist running this skill. Plan
-review is the design-stage gate — specialist subagents pressure-test
-the design before code is written. Implementation-stage concerns can
-still be caught during code-review, so self-judge on borderline calls
-without committing to specialist depth.
+You are the principal-engineer-generalist running this skill. The
+specialist subagents below have narrower lane depth, not broader
+context — you have the conversation, the plan in full, prior rounds
+of review, and the user's calibration cues; they have one lane each.
 
 **Default to your own review across all detected domains using the
-checklists.** Spawn a specialist only when one of these applies (same
-escalation logic as code-review's Ripple-effect triage — repeated
-because skills load on demand, not because the rule differs):
+checklists.** Spawn a specialist subagent only when at least one
+escalation criterion applies:
 
-- **Below-staff-level depth** in a specialized kernel (cryptography,
-  query-planner internals, fine-grained accessibility, kernel-level
-  concurrency, etc. — not "general domain knowledge").
-- **High-stakes domain boundary**: RLS, auth, billing, payment, data
-  migration, privileged operations.
-- **Holistic-reasoning overload** across multiple domains.
-- **Convergence-as-design-tell** from a prior round (see Reconciliation).
-- **Explicit user request**.
+- **Below-staff-level depth in a specific domain.** "General backend
+  knowledge" doesn't qualify; specialized kernels do — query-planner
+  internals, cryptography primitives, fine-grained accessibility,
+  kernel-level concurrency, etc.
+- **High-stakes domain boundary.** RLS, auth, billing, payment, data
+  migration, privileged operations. A specialist's second pair of eyes
+  is worth the spawn even at staff generalist depth.
+- **Holistic-reasoning overload.** A multi-domain change where you
+  genuinely can't hold all the lanes in working memory simultaneously.
+- **Convergence-as-design-tell signal** from a prior round (see
+  Reconciliation below).
+- **Explicit user request** for specialist review.
 
 Always spawn `ciso-reviewer` when the plan touches authentication or
 authorization, secrets, tokens, data exposure, logging of sensitive
@@ -322,9 +324,6 @@ project-specific reviewer roles and focus areas, but must not remove or
 narrow the `ciso-reviewer` trigger conditions.
 
 ## Reconciliation
-
-*Same logic as code-review's Reconciliation — repeated because skills
-load on demand.*
 
 After spawned reviewers return findings, pause if findings concentrate
 on a single surface — the same feature, implementation detail, or
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -270,27 +270,25 @@ S6. **Secret lifecycle** — Does the plan describe provisioning, storage, rotat
 
 ## Reviewer roles
 
-You are the principal-engineer-generalist running this skill. The
-specialist subagents below have narrower lane depth, not broader
-context — you have the conversation, the plan in full, prior rounds
-of review, and the user's calibration cues; they have one lane each.
+You are the principal-engineer-generalist running this skill. Plan
+review is the design-stage gate — specialist subagents pressure-test
+the design before code is written. Implementation-stage concerns can
+still be caught during code-review, so self-judge on borderline calls
+without committing to specialist depth.
 
 **Default to your own review across all detected domains using the
-checklists.** Spawn a specialist subagent only when at least one
-escalation criterion applies:
+checklists.** Spawn a specialist only when one of these applies (same
+escalation logic as code-review's Ripple-effect triage — repeated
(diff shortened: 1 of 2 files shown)


claude/.claude/skills/plan-review/SKILL.md
```diff
@@ -244,12 +270,43 @@ S6. **Secret lifecycle** — Does the plan describe provisioning, storage, rotat
...
+  internals, cryptography primitives, fine-grained accessibility,
+  kernel-level concurrency, etc.
+- **High-stakes domain boundary.** RLS, auth, billing, payment, data
+  migration, privileged operations. A specialist's second pair of eyes
+  is worth the spawn even at staff generalist depth.
+- **Holistic-reasoning overload.** A multi-domain change where you
+  genuinely can't hold all the lanes in working memory simultaneously.
+- **Convergence-as-design-tell signal** from a prior round (see
+  Reconciliation below).
+- **Explicit user request** for specialist review.
```
jcdendrite: This is duplicated with code-review skill. Is that intentional? Should there be refinements to make this more specific to each skill? It can't be the case that plan review and code review require the exact same guidance, right?

jcdendrite: **[Claude Code]** Good catch — duplication is significant (~75-80% verbatim across the two sections). Walked through whether differentiation is possible per criterion:

| Criterion | Genuinely different per skill? |
|---|---|
| Below-staff-level depth | No — same specialized kernels (cryptography, query-planner, accessibility) regardless of plan-vs-diff |
| High-stakes domain boundary | No — RLS/auth/billing/migration are equally sensitive at plan time and code time |
| Holistic-reasoning overload | No — multi-domain is multi-domain |
| Convergence-as-design-tell | No — same Reconciliation logic |
| Explicit user request | No |

The criteria themselves genuinely are the same logic. Same generalist running the skill, same specialist agents available, same "do I have enough depth" question.

**Where differentiation IS possible: the framing of stakes.**

- Plan-review = design-stage gate. Mistakes are cheap (revisit during code-review). Self-judging on borderline calls is OK.
- Code-review = last gate before changes ship. Specialist miss = production regression. Be more conservative than at plan-review.

That distinction belongs in the opening paragraph of each section, not in the criteria text.

**Two real inconsistencies the duplication review surfaced:**

1. **Plan-review has a forced-spawn carve-out for `ciso-reviewer` (auth/secrets/tokens/etc.) and `staff-product-engineer` (user-facing behavior) — code-review doesn't.** That's an oversight, not a designed difference. If anything, code-review should be *stricter* on the forced-spawn (last gate), not looser.
2. **Reconciliation sections are also near-verbatim** — same two-readings logic, same recommended response. Same conclusion: identical logic, can't be meaningfully differentiated.

**Why duplication can't be fully eliminated:**

Per `ai-instruction-and-memory-files`'s "duplicate when load paths differ" guidance: skills load on demand, one at a time. When plan-review runs, code-review's body isn't in context, and vice versa. So the rule has to fire from each skill's own body. Unlike `@AGENTS.md` for CLAUDE.md, skills don't have a documented import mechanism (I'd want to verify this before relying on it).

**Proposal — apply now in this PR:**

1. **Differentiate the opening framing.** Plan-review opens with "design-stage gate, errors recoverable in code-review"; code-review opens with "last gate before ship, specialist miss = regression."
2. **Add the forced-spawn carve-out to code-review** (ciso-reviewer + staff-product-engineer), matching plan-review's policy. Probably stricter at code-review (last gate).
3. **Compress the criteria text in both** — same content, fewer words. The verbose explanations duplicated across both files was the most painful part of the duplication.
4. **One-liner in each section** acknowledging intentional duplication: "Same escalation logic as plan-review's Reviewer roles / code-review's Ripple-effect triage — repeated because skills load on demand." That makes the duplication legible, so a future reader understands it's deliberate.
5. **Reconciliation: leave duplicated** with the same one-liner. Genuinely shared logic.

Net change: criteria sections in both files get shorter; framing gets differentiated; the previously-missing carve-out in code-review is added. Total churn moderate (~30 lines edited, mostly compression and reframing).

Want me to apply (1)-(5) on this PR, or defer to the broader length audit follow-up? My read: (2) is a real correctness fix and should land here; (1), (3), (4), (5) are about reducing the visible duplication and could go either way.

jcdendrite: **[Claude Code]** Resolution: of the five proposed items, only **#2 landed** in this PR — the missing forced-spawn carve-out for `ciso-reviewer` (auth/secrets/etc.) and `staff-product-engineer` (user-facing behavior) on code-review's Ripple-effect triage. That was the real correctness fix, not a deduplication concern.

The other four (framing differentiation, criteria compression, cross-reference notes in both Reviewer-roles/Ripple-triage and Reconciliation sections) are deferred to the broader length audit follow-up, with concrete instructions in `~/.claude/plans/handoff-claude-config-skills-length-audit.md` under "Deferred items from PR #61's review."

Final state of this PR: commits `388c9f9` (additions) → `23f1b39` (trim) → `98171ef` (temp dedup/framing/carve-out) → `411b1bb` (revert dedup/framing, keep only the carve-out). Plan-review back to 411 lines (post-trim baseline); code-review at 401 lines (post-trim + 6 lines for the kept carve-out).

jcdendrite (commented):
lgtm

PR #68, opened 2026-04-29
ready-for-review: add push gate (marker + hook)

jcdendrite:
## Summary

Mechanically enforces that `/ready-for-review` has run on the artifact about to be published. Closes the failure mode where Claude pushes wrap-up commits to a PR branch (or runs `gh pr ready`) without going through the gate.

Layered on top of #65, which fixed the trigger conditions; this PR adds the marker + hook so the gate is enforced at the tool boundary, not just by skill prose.

## What it does

**Two-marker pattern** (mirrors `require-respond-pr.sh`'s active-bypass shape and `require-code-review.sh`'s content-hash shape):

- **Active marker** `~/.claude/.ready-for-review-active.d/<session>` — presence-only, fresh <60min, mtime refreshed on bypass. Written at step 0, removed at step 7. Bypasses the gate during the skill's own iteration (step 3 fix → push → loop back to step 2) so the skill can't self-deny.
- **Completion marker** `~/.claude/ready-for-review-markers/<repo-hash>.<session>` — contents are local HEAD SHA at gate-completion time. Written at step 7 only when every halt-on-fail step passed. Pushes against the recorded HEAD are allowed; new commits invalidate it (HEAD moves) and force a re-run.

**Gated commands:** `git push` to a branch with an open PR, and `gh pr ready` (draft → ready transition — explicit handoff signal).

**Bypass shapes (no marker required):**
- `--dry-run`
- `--delete` / `-d` / `origin :branch` (deletion forms)
- `--tags`-only (no branch refspec)
- Default-branch pushes
- Branches without an open PR
- `gh pr view` failure (network glitch / gh not configured) — fails open to keep the user unblocked; the skill's prose triggers still cover the case

**Settings registration** uses scoped `if` clauses (`Bash(git push *)` and `Bash(gh pr ready *)`) plus internal command filtering inside the hook — defense-in-depth per `feedback_hook_defense_in_depth`.

## Bug caught during testing

`git rev-parse --abbrev-ref origin/HEAD` returns the literal string `"origin/HEAD"` (not empty) when origin/HEAD isn't a symbolic ref, which silently defeated the default-branch fallback path. Switched to `git symbolic-ref --quiet refs/remotes/origin/HEAD`. This was the only finding from running the test suite.

## Test plan

- [x] 30 new tests in `TestRequireReadyForReview`: bypass shapes (non-push, --dry-run, --delete, :branch, --tags-only, default-branch, no-PR, outside-repo); active-marker (fresh / stale / cross-session leak / mtime refresh); completion-marker (matching / stale-HEAD / cross-session leak / no-session-id); gh-failure fail-open; gh pr ready gating; skill-hook alignment via HOOK_TEST_FIXTURE recipes.
- [x] All 261 tests pass (231 pre-existing + 30 new).
- [x] Skill-on-its-own-diff: no contradictions with the existing intro (`Run steps in order. Halt on failures unless...`) or step semantics. Step 0 (setup) and step 7 (teardown) bracket the existing 1–6 flow without altering them.

## Notes

- The hook is only live for users who pull this PR (after merge); pushing this branch doesn't activate the gate against itself.
- This makes #65's trigger fix mechanically enforceable. Per `feedback_dont_auto_merge_user_prs`: not merging — that's your call.

### DIFF
--- a/claude/.claude/hooks/require-ready-for-review.sh
+++ b/claude/.claude/hooks/require-ready-for-review.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+# Gate: require /ready-for-review to have run before pushing to a branch
+# with an open PR (or marking a draft PR ready). Verified via marker file.
+#
+# Why: pre-handoff gate (/ready-for-review) verifies tests/lint/typecheck,
+# runs cumulative /code-review, syncs PR body, and checks CI before the
+# human reviews. Without enforcement, Claude habitually pushes wrap-up
+# commits straight to a PR without running the gate. Mirrors the
+# require-code-review.sh pattern: marker keyed by session, content matches
+# the artifact about to be published (HEAD SHA, not staged-diff hash —
+# the analogue is the squash-merge artifact reviewers see, not the
+# upcoming commit).
+#
+# Two-marker pattern:
+# - Active marker (~/.claude/.ready-for-review-active.d/<session_id>):
+#   presence-only, fresh mtime <60 min. Written by /ready-for-review at
+#   step 0; removed at step 7 (completion). Bypasses the gate so the
+#   skill's own iteration pushes (step 3 fix → push → loop) don't
+#   self-deny. mtime refreshed on each bypass to handle long runs.
+# - Completion marker (~/.claude/ready-for-review-markers/<repo-hash>.<session_id>):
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

PR #69, opened 2026-04-29
plan-review: add user-surface declaration + threat-model bounding to Step 2

jcdendrite:
## Summary

- Reframes the design-fitness gate (Step 2) around two questions: which user surface the change serves (production / internal-only / dev-only / mixed), and whether the design is sized for the user pain on that surface. Threat model follows from the surface — persona reviewers must scope findings to the declared surface, not default to a worst-case external attacker.
- Adds a surface-bounded skip clause to the always-spawn `ciso-reviewer` rule: a dev-only flow with no production reachability, or an internal path where engineers are the only callers and no privilege boundary is crossed, can skip the CISO spawn — but the surface must be named in the review output, never silently skipped.
- Refines the over-elaboration markers list — replaces "scenarios that haven't been observed" with "defensive layers beyond what the declared user surface and threat model justify"; adds "could be done in N lines" as a valid challenge even after persona-shaping.

## Why

`/plan-review` already had a design-fitness gate (Step 2). What it lacked was a way for plan authors to declare the user surface up front, so persona reviewers could scope their threat models accordingly. Without that, persona reviewers default to worst-case (external attacker, anonymous reachability) on every plan, inflating defensive layers for changes that don't warrant them — internal-only DX fixes, dev-only smoke-test tooling, IDE-preview-only flows.

The change is +10/-3 lines, augmenting Step 2 in place rather than introducing a new step. Surfaced after a recent change where persona-shaped plan elaboration shipped a heavy popup-with-postMessage flow for what was fundamentally an internal-only DX fix on a code path production users never reach.

## Test plan

- [x] `/code-review` run on the diff (no findings; recorded marker)
- [x] Self-check against the skill's own design-fitness gate — not over-elaborated, one marker swap rather than addition, structure preserved
- [ ] Reviewer: confirm the surface-bounded skip clause is sound (named-surface requirement + downstream `code-review` `ciso-reviewer` spawn intact = adequate safety net)
- [ ] Reviewer: confirm Step 2 reads as parallel to `/code-review`'s Step 1 without unnecessary divergence

### DIFF
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -34,15 +34,22 @@ Read the plan and classify which domains it touches:
 
 ## Step 2 — Design-fitness gate
 
-Before evaluating gaps, answer: **is the design appropriately sized for the ticket scope?** Gap-finding on an over-elaborate design elaborates it further (each finding closes a gap by adding more machinery), and the checklist won't surface "this whole design is the wrong shape."
+Before evaluating gaps, answer two questions in order:
+
+1. **What user surface does this serve?** Production users, internal-only (engineers / admins), dev-only (CI / test harness / IDE preview), or mixed (name the path each surface takes). The threat model follows from the surface — a dev-only smoke-test flow has no external attacker; an internal-only admin tool has no anonymous attacker. Persona reviewers must scope findings to the declared surface, not default to the worst-case external-attacker model.
+
+2. **Is the design appropriately sized for the user pain it solves on that surface?** Gap-finding on an over-elaborate design elaborates it further (each finding closes a gap by adding more machinery), and the checklist won't surface "this whole design is the wrong shape."
+
+If the plan doesn't declare a surface, demand one before proceeding — "an internal-only DX fix" or "production users never reach this code path" is a valid one-line declaration. Without it, persona reviewers default to worst-case threat models and inflate plan size for changes that don't warrant it.
 
 Markers of over-elaboration:
 
+- Defensive layers stacked beyond what the declared user surface and threat model justify.
 - Conditional logic for future phases that may not arrive.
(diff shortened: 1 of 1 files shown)


claude/.claude/skills/plan-review/SKILL.md
```diff
@@ -34,15 +34,22 @@ Read the plan and classify which domains it touches:
 
 ## Step 2 — Design-fitness gate
 
-Before evaluating gaps, answer: **is the design appropriately sized for the ticket scope?** Gap-finding on an over-elaborate design elaborates it further (each finding closes a gap by adding more machinery), and the checklist won't surface "this whole design is the wrong shape."
+Before evaluating gaps, answer two questions in order:
+
+1. **What user surface does this serve?** Production users, internal-only (engineers / admins), dev-only (CI / test harness / IDE preview), or mixed (name the path each surface takes). The threat model follows from the surface — a dev-only smoke-test flow has no external attacker; an internal-only admin tool has no anonymous attacker. Persona reviewers must scope findings to the declared surface, not default to the worst-case external-attacker model.
```
jcdendrite: This feels over-engineered and explained to the one design flaw case that happened for the private project. We're carving out an awful lot of space for the "surface" callout. Contrast that with #2 which is more general and more appropriate.

jcdendrite: **[Claude Code]** Agreed — the surface taxonomy and "demand a declaration" enforcement paragraph were over-engineered for a concept the original handoff said should be "a one- or two-line free-form statement, not a structured form." Compressed Q1 to a single sentence in 9b5f4b5: question + brief example list, no enumerated categories, no enforcement paragraph. Q1 and Q2 now carry the same weight. The spawn-rule skip clause still resolves to "the Step 2 user-surface declaration" — the declaration is lighter, not absent.

jcdendrite (commented):
lgtm

PR #70, opened 2026-04-29
Extract skill review into dedicated skill-review skill

jcdendrite:
## Summary

Items 33 and 34 of `code-review` (skill trigger accuracy, context budget) were too narrow as bullets in `code-review`'s flat numbered list. Skill-file review has real operational depth — frontmatter rules, trigger design, voice, length targets, operational-vs-narrative content test, cross-reference vs duplication tradeoffs — that doesn't fit as line items in a generic review skill.

Same pattern as the recent `lovable-knowledge` extract: pull the domain into its own skill that owns it end-to-end and expand with the operational content that didn't fit.

## What's extracted

New skill `claude/.claude/skills/skill-review/SKILL.md` (186 lines, `user-invocable: false`) covering:

1. What makes a skill load — required and optional frontmatter, harness load model.
2. Trigger design — TRIGGER / DO NOT TRIGGER as parallel lists; specificity priority (file globs > action verbs > context cues); the false-positive vs false-negative cost tradeoff.
3. Voice and structure — imperative second person, declarative numbered items, section-header rules, table/prose/bullet usage.
4. Length and the behavior test — defers to `ai-instruction-and-memory-files` §2 rather than restating.
5. Operational vs narrative content — passes-the-test and fails-the-test categories with a tone-marker table.
6. Cross-references vs duplication — mirrors the framework in `ai-instruction-and-memory-files` §3.
7. Review checklist — 10 concrete items for reviewing `.claude/skills/**/SKILL.md` PRs.

## What stayed in code-review

- Item 35 (permission scope) — already references `review-permissions` skill.
- Item 36 (hook correctness) — out of scope here (separate parallel extraction).
- The "Domain: Claude Code config" section now has a single pointer line at the top: `For .claude/skills/**/SKILL.md review, see the skill-review skill.` (Mirrors the existing Lovable pointer pattern.)
- Item ownership table row `12, 33, 34. Judgment items` → `12. Pre-existing issues in unchanged code`.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/` — 231 passed
- [x] `/code-review` on the staged change — no issues, no specialist spawn warranted (doc-only, no auth / schema / migration / secrets / runtime-config crossings)
- [x] `skill-review` applied to its own diff — frontmatter, voice, length (under target), behavior test per line, redaction (PROJ- / TICKET- placeholders), cross-references all check out
- [ ] Reviewer confirms the new skill reads coherently against the behavior test
- [ ] Reviewer confirms `code-review`'s "Domain: Claude Code config" section reads coherently after the trim

### DIFF
--- a/claude/.claude/skills/skill-review/SKILL.md
+++ b/claude/.claude/skills/skill-review/SKILL.md
@@ -2,10 +2,9 @@
 name: skill-review
 description: >
   How to review and audit Claude Code skill files
-  (`.claude/skills/**/SKILL.md`): frontmatter conventions, TRIGGER /
-  DO NOT TRIGGER design, voice, length targets, the
-  operational-vs-narrative content test, and the cross-reference vs
-  duplication framework.
+  (`.claude/skills/**/SKILL.md`): frontmatter conventions, trigger
+  design, voice, length targets, the operational-vs-narrative content
+  test, and the cross-reference vs duplication framework.
   TRIGGER when: editing or reviewing `.claude/skills/**/SKILL.md`;
   auditing trigger accuracy across a skill set; restructuring or
   pruning skills.
--- a/claude/.claude/skills/skill-review/SKILL.md
+++ b/claude/.claude/skills/skill-review/SKILL.md
@@ -5,11 +5,12 @@ description: >
   (`.claude/skills/**/SKILL.md`): frontmatter conventions, trigger
   design, voice, length targets, the operational-vs-narrative content
   test, and the cross-reference vs duplication framework.
-  TRIGGER when: editing or reviewing `.claude/skills/**/SKILL.md`;
-  auditing trigger accuracy across a skill set; restructuring or
-  pruning skills.
-  DO NOT TRIGGER when: creating a brand-new skill from scratch (use
-  `skill-creator`); editing CLAUDE.md or AGENTS.md (use
+  TRIGGER when: reviewing a `.claude/skills/**/SKILL.md` change
+  (PR / diff / pre-commit audit); auditing trigger accuracy across a
+  skill set; restructuring or pruning skills.
+  DO NOT TRIGGER when: authoring, iterating on, or optimizing a skill
+  (use `skill-creator` for scaffolding, eval / benchmark loops, and
+  description tuning); editing CLAUDE.md or AGENTS.md (use
   `ai-instruction-and-memory-files`); editing `.lovable/*.md` (use
   `lovable-knowledge`); reviewing code that isn't a skill file.
 user-invocable: false
@@ -17,9 +18,12 @@ user-invocable: false
 
 # Skill Review — Architecture & Checklist
 
-Complements `skill-creator` (scaffolds new skills, runs evals, tunes
-description text). This skill applies a review checklist to skill-file
-diffs.
+Fires in **review and audit** contexts (applying a checklist to a
+skill-file change). `skill-creator` covers the authoring loop —
+scaffolding, evals, benchmarking, description tuning. The lane split
+is by context, not by output type: skill-review stays out of authoring
+turns to avoid dual-firing with `skill-creator`'s broad "edit existing
+skill" claim.
 
 ## 1. What makes a skill load
 
@@ -64,13 +68,13 @@ DO NOT TRIGGER when: <obvious misfire>; <adjacent skill's surface>; <out of scop
 **Tradeoff.** Over-broad triggers load the skill's body into
 unrelated turns and steal context from the active task. Over-narrow
 triggers leave the skill dormant when it should fire and the user
-ends up restating what the skill already knows. Both fail; tune by
-sampling real recent transcripts where the skill should have fired.
+ends up restating what the skill already knows.
 
 **DO NOT TRIGGER carries equal weight.** Name the adjacent skills
-whose surfaces overlap. A `skill-review` description that says only
-"editing skill files" without naming `skill-creator` will fire on
-brand-new-skill scaffolding turns where `skill-creator` should own.
+whose surfaces overlap. A `skill-review` description that doesn't
+list `skill-creator` under DO NOT TRIGGER will dual-fire on every
+authoring turn `skill-creator` claims to own — costing context
+budget without adding marginal value.
 
 ## 3. Voice and structure
 
--- a/claude/.claude/skills/skill-review/SKILL.md
+++ b/claude/.claude/skills/skill-review/SKILL.md
@@ -18,10 +18,9 @@ user-invocable: false
 
 # Skill Review — Architecture & Checklist
 
-Complements `skill-creator` (scaffolds new skills). Reuses framing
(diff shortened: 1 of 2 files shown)


claude/.claude/skills/skill-review/SKILL.md
```diff
@@ -0,0 +1,186 @@
+---
+name: skill-review
+description: >
+  How to review and audit Claude Code skill files
+  (`.claude/skills/**/SKILL.md`): frontmatter conventions, TRIGGER /
```
jcdendrite: will the caps TRIGGER and DO NOT TRIGGER in the description be misinterpreted as a false trigger or do-not-trigger

claude/.claude/skills/skill-review/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+overspend hurts more than body overspend.
+
+## 2. Trigger design
+
+Format the description with two parallel lists:
+
+```
+TRIGGER when: <specific situation>; <another>; <another>.
+DO NOT TRIGGER when: <obvious misfire>; <adjacent skill's surface>; <out of scope>.
+```
```
jcdendrite, line 56: Doesn't this part overlap with the skill creator skill? If I'm not mistaken, the skill creator skill triggers when a skill is created for the first time, and this skill-review skill would be triggered when a skill is modified. Is my understanding correct?

claude/.claude/skills/skill-review/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+  "Review checklist") — never "Introduction," "Background,"
+  "Conclusion." The headers are skim navigation; padding headers add
+  load without a routing payoff.
+- **Tables for matrices, prose for narratives, bullets for parallel
+  options.** A four-column comparison wants a table; a two-step
+  decision wants prose.
+
+## 4. Length and the behavior test
+
+`ai-instruction-and-memory-files` §2 sets the targets — 200 lines per
```
jcdendrite: Why would we redirect to another skill here - isn't that going to cause decay? What's the harm in restating the targets here for skill length?

claude/.claude/skills/skill-review/SKILL.md
```diff
@@ -0,0 +1,186 @@
...
+| Tone marker | Operational | Narrative (cut) |
+|---|---|---|
+| Subject | "Check whether..." | "We've found that..." |
+| Tense | imperative present | past, retrospective |
+| Audience | Claude, mid-task | a human reading cold |
+| Specificity | concrete pattern | generalized lesson |
+
+## 6. Cross-references vs duplication
+
+`ai-instruction-and-memory-files` §3 sets the duplication framework
```
jcdendrite: Is this level of indirection going to be helpful to Claude? Duplication is not a concept that exists solely with instruction and memory files. It might be more helpful to Claude sessions to state this here.

jcdendrite (commented):
high-level prelim review

claude/.claude/skills/skill-review/SKILL.md
jcdendrite: **[Claude Code]** You're right — I misread your comment. You meant the bare uppercase `TRIGGER /` and `DO NOT TRIGGER` tokens in the **overview list** (lines 5–6, "frontmatter conventions, TRIGGER / DO NOT TRIGGER design, voice..."), not the structured `TRIGGER when:` / `DO NOT TRIGGER when:` labels later in the description.

That's a real concern: those bare tokens sit on the always-loaded matching surface as descriptive content, and a dispatcher matching for trigger conditions could plausibly be confused by them.

Fixed in commit aa8d7e9: the overview now reads "frontmatter conventions, trigger design, voice, ..." (lowercase, matching the §2 section title in the body). The structured `TRIGGER when:` / `DO NOT TRIGGER when:` labels are preserved verbatim — those are the format-recognized parse anchors used by every other skill in this repo.

Bare uppercase tokens still appear in the body (§2 fenced format example, the §2 sub-heading), but the body is on-demand-loaded only when the skill fires, so they're off the always-loaded matching surface and don't pose the same risk.

claude/.claude/skills/skill-review/SKILL.md
jcdendrite: **[Claude Code]** Update — verified by reading skill-creator's actual SKILL.md (`~/.claude/plugins/marketplaces/claude-plugins-official/plugins/skill-creator/skills/skill-creator/SKILL.md` line 3):

> "Create new skills, modify and improve existing skills, and measure skill performance. Use when users want to create a skill from scratch, **edit, or optimize an existing skill**, run evals to test a skill, benchmark skill performance with variance analysis, or optimize a skill's description for better triggering accuracy."

So skill-creator does claim "edit existing skill" — your dual-fire concern was justified, and my earlier "lane split by output, accept dual-fire" framing was the wrong call.

Fixed in commit 618c9ad:

- TRIGGER narrowed to "reviewing a `.claude/skills/**/SKILL.md` change (PR / diff / pre-commit audit); auditing trigger accuracy across a skill set; restructuring or pruning skills." The bare "editing" verb is gone.
- DO NOT TRIGGER expanded to "authoring, iterating on, or optimizing a skill (use `skill-creator` for scaffolding, eval / benchmark loops, and description tuning)" — matches the scope skill-creator's own description claims.
- Intro paragraph rewritten to make the lane split **by context** (not by output) explicit: skill-review fires in review/audit contexts only.

The lane split has to be defended from `skill-review`'s side since `skill-creator:skill-creator` is a plugin-namespaced built-in we can't edit. The narrowed TRIGGER + expanded DO NOT TRIGGER is the targeted fix.

claude/.claude/skills/skill-review/SKILL.md
jcdendrite: **[Claude Code]** Good catch — you're right. `skill-review` and `ai-instruction-and-memory-files` fire on different triggers (skill files vs CLAUDE.md / AGENTS.md / memory), so when this skill loads, the §2 content isn't guaranteed to be in context. The pointer was a dead reference at the moment of need.

Fixed in the latest push: §4 now states the targets inline (200-line target, 300 diminishing-returns ceiling, lost-in-the-middle decay, prose ~70% vs structural 100% compliance). The pointer is gone. This passes the duplication test in §6 — different load surfaces, content critical to behavior, the cross-reference path could silently fail.

claude/.claude/skills/skill-review/SKILL.md
jcdendrite: **[Claude Code]** Agreed — same failure mode as your previous comment. The duplication framework is general (applies across skills, instruction files, agent docs), and `skill-review` fires on a different trigger than `ai-instruction-and-memory-files`, so the §3 pointer was indirection that may not resolve at the moment of need.

Fixed in the latest push: dropped the §3 pointer line. The framework is already restated inline in §6 just below it (cross-reference default, three-condition duplication test, drift warning). No content was lost — just the redundant pointer.

jcdendrite (commented):
lgtm

PR #72, opened 2026-04-29
deny-private-project-refs: scan commit-message files and gh api calls

jcdendrite:
## Summary

Two paths existed where tracker IDs or blocklist terms could bypass the `deny-private-project-refs.sh` redaction gate and reach a commit or PR without being scanned:

- **`git commit --file <path>`** — the hook scanned `-m` message arguments but never read file-sourced commit messages.
- **`gh api` mutating calls** — the hook had no dispatch entry for `gh api` at all; review-thread replies, issue comments, and GraphQL mutations were entirely unscanned.

This PR closes both gaps and adds 96 new tests (231 → 327 passing).

## Gap 1 — `git commit --file <path>`

Added `extract_commit_message_source_paths()` to parse `-F`/`--file` arguments from the git commit command and scan the file contents with the same tracker-ID and private-project-name checks already applied to `-m` arguments.

**Edge cases:**
- Pseudo-paths (`-`, `/dev/stdin`, `/dev/fd/*`, `/proc/*/fd/*`) fail closed — the hook cannot read these safely at hook time, so it denies rather than allowing.
- Unreadable files also fail closed for the same reason.
- OSS allowlist passthrough (`CVE-`, `RFC-`, `PEP-`, etc.) is preserved.

## Gap 2 — `gh api` mutating calls

Added a `Bash(gh api *)` dispatch entry to `settings.json` (matching the existing `git commit *` and `gh pr create *` pattern) and a new `IS_GH_API` branch in the hook.

**What is now scanned:**

| Form | What's scanned |
|------|---------------|
| `gh api … --input <path>` | request-body file at `<path>` |
| `gh api … -f key=@<path>` | field-from-file at `<path>` (all four field flag variants) |
| `gh api … -f body="TICKET-123"` | inline field value via existing inline-args scan |

**Method dispatch — what counts as a mutating call:**
- Explicit `-X POST`, `--method POST`, `--method=POST`, `-XPOST` (concatenated), `-X=POST` — and the same for PATCH, PUT, DELETE.
- **Implicit POST auto-promote**: `gh` promotes to POST automatically when any field flag (`-f`/`-F`/`--field`/`--raw-field`) is present, even with no explicit method flag. The dispatch detects this.
- Default GET with no body flags is explicitly allowed (no false positives on read-only queries).

**Pseudo-path / unreadable handling** mirrors Gap 1 — fail closed on both `--input` and `key=@<path>` forms.

## Bypass fixes surfaced in review

Four bypass vectors were identified and closed during code review:

1. **Implicit-POST auto-promote** — the initial dispatch only matched explicit `-X/--method`; `gh api -f body="TICKET-123"` (no explicit method) was not flagged. Fixed: `IS_GH_API` now triggers when any field flag is present regardless of method.
2. **`key=@<path>` field-from-file** — `gh api -f key=@secret-file.txt` reads the file at gh invocation time but the hook only scanned the command string. Fixed: `extract_gh_api_field_at_paths()` extracts the `@<path>` target and scans it.
3. **`-XPOST` concatenated form** — `grep -E '\-X POST'` missed `-XPOST`. Fixed: dispatch regex uses `(-X|--method)(=|[[:space:]]+)?` pattern consistent with existing style.
4. **`\b` portability** — `\b` word boundaries behave differently between BSD and GNU grep. Fixed: replaced with `[^A-Za-z0-9_]` / `([^A-Za-z]|$)` anchors.

## Follow-up commit — quote-unaware false-positive fix

While opening this PR, the hook itself blocked the `gh pr create` command because the PR title contained the literal text `-F`. Root cause: all four `extract_*_paths()` functions grepped the raw command string with no shell-quote awareness, so a flag-shaped substring inside a quoted argument value was matched as a real flag.

A first attempt stripped entire quoted regions before extraction (awk state machine), but that also removed quoted file-path arguments — a false-allow regression where `--body-file "/path/body.md"` left the file unscanned.

Fixed in the follow-up commit: replaced the `grep | sed | sed` pipeline in each of the four extractors with `xargs -n1 | awk`. `xargs` tokenizes the command string respecting shell quoting (emitting one token per line); the awk does exact-match flag detection against those tokens. A multi-word token like `"close git commit -F and ..."` (from a quoted `--title` value) is never matched by `/^-F$/`. Quoted file-path arguments (`--body-file "/path"`) are extracted correctly because `xargs` strips only the surrounding quotes. Tracker-ID and blocklist scans continue to run on the original unstripped command.

12 additional tests lock in false-positive prevention, quoted-path extraction, and the `=`-concatenated flag form for all four extractors.

## Known narrow gaps (documented in hook header)

- **GraphQL `-F query=@file` secondary injection**: a GraphQL query file could contain a tracker ID in a comment or string value. The hook scans the file for raw tracker-ID patterns, but an attacker who controls the file content could embed IDs in query variables. Out of scope for this PR; noted in header.
- **Editor-flow commit messages (`git commit` with neither `-m` nor `--file`)**: the hook has never scanned messages written interactively via an editor; this gap predates and is separate from the `--file` fix.

## Test coverage

New `TestDenyPrivateProjectRefs` class in `test_hooks.py` adds 96 cases (231 → 327 passing):

- Gap 1: deny/allow on `--file`, pseudo-path rejection, unreadable fail-closed, OSS allowlist passthrough, blocklist generic-message invariant
- Gap 2: deny/allow on `--input`, `key=@<path>` (all field flag forms), inline field values; explicit method forms (space, `=`, concatenated); implicit-POST detection; default GET allow; pseudo-path and unreadable fail-closed; blocklist generic-message
- Bypass regression locks: chained-prefix gh api, tracker-vs-blocklist priority, implicit-POST parametrized over all four field flag forms
- Quote-aware extraction: false-positive locks (one per extractor), quoted-path deny+allow controls for all four extractors, `=`-form awk branch coverage

## Files changed

- `claude/.claude/hooks/deny-private-project-refs.sh` — gap implementations + bypass fixes + quote-aware extraction fix + header docs
- `claude/.claude/hooks/tests/test_hooks.py` — 96 new test cases
- `claude/.claude/settings.json` — new `Bash(gh api *)` dispatch entry

### DIFF
--- a/claude/.claude/hooks/deny-private-project-refs.sh
+++ b/claude/.claude/hooks/deny-private-project-refs.sh
@@ -1,17 +1,19 @@
 #!/bin/bash
-# Gate: reject `git commit`, `gh pr create`, and `gh pr edit` if their
-# content (staged diff, commit message, PR title/body, or body-source
-# file contents) contains tracker-ID tokens that aren't on the open-
-# source allowlist. Enforces the tracker-ID piece of the repo-root
-# CLAUDE.md redaction rule ("Redact private-project-identifying
-# content").
+# Gate: reject `git commit`, `gh pr create`, `gh pr edit`, and mutating
+# `gh api` calls if their content (staged diff, commit message, PR
+# title/body, body-source file contents, gh-api JSON body, or
+# referenced --input file) contains tracker-ID tokens that aren't on
+# the open-source allowlist. Enforces the tracker-ID piece of the
+# repo-root CLAUDE.md redaction rule ("Redact private-project-
+# identifying content").
 #
 # NOTE — `if`-dispatch is advisory; the real gate is the internal regex
-# at the top of this script. settings.json wires three `if` entries
-# (`Bash(git commit *)`, `Bash(gh pr create *)`, `Bash(gh pr edit *)`)
-# for zero-cost early dispatch, but any drift between those patterns
-# and the IS_GIT_COMMIT / IS_GH_PR regexes here creates silent coverage
-# gaps. Update both surfaces when extending coverage.
+# at the top of this script. settings.json wires `if` entries
+# (`Bash(git commit *)`, `Bash(gh pr create *)`, `Bash(gh pr edit *)`,
+# `Bash(gh api *)`) for zero-cost early dispatch, but any drift between
+# those patterns and the IS_GIT_COMMIT / IS_GH_PR / IS_GH_API regexes
+# here creates silent coverage gaps. Update both surfaces when extending
+# coverage.
 #
 # Scope and limits:
 # - Catches the mechanical category (tracker IDs shaped like [A-Z]{2,}-\d+).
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #73, opened 2026-04-29
Add Agent Briefing section documenting worktree isolation

jcdendrite:
## Summary

Documents the issue where sub-agents spawned from plan mode with `isolation: "worktree"` should not include an explicit "Working directory" line. When included, the harness reads it as a base path for `git -C` operations, which bypasses worktree isolation and mutates the main working tree directly.

This prevents a failure mode observed with Opus spawning Sonnet for execution after exiting plan mode — the subprocess's git operations would unwittingly operate on the main tree instead of the isolated worktree.

## Why this matters

Worktree isolation is a core safety mechanism to prevent concurrent agents from stepping on each other. The briefing ensures agent authors won't inadvertently break it.

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -31,6 +31,10 @@
 
 - After writing or modifying an implementation plan, run `/plan-review` before presenting the plan to the user. If the review finds issues, address them first, then present the final version.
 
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #74, opened 2026-04-29
Extract hook review into dedicated hook-review skill

jcdendrite:
## Summary

Extracts item 36 (Hook correctness) from \`code-review\` into a new dedicated \`claude-hook-review\` skill, mirroring the \`skill-review\` extraction in #70.

- **New skill** at \`claude/.claude/skills/claude-hook-review/SKILL.md\` (107 lines) — owns the full hook-review surface: event/matcher selection, script skeleton idioms (with \`emit_deny\` quoted directly from the hooks), fail-open vs fail-closed posture, \`if\`-dispatch drift risk, the OSS_ALLOWLIST vs \`~/.claude/private-projects.md\` distinction, performance budget, test patterns, and a review checklist.
- **\`code-review\` updated**: item 36 replaced with a pointer; ownership-table row updated to reference the \`claude-hook-review\` skill while preserving \`staff-platform-engineer\` + \`ciso-reviewer\` as routing targets.

The \`claude-\` prefix was added after PR review — "hook" alone is ambiguous with git hooks, husky, and lefthook; the DO NOT TRIGGER section already called these out.

## What's in the skill body (not in the original item 36 one-liner)

The new skill adds actionable detail that the single item 36 line couldn't carry:

- The \`emit_deny\` helper quoted verbatim so reviewers recognize the shape on sight.
- \`"permissionDecision"\` must be exactly \`"deny"\` (lowercase) — case-sensitive; any variant silently allows.
- \`jq -r '... // empty'\` exits 0 on missing fields (not non-zero) — the \`JQ_EXIT\` guard only catches parse failures; allow-list-style gates need a separate \`[ -z "\$COMMAND" ]\` guard.
- Two distinct extension surfaces: \`OSS_ALLOWLIST\` in the committed script (open-source refs only — adding a private name here publishes it) vs \`~/.claude/private-projects.md\` (private identifiers, user-local, never committed).
- False-positive risk checklist item: matcher + internal filter + repo-url guard should be narrow enough to exclude operations outside the gate's stated purpose.

## What stays in \`code-review\`

Item 35 (Permission scope → \`/review-permissions\`) is unchanged. The ownership-table row for item 36 is retained with a note pointing to the \`claude-hook-review\` skill, so \`staff-platform-engineer\` and \`ciso-reviewer\` remain in the dispatch path for hook changes.

## Test plan

- [x] \`pytest claude/.claude/hooks/tests/test_hooks.py\` — 261 passed
- [x] \`/code-review\` run on staged diff (specialists \`staff-platform-engineer\` + \`ciso-reviewer\` consulted per ripple-triage ownership-table rule; all findings addressed before commit)
- [x] \`code-review\` Domain: Claude Code config section reads coherently: skill-review pointer → item 35 → claude-hook-review pointer
- [x] Ownership table has row for item 36 → \`claude-hook-review\` skill, with correct owners

### DIFF
--- a/claude/.claude/skills/hook-review/SKILL.md
+++ b/claude/.claude/skills/hook-review/SKILL.md
@@ -1,0 +1,107 @@
+---
+name: hook-review
+description: >
+  How to review Claude Code hook scripts (`claude/.claude/hooks/*.sh`)
+  and hook entries in `settings.json` (PreToolUse, PostToolUse,
+  SessionStart): event/matcher selection, dispatch design,
+  fail-open vs fail-closed posture, the `emit_deny` JSON shape,
+  performance budget, and test patterns.
+  TRIGGER when: reviewing a change to a hook script under
+  `.claude/hooks/`; reviewing a change to a hook entry in
+  `settings.json` (`hooks.PreToolUse`, `hooks.PostToolUse`,
+  `hooks.SessionStart`, etc.); designing a new hook.
+  DO NOT TRIGGER when: editing `permissions.allow` rules (use
+  `review-permissions`); editing other `settings.json` fields (env,
+  model, theme — use `update-config`); reviewing non-Claude hook
+  systems (git hooks, husky, lefthook).
+allowed-tools: Read, Grep, Glob, Bash
+user-invocable: false
+---
+
+## 1. Hook events and matchers
+
+**Event type** selects when a hook fires: `PreToolUse` (fires before the tool call; can deny), `PostToolUse` (fires after; observation only), `SessionStart` (setup state at session open).
+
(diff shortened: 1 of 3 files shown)


claude/.claude/skills/hook-review/SKILL.md
```diff
@@ -0,0 +1,107 @@
+---
+name: hook-review
```
jcdendrite: Rename to claude-hook-review and update other references accordingly. One of your caveats in the definition made me realize that just referring to "hook" is ambiguous.

jcdendrite: **[Claude Code]** Done — renamed to `claude-hook-review` throughout (directory, `name:` frontmatter, and both references in `code-review`). Good catch; the DO NOT TRIGGER section already called out git hooks / husky / lefthook, so the ambiguity was real.

jcdendrite (commented):
lgtm

PR #76, opened 2026-04-30
Hook: corrective note when git -C is used from main working tree

jcdendrite:
## Summary

When an agent runs `git -C <path> <write-op>` from the main working tree, the worktree-enforcement hook correctly denies the write — but the deny message gives no hint that `-C` is the problem. The hook checks the session-persisted CWD from the tool-input JSON, not the `-C` path, so the op is blocked even when `-C` points at a linked worktree.

The chained `cd && git` mistake already has a tailored note (`cwd_anchor_note_if_chained`). This PR adds a parallel note for the `git -C` case, making the hook's feedback symmetric for both common "I'm pointing at the worktree but it's blocked" patterns.

## What changed

- Added `git_C_note_if_present` function in `require-worktree-for-git-writes.sh`, mirroring `cwd_anchor_note_if_chained`
- Detection is scoped to the **global** `-C` flag position only — subcommand-level uses (`git commit -C HEAD`, `git diff -C`, etc.) do not trigger the note (the loop stops at the first non-flag word/subcommand)
- Wired into both `emit_deny` call sites (parse-failure and allowlist-miss)
- Note is appended after the existing chained-cd note; both can coexist

## Tests

Five new tests in `TestRequireWorktreeForGitWrites`:

| Test | Input | Expected |
|------|-------|----------|
| `test_git_dash_C_write_appends_C_note` | `git -C /tmp commit -m foo` | deny; `-C path` in reason |
| `test_git_dash_C_readonly_no_C_note` | `git -C /tmp log` | allow |
| `test_plain_git_no_C_note` | `git commit -m foo` | deny; `-C path` NOT in reason |
| `test_subcommand_dash_C_no_C_note` | `git commit -C HEAD` | deny; `-C path` NOT in reason (false-positive guard) |
| `test_chained_cd_and_git_C_appends_both_notes` | `cd /tmp && git -C /tmp commit -m foo` | deny; both `chained 'cd` and `-C path` in reason |

All 59 existing `TestRequireWorktreeForGitWrites` tests continue to pass.

### DIFF
--- a/claude/.claude/hooks/require-worktree-for-git-writes.sh
+++ b/claude/.claude/hooks/require-worktree-for-git-writes.sh
@@ -48,6 +48,54 @@ cwd_anchor_note_if_chained() {
   fi
 }
 
+# When the command uses `git -C <path> <write-op>` from the main tree,
+# the agent likely expected the -C path to be treated as the working
+# tree. The hook checks the session-persisted CWD (from the tool input
+# JSON) — not the -C path — so the op is blocked even when -C points at
+# a linked worktree. Returns a self-correction note for the agent when
+# this pattern is detected; empty otherwise.
+#
+# Detection is scoped to `-C` in the GLOBAL flag position only —
+# subcommand-level uses (`git commit -C HEAD`, `git diff -C`, etc.)
+# do not trigger the note, since the hint would not apply there.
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #82, opened 2026-05-02
respond-pr: document marker lifecycle to prevent mid-session cleanup

jcdendrite:
## Summary

- Adds a **Marker lifecycle** paragraph to step 0 of the respond-pr skill
- Documents two failure modes that caused real token burn in a session:
  1. Another skill's cleanup step (e.g., `/ready-for-review`) removing the respond-pr bypass marker before all `gh` API calls complete — subsequent calls are blocked by the gate
  2. Attempting `touch marker && gh api ...` as one compound Bash call fails because the PreToolUse hook fires before the shell executes the compound — the marker doesn't exist at hook-check time

## Test plan

- [ ] Read the updated step 0 — confirm the lifecycle paragraph follows the existing explanation without breaking the `HOOK_TEST_FIXTURE` comment sentinel structure
- [ ] Verify the enable-bypass fenced block is still the only block with the `enable-bypass` fixture tag (the paragraph is prose, not a code block)


jcdendrite (commented):
lgtm

PR #112, opened 2026-05-02
statusline: fix rounding, shrink bar, dynamic path/branch truncation

jcdendrite:
## Summary
- Round `h_pct`/`d_pct` via `printf "%.0f"` so float values like `7.0000000000000001` render as integers
- Shrink progress bar from 20 to 10 chars wide, freeing ~10 visible chars
- Replace hardcoded path/branch truncation limits with `stty size </dev/tty`-based allocation: path gets 55% and branch name 45% of chars remaining after fixed elements, scaling with actual terminal width on each invocation. Falls back to `$COLUMNS`, then 80, when the controlling TTY is unavailable. (`tput cols` was tried first but fails silently when the hook subprocess has stdout redirected.)

## Test plan
- [ ] Verify rate limit percentages display as integers (e.g. `5h:22% 7d:7%`)
- [ ] Verify path and branch expand when the terminal window is widened
- [ ] Verify both truncate gracefully on a narrow terminal

### DIFF
--- a/claude/.claude/statusline-command.sh
+++ b/claude/.claude/statusline-command.sh
@@ -59,8 +59,8 @@ rate_color() {
 }
 
 if [ -n "$rate_5h" ] || [ -n "$rate_7d" ]; then
-    h_pct="${rate_5h:-0}"
-    d_pct="${rate_7d:-0}"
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #113, opened 2026-05-02
Add SubagentStart session capture, settings model/effort guard, and plan-review gate

### LINKED ISSUE #87: F-07: Hook on `git commit` for settings.json model/effort changes
Today, casual mid-session changes to `model` or `effortLevel` in `claude/.claude/settings.json` get accidentally committed. Add a PreToolUse hook (matcher `Bash(git commit *)`) that:
1. Diffs staged `claude/.claude/settings.json` against `main`.
2. If `model` or `effortLevel` keys changed, prompt for explicit confirmation ("intentional?").
3. Default = block; "yes" = let through.

**Acceptance:** Modify `model` mid-session, `git commit` prompts. Answer "no" → blocked. Answer "yes" → through. Internal filtering per the defense-in-depth rule (don't rely solely on the matcher).

### LINKED ISSUE #89: F-06: Register `capture-session-id.sh` on `SubagentStart`
Currently `settings.json:3-12` registers `capture-session-id.sh` only on `SessionStart`. Sub-agents don't get session-id captured, so marker-creating skills invoked inside a Task-spawned sub-agent fail silently. `SubagentStart` is a real Claude Code event (see hooks docs).

Add a `SubagentStart` registration alongside `SessionStart`. Verify the hook's stdin payload includes `session_id` for subagents (docs imply yes via "common input fields"). Test by spawning a subagent that runs a marker-creating skill recipe.

Verify subagent-spawned hooks can write under `~/.claude/sessions/`. If they can't, the fallback is to capture session-id at the parent's PreToolUse on `Task` invocations.

**Acceptance:** Spawn a sub-agent, run `/code-review` inside it, marker file populates correctly, gated `git commit` accepts the marker.

### LINKED ISSUE #90: F-05: Hook-enforce plan-review (mirror code-review pattern)
Add `require-plan-review.sh` PreToolUse hook gating `Write`/`Edit` on implementation files when a plan exists in `.claude/plans/` without a recent plan-review marker. Mirror the marker pattern used by `require-code-review.sh`.

Open question: scope of "implementation files" — likely `/home/jared/MyCode/<project>/**` minus docs. Make it opt-in via an env var or repo-level config so non-claude-config projects aren't forced into the gate.

**Acceptance:** Hook tests pass; smoke test confirms editing `src/foo.ts` without a plan-review marker for the active plan denies.

jcdendrite:
## Summary

- **F-06 / Closes #89**: Register \`capture-session-id.sh\` on \`SubagentStart\` alongside the existing \`SessionStart\` entry. Sub-agents now get session-id captured so marker-creating skills invoked inside a subagent work correctly.

- **F-07 / Closes #87**: Add \`guard-settings-model-effort.sh\` — a PreToolUse hook that fires on \`git commit\` and blocks when \`claude/.claude/settings.json\` has staged \`model\` or \`effortLevel\` changes relative to \`main\`. Prevents accidentally committing ephemeral per-session \`/config\` overrides. Registered with \`"if": "Bash(git commit *)"\`.

- **F-05 / Closes #90**: Add \`require-plan-review.sh\` — a PreToolUse hook that gates \`Write\`/\`Edit\` when a plan file exists in \`.claude/plans/\` without a current plan-review marker for this session. Globally applied (no opt-in), consistent with \`require-code-review.sh\`, \`require-ready-for-review.sh\`, and \`require-respond-pr.sh\` — the plan-file existence check is the per-project filter. Mirrors the per-session marker pattern from \`require-code-review.sh\`.

All hooks implement defense-in-depth (filter by tool name internally; do not rely solely on \`settings.json\` \`if\` conditions).

## Test plan

- [x] \`pytest claude/.claude/hooks/tests/\` — 354 tests passing (11 new for F-07, 11 new for F-05 including the fail-closed no-session-id case)
- [x] \`ruff check claude/.claude/hooks/tests/\` — clean
- [ ] \`bash -n\` syntax check on both new hook scripts — clean
- [ ] Verify \`SubagentStart\` block mirrors \`SessionStart\` shape exactly in \`settings.json\`
- [ ] Verify \`guard-settings-model-effort.sh\` does not fire on non-\`git commit\` commands or when \`settings.json\` is not staged
- [ ] Verify \`require-plan-review.sh\` gates any repo with a populated \`.claude/plans/\` directory (no opt-in needed)

### DIFF
--- a/claude/.claude/hooks/require-plan-review.sh
+++ b/claude/.claude/hooks/require-plan-review.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
-# PreToolUse hook: block Write/Edit on implementation files when a plan file
-# exists in .claude/plans/ without a current plan-review marker for this session.
+# PreToolUse hook: block Write/Edit when a plan file exists in .claude/plans/
+# without a current plan-review marker for this session.
 #
-# Opt-in: only active when CLAUDE_REQUIRE_PLAN_REVIEW=1 is set OR a
-# .claude/require-plan-review sentinel file exists in the repo root.
-# This prevents non-claude-config projects from being forced into the gate.
+# Globally applied (no opt-in), consistent with require-code-review.sh,
+# require-ready-for-review.sh, and require-respond-pr.sh.
+# Projects without a .claude/plans/ directory or with no plan files pass
+# through silently — the plan-existence check is the built-in filter.
 #
 # Marker layout:
 #   ~/.claude/plan-review-markers/<repo-hash>.<session_id>
@@ -27,21 +28,8 @@ if [ "$TOOL_NAME" != "Write" ] && [ "$TOOL_NAME" != "Edit" ]; then
   exit 0
 fi
 
-# Check opt-in: env var or sentinel file.
-REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
-
-OPT_IN=0
-if [ "${CLAUDE_REQUIRE_PLAN_REVIEW:-}" = "1" ]; then
-  OPT_IN=1
-elif [ -n "$REPO_ROOT" ] && [ -f "$REPO_ROOT/.claude/require-plan-review" ]; then
-  OPT_IN=1
-fi
-
-if [ "$OPT_IN" -eq 0 ]; then
-  exit 0
-fi
-
 # Not in a git repo — can't check for plan files or key the marker.
+REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
 if [ -z "$REPO_ROOT" ]; then
   exit 0
 fi
--- a/claude/.claude/hooks/require-plan-review.sh
+++ b/claude/.claude/hooks/require-plan-review.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+# PreToolUse hook: block Write/Edit on implementation files when a plan file
+# exists in .claude/plans/ without a current plan-review marker for this session.
+#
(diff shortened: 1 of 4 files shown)


claude/.claude/hooks/require-plan-review.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+# PreToolUse hook: block Write/Edit on implementation files when a plan file
+# exists in .claude/plans/ without a current plan-review marker for this session.
+#
+# Opt-in: only active when CLAUDE_REQUIRE_PLAN_REVIEW=1 is set OR a
+# .claude/require-plan-review sentinel file exists in the repo root.
+# This prevents non-claude-config projects from being forced into the gate.
+#
```
jcdendrite: Why is plan review opt-in when the other markers are not? I don't think we should deviate from protocol for just one marker. The review markers are intended to be used for skills/steps of the SDLC that Claude should not skip. I don't know why we would treat plan-review any different from others. The SDLC steps with markers (code-review, ready-for-reviewer, respond-pr) are intended to be globally-applied.

jcdendrite (commented):
lgtm

claude/.claude/hooks/require-plan-review.sh
jcdendrite: **[Claude Code]** Agreed — the opt-in was inconsistent with how the other three gates work. Removed it in the follow-up commit. The plan-file existence check in `.claude/plans/` is now the per-project filter, same pattern as `require-code-review.sh`.

Issue #104, opened 2026-05-02
F-21: Specialist-agent output budget (≤2K tokens, structured)


jcdendrite:
Specialist reviewer agents currently return narrative findings that can run long. Add to `code-review/SKILL.md` and `plan-review/SKILL.md`: "Specialist agents must return ≤2K tokens of structured findings (checklist-item-keyed bullets), not narrative."

**Acceptance:** A `/code-review` invocation with 4 specialists returns under 8K tokens of agent output.

jcdendrite:
Both A1 (PR #117) and A2 (PR #115) halves merged. Specialist output budget (≤2K tokens with safety valve) is now live in both code-review and plan-review SKILL.md.

Issue #99, opened 2026-05-02
F-17: Mirror plan-review ciso-reviewer skip path in code-review


jcdendrite:
`plan-review/SKILL.md:190` (Step-2 user-surface declaration) lets ciso skip dev-only flows. `code-review/SKILL.md:182` always fires ciso. Mirror the skip path so dev-only diffs don't pull in the security pass unnecessarily.

**Acceptance:** A commit to a dev-only auth helper file (clearly declared) doesn't fire `ciso-reviewer`.

Issue #97, opened 2026-05-02
F-16: Dedupe overlapping findings in code-review/plan-review reconciliation


jcdendrite:
Add to Reconciliation section in both skills: "If two findings reference the same `file:line` with the same root cause, present once with both reviewer attributions."

**Acceptance:** A schema diff that triggers backend + analytics + data each flagging the same column produces one merged finding, not three.

jcdendrite:
Both A1 (PR #117) and A2 (PR #115) halves merged. Dedupe sentence is now live in Reconciliation sections of both code-review and plan-review SKILL.md.

Issue #96, opened 2026-05-02
F-15: Encode refined schema fan-out rule in code-review + plan-review


jcdendrite:
Per the refined rule:

| Change | Reviewers |
|---|---|
| New nullable column | backend |
| New index | backend |
| New view | backend |
| **New table** | backend + analytics |
| Rename, drop, type change | backend + data + analytics |
| NOT NULL added, partition key, RLS policy | backend + data + analytics |

Update `code-review/SKILL.md:197` (schema row in spawn table); mirror in `plan-review`.

**Acceptance:** Three test diffs (additive column / new table / column rename) trigger the right reviewer set.

jcdendrite:
Both A1 (PR #117) and A2 (PR #115) halves merged. Schema fan-out routing is now live in both code-review and plan-review SKILL.md.

PR #115, opened 2026-05-02
plan-review: agent selection guidance, schema routing, dedupe rule, output budget

### LINKED ISSUE #95: F-13: Update plan-mode Phase-1 guidance for design-doc audits
Plan-mode currently prescribes `Explore` agents in Phase 1. But `Explore`'s description warns against using it for design-doc auditing or cross-file consistency (it reads excerpts and misses content). Add to plan-mode guidance: "if the task is design-doc auditing or cross-file consistency, use `general-purpose` instead."

**Acceptance:** Plan-mode docs reference both `Explore` and `general-purpose` with clear use-case split.

jcdendrite:
## Summary

- **F-13 (Closes #95):** Step 1 now warns that `general-purpose` must be used (not `Explore`) for codebase research during plan context gathering; `Explore` reads excerpts and misses content past its read window, making it unsuitable for design-doc auditing or cross-file consistency checks.
- **F-15 (Refs #96):** Bold schema-change routing callout added after the domain list in Step 1, keying on change type (nullable column/index/view → backend only; new table → backend + analytics; rename/drop/type/NOT NULL/partition key/RLS → backend + data + analytics; + product if user-visible). Identical wording to A1's code-review addition.
- **F-16 (Refs #97):** First sentence added in Reconciliation — same `file:line` same-root-cause findings from two specialists are presented once with both reviewer attributions. Identical wording to A1's code-review addition.
- **F-21 (Refs #104):** Specialist output budget (≤2K tokens, checklist-item-keyed bullets, not narrative prose) added after the dispatch table and before Reconciliation. When findings genuinely exceed the budget, agents must prioritize by severity and explicitly note that lower-severity items were omitted. Identical wording to A1's code-review addition.
- **F-17 (#99):** Verified — ciso-reviewer skip clause already present and in good shape. No changes made; omitted from this PR.

## Test plan

- [ ] Read the updated Step 1 and confirm the `Explore`-vs-`general-purpose` note is discoverable at the point where a practitioner selects an agent for codebase research
- [ ] Read the schema routing callout and confirm change-type rows cover the full spectrum (additive, new table, destructive/structural)
- [ ] Read the Reconciliation section and confirm the dedupe sentence appears as the first sentence after the opening paragraph, before the "two readings" list
- [ ] Read the Reviewer roles section and confirm the output budget sentence appears after the dispatch table and before `## Reconciliation`, with the safety-valve sentence for over-budget findings
- [ ] Confirm the "repeated because skills load on demand" duplication notes remain intact (deliberate, not removed)
- [ ] Confirm `ciso-reviewer` skip clause at line 192 is unchanged

### DIFF
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -208,7 +208,7 @@ When you spawn: pick the specialist that serves the question (table below is ref
 
 Project-level plan-review skills may extend this table with project-specific reviewer roles, but must not remove or narrow the `ciso-reviewer` trigger conditions.
 
-Specialist agents must return ≤2K tokens of structured findings (checklist-item-keyed bullets), not narrative prose. When spawning, include this constraint in the agent prompt.
+Specialist agents must return ≤2K tokens of structured findings (checklist-item-keyed bullets), not narrative prose. If findings genuinely exceed the budget, the agent must prioritize by severity and explicitly note that lower-severity items were omitted. When spawning, include this constraint in the agent prompt.
 
 ## Reconciliation
 
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -22,7 +22,7 @@ Find the plan to review. Check, in order:
 
 ## Step 1 — Detect domains
 
-Read the plan and classify which domains it touches:
+Read the plan and classify which domains it touches. When using an agent to explore the codebase for plan context, use `general-purpose` — not `Explore`. The `Explore` agent reads excerpts and is not suitable for design-doc auditing or cross-file consistency checks; it misses content past its read window.
 
 - **Infrastructure**: CI/CD, workflows, deployment, hosting, config files
 - **Data infrastructure**: Database migrations, schema DDL, RLS policies, CDC / change-stream config, ETL/ELT pipelines, warehouse ingestion connectors, raw landing schemas, schema-drift handling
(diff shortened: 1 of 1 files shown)


claude/.claude/skills/plan-review/SKILL.md
```diff
@@ -22,7 +22,7 @@ Find the plan to review. Check, in order:
 
 ## Step 1 — Detect domains
 
-Read the plan and classify which domains it touches:
+Read the plan and classify which domains it touches. When using an agent to explore the codebase for plan context, use `general-purpose` — not `Explore`. The `Explore` agent reads excerpts and is not suitable for design-doc auditing or cross-file consistency checks; it misses content past its read window.
```
jcdendrite, line 25: Is the explanatory line starting with "The `Explore` agent reads excerpts and is not suitable..." important for Claude or important for skill editors? If only the latter, put it in REFERENCE.md for this skill as an explanatory note.

claude/.claude/skills/plan-review/SKILL.md
```diff
@@ -206,11 +208,13 @@ When you spawn: pick the specialist that serves the question (table below is ref
 
 Project-level plan-review skills may extend this table with project-specific reviewer roles, but must not remove or narrow the `ciso-reviewer` trigger conditions.
 
+Specialist agents must return ≤2K tokens of structured findings (checklist-item-keyed bullets), not narrative prose. When spawning, include this constraint in the agent prompt.
```
jcdendrite: What if there are more than 2K tokens of structured findings? How do we ensure those don't get lost? Or what if this 2K limitation dilutes the substance of the reviews?

claude/.claude/skills/plan-review/SKILL.md
jcdendrite: **[Claude Code]** The explanatory clause is load-bearing for Claude, not just for editors. Without "reads excerpts and is not suitable for design-doc auditing," the instruction reduces to a bare "use general-purpose not Explore" — Claude needs the failure mode to make the right call in context (e.g., when Explore seems like the obvious fast choice for codebase research). Kept in SKILL.md.

claude/.claude/skills/plan-review/SKILL.md
jcdendrite: **[Claude Code]** Good catch. The 2K budget is a conciseness constraint, not a hard truncation instruction — agents structuring findings as keyed bullets rarely hit it in practice (5–8 findings at a few sentences each). But the instruction as written gave no guidance for the case where real findings do exceed it.

Fixed in a follow-up commit: agents that genuinely hit the limit must now prioritize by severity and explicitly note that lower-severity items were omitted, rather than silently dropping them. The commit also applies the same safety valve to `code-review/SKILL.md` (PR #117) to keep them in sync.

jcdendrite (commented):
lgtm

PR #118, opened 2026-05-03
hook: gate segmentless pulls/comments/{id} to close PATCH-overwrite gap

jcdendrite:
## Summary

- **Root cause:** `require-respond-pr.sh` matched `(pulls|issues)/[0-9]+/(comments|reviews)` but the destructive edit URL `pulls/comments/{id}` has no PR-number segment — the regex never matched, so the hook exited 0 (allow) before the bypass-marker check even ran. The failed session's post-mortem attributed the miss to the bypass marker being active; that was incorrect.
- **Enforcement fix:** Added a second `elif` clause matching `repos/{o}/{r}/(pulls|issues)/comments/{id}` (segmentless edit form); extended the cross-repo `sed` extractor to parse the repo slug from the new URL shape.
- **Knowledge fix:** Revised `SKILL.md` Step 4 to give one explicit reply path per comment type (inline → `/replies`; top-level reviews and issue-level → `POST /issues/{n}/comments`); updated the Attribution example; replaced the Guidelines PATCH bullet's broken `user.login` check with an attribution-prefix check (`**[Claude Code]**`) — the only reliable way to distinguish Claude-authored comments from the user's own when both use the same GitHub token.
- **Tests:** 8 new parametrize entries covering denied-without-marker, allowed-with-marker, non-matching shapes, and cross-repo bypass for the new URL family; all green, ruff clean.

## Test plan

- [ ] `pytest claude/.claude/hooks/tests/test_hooks.py::TestRequireRespondPr -v` — all 41 pass, including skill ↔ hook alignment tests (HOOK_TEST_FIXTURE blocks untouched)
- [ ] `ruff check claude/.claude/hooks/tests/` — clean
- [ ] Manual smoke (no marker): `printf '{"tool_input":{"command":"gh api repos/foo/bar/pulls/comments/12345 -X PATCH -F body=x"},"session_id":"test"}' | ~/.claude/hooks/require-respond-pr.sh; echo $?` — expect non-zero
- [ ] Manual smoke (fresh marker): create `~/.claude/.respond-pr-active.d/test`, re-run above — expect 0


jcdendrite (commented):
lgtm

PR #121, opened 2026-05-04
docs: auto mode activation, hard-floor deny rules, settings.local.json template

jcdendrite:
## Summary

- Adds a full **Auto mode** section to README covering requirements, activation, the two-layer classifier pipeline, and tuning commands
- Documents the `permissions.deny` hard-floor rules (sudo, .env, credentials.json) shipped in `settings.json` and why they complement the classifier
- Provides a `settings.local.json` `autoMode.environment` template with a `"$defaults"` retention warning and the broad-allow-rules-drop gotcha
- Updates the `settings.json` bullet in the "Other" config table to reflect the new deny block

`settings.json` change: adds the `permissions.deny` block. These rules run before the classifier and cannot be overridden by `autoMode.allow` — they harden the sudo and secret-read gaps that were previously advisory-only in `CLAUDE.md`.

## Test plan

- [ ] Verify `Read(**/.env)` and `Bash(sudo *)` entries in `claude/.claude/settings.json` match the documented table
- [ ] Confirm `"$defaults"` warning in the template is accurate against the [auto-mode-config reference](https://code.claude.com/docs/en/auto-mode-config#override-the-block-and-allow-rules)
- [ ] Confirm activation instructions (Shift+Tab, `--permission-mode auto`, `defaultMode` setting) are current
- [ ] Spot-check `claude auto-mode defaults/config/critique` command names against CLI help

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Schema-change diffs nominally route three ways — `staff-backend-engineer` (des
 ### Other
 
 - **`CLAUDE.md`** — baseline engineering instructions (judgment heuristics, working style, safety rules).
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #125, opened 2026-05-04
docs: add skill-SKILL.md row to rule-routing decision flow

jcdendrite:
## Summary

- Adds a new row to the §4 quick-decision table routing context-specific rules (those that only fire inside a particular skill's flow) to that skill's `SKILL.md` rather than `CLAUDE.md`.
- Adds a matching row to the §5 "Where does a given rule belong?" table with the same destination.

Both additions introduce `skill's SKILL.md` as a third routing destination alongside `CLAUDE.md/AGENTS.md` and auto-memory, closing the gap for rules like heredoc quoting behavior that only apply during a specific skill's execution.

## Test plan

- [ ] Read §4 table and confirm the new row sits between "Am I adding a new guardrail?" and "The repo has CLAUDE.md but no AGENTS.md".
- [ ] Read §5 table and confirm the new row sits between "Personal preference or workflow specific to this user" (Auto-memory) and "Past incident 'why'...".
- [ ] Confirm no other content was changed.


jcdendrite (commented):
lgtm

PR #129, opened 2026-05-04
plan-review B5: require verification of cited code shapes

jcdendrite:
## Summary

- Strengthens `B5. Evidence` → `B5. Evidence and verification` in `plan-review/SKILL.md`
- Adds explicit requirement: when a plan asserts a specific code shape (function signature, exact line number, type field, import path), the reviewer must verify the cited source actually matches — citation alone is insufficient
- Aligns the Item ownership table shorthand with the renamed check

## Why

Plans that assert unverified code shapes — invented signatures, wrong line numbers, invalid type fields — produce phantom code-review findings when implementers correctly diverge from the bad assertion. The evidence check already required citations; it now also requires that citations be verified against the assertion.

B1 (unstated assumptions) was scoped to SDK/library behavior. B13 (ambiguous instructions) was about misinterpretability, not factual correctness. Neither caught the "clear but wrong" case. B5 is the right home.

## Test plan

- [ ] `pytest claude/.claude/` — all expected to pass
- [ ] `ruff check claude/.claude/` — no issues
- [ ] Diff: only two lines change (B5 body, shorthand in Item ownership table)

_Note: `claude/.claude/scripts/` was introduced in a prior merge; no new top-level stow entries in this PR. If merging onto a fresh clone, run `./install.sh` to link all top-level entries._

### DIFF
--- a/claude/.claude/skills/plan-review/SKILL.md
+++ b/claude/.claude/skills/plan-review/SKILL.md
@@ -86,7 +86,7 @@ B3. **Breaking intermediate states** — During phased migrations, is there a wi
 
 B4. **Unresolved external dependencies** — Does the plan depend on external services, APIs, or tools whose availability, rate limits, or behavior the author hasn't verified?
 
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #130, opened 2026-05-04
feat: enforce 200-line SKILL.md ceiling via pre-commit hook

jcdendrite:
## What

- **New hook `check-skill-length.sh`** — PreToolUse:Bash gate that blocks `git commit` when a staged SKILL.md grows past the 200-line limit. Policy: deny when `new > 200 AND new > old`, so already-over-limit files can be reduced commit-by-commit without being blocked.
- **13 pytest cases** covering: non-commit commands, non-Bash tools, no staged SKILL.md, exactly-at-200, new-file-over-limit, growing/reducing/same-size already-over-limit, staged deletion, deny message contents, and a CWD regression test (the prior `[ -f "$f" ]` guard was CWD-dependent and caused false negatives when the hook ran from a repo subdirectory).
- **Settings registration** — hook added to PreToolUse > Bash array in `settings.json`.
- **`ready-for-review/SKILL.md` trimmed 212 → 200** — five prose compressions (session-chain failure note, pre-existing failures callout, test-to-fit rule, cumulative review rationale, no-PR note), all behavior-preserving.

## Why

`skill-review`'s own rule: *"When a rule can be encoded mechanically, prefer mechanical enforcement."* The 200-line ceiling existed as advisory text only. This PR makes it a commit gate — same enforcement point and failure shape as `require-code-review.sh`.

## What was reviewed

- Per-commit `/code-review` on each staged diff before committing
- Cumulative `/code-review` (full branch-vs-main diff) before PR open
- `claude-hook-review` subagent on the hook script
- All three review rounds produced findings that were fixed before landing

## Test results

```
428 passed in ~21s
ruff: All checks passed
```

## Existing PRs

No prior PRs address skill length enforcement via a pre-commit hook. The 200-line limit has been referenced in prose (skill-review skill) but never mechanically enforced.

### DIFF
--- a/claude/.claude/hooks/check-skill-length.sh
+++ b/claude/.claude/hooks/check-skill-length.sh
@@ -22,8 +22,8 @@ REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
 FAIL=0
 MESSAGES=""
 while IFS= read -r f; do
-  new=$(git show ":$f" 2>/dev/null | wc -l) || continue
-  old=$(git show "HEAD:$f" 2>/dev/null | wc -l || echo 0)
+  new=$(git show ":$f" 2>/dev/null | awk 'END{print NR}')
+  old=$(git show "HEAD:$f" 2>/dev/null | awk 'END{print NR}')
   if [ "$new" -gt 200 ] && [ "$new" -gt "$old" ]; then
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

PR #131, opened 2026-05-04
fix: deny chained `cd ... && git` regardless of session cwd

jcdendrite:
## Summary

- **Bug:** A session whose persisted cwd was a linked worktree could run `cd /main-repo && git merge ... && git push ...` and bypass the worktree enforcement hook. The hook read the session-persisted cwd (worktree shape → exit 0) before ever reaching the fragment loop. The merge landed on local main unblocked.
- **Root cause:** `require-worktree-for-git-writes.sh` short-circuits at the worktree-shape check (line 134) when the persisted cwd is a linked worktree. An inline `cd /main-tree` in the same command runs in the subshell *after* the hook exits — the hook never sees it.
- **Fix:** Extract `command_chains_cd_then_git()` predicate (reuses the existing regex from `cwd_anchor_note_if_chained`) and call it **before** the worktree-shape early-exit. Deny unconditionally when the pattern matches — the effective cwd at the time git runs cannot be determined from the tool-input JSON alone.

## Why not parse the `cd` target path?

The handoff doc proposed extracting the cd target and re-running the worktree check against it. That approach has hidden parsing complexity: tilde/env-var/quote/relative/command-substitution/multiple-`cd` cases all require shell expansion the hook can't safely replicate. Failing-open on parse failure would preserve the bypass for those cases. The refuse-to-evaluate approach is simpler, closes both directions of the bypass (worktree→main and main→worktree), and matches the advice the hook already gives in its self-correction note.

**Trade-off:** A benign `cd /sibling-worktree && git commit` from a worktree-persisted session is also denied. Fix is one standalone `cd` Bash call — same advice the hook already emits.

## Test plan

- [x] 7 new regression tests: merge and push from worktree-cwd chained cd → deny; `;` and `||` operators → deny; readonly subcommand chained cd → deny (shape-based, not subcmd-based); plain git from worktree → allow; non-cd chain from worktree → allow
- [x] Updated `test_chained_cd_and_git_C_appends_both_notes` → `test_chained_cd_and_git_C_only_cd_chain_deny`: early-deny fires before fragment loop, so -C note is never appended
- [x] `pytest claude/.claude/` — 454 passed (up from 415)

### DIFF
--- a/claude/.claude/hooks/require-worktree-for-git-writes.sh
+++ b/claude/.claude/hooks/require-worktree-for-git-writes.sh
@@ -31,19 +31,23 @@ emit_deny() {
     "$reason_json"
 }
 
+# Returns success when the command chains `cd ... <op> ... git ...`. The
+# hook reads cwd from Claude Code's session-persisted bash state (set by
+# prior Bash calls), not from the inline cd in the current command, because
+# the hook fires before the subshell runs. When this shape is detected the
+# effective cwd at the time git runs cannot be determined from the tool-input
+# JSON alone. Pattern requires `cd` as a word, then a chain operator, then
+# `git` somewhere later — avoids false positives on paths containing `cd`.
+command_chains_cd_then_git() {
+  printf '%s' "$1" | grep -qE '(^|[[:space:]])cd[[:space:]].*(&&|\|\||;).*git'
+}
+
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #137, opened 2026-05-05
fix: scope require-plan-review gate to repo-internal writes

jcdendrite:
## Summary

- `require-plan-review.sh` was denying **all** `Write`/`Edit`/`MultiEdit` calls — anywhere on disk — when the repo had unreviewed plans. This collaterally blocked plan-mode from authoring its own scratch plan file at `~/.claude/plans/<session-slug>.md`, since the harness controls that path and a hook can only deny (not redirect).
- Fix: scope the deny to writes resolving inside `$REPO_ROOT`. Out-of-repo writes (user-home, `/tmp`, sibling repos) pass through.
- The gate's semantics are now: *"this repo's code is gated by this repo's planning"* — which is what it was always intended to mean.

## Changes

**`require-plan-review.sh`** — adds a path-scope filter after the marker checks conclude no valid marker exists. Uses `realpath -m` on both sides for symlink-consistent comparison; trailing-slash prefix match prevents sibling-repo false positives; empty `file_path` falls through to deny (fail-closed).

**`test_require_plan_review.py`** — two changes:
1. Updated 10 existing deny-without-marker tests from `/tmp/foo.py` to `<repo>/src/foo.py` — those tests were asserting the right behavior but had drifted to a path that now passes through.
2. Added 5 new tests: home-dir plan write pass-through, `/tmp` pass-through, in-repo deny (positive control), sibling-path prefix guard, missing `file_path` fail-closed.

## Test plan

- [ ] `pytest claude/.claude/hooks/tests/test_require_plan_review.py -v` → 26/26 pass
- [ ] `pytest claude/.claude/` → full suite green (483 passed)
- [ ] `ruff check claude/.claude/` → all checks passed
- [ ] Manual: open a plan-mode session in a repo that has an unreviewed plan file; verify the plan-mode scratch write to `~/.claude/plans/` succeeds; verify a write to a repo file is still denied

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/claude/.claude/hooks/require-plan-review.sh
+++ b/claude/.claude/hooks/require-plan-review.sh
@@ -72,6 +72,18 @@ if [ -n "$SESSION_ID" ]; then
   fi
 fi
 
+# Scope the deny to writes inside this repo. Writes targeting user-home
+# directories (~/.claude/plans/), /tmp, or other repos are outside the gate's
+# intent — the gate guards this repo's code, not all files on disk.
+TARGET_PATH=$(printf '%s\n' "$INPUT" | jq -r '.tool_input.file_path // empty')
+if [ -n "$TARGET_PATH" ]; then
+  REAL_REPO=$(realpath -m "$REPO_ROOT")
+  REAL_TARGET=$(realpath -m "$TARGET_PATH")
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #138, opened 2026-05-05
fix: drop wallclock-time handoff heuristic

jcdendrite:
## Summary

- Removes `~3 hours of wallclock time` as a handoff trigger — wallclock is a poor proxy since idle sessions don't burn tokens while the cache stays warm
- Retains the `~300K output tokens` threshold, which is the real cost signal
- Adds `analyze-context.py` as a verification step so the threshold is checkable rather than guess-based
- Replaces the vague `/tmp/<task>-handoff.md` placeholder with an explicit instruction to use a real descriptive slug, so the user knows where to find the file
- Replaces "start a fresh session that reads it" (Claude can't do this) with "tell the user the exact path and ask them to open a new session with the prompt to paste" — honest about the human/agent responsibility split

Surfaced during a session-burn analysis where two sessions flagged by this heuristic (67h and 42h elapsed) were misleadingly described as over-threshold by wallclock when the actual driver was turn count, not idle time.


jcdendrite (commented):
lgtm

PR #141, opened 2026-05-05
feat: plan-review routing extraction + ROUTING.md consultation enforcement

jcdendrite:
## Summary

**Routing extraction (1F.iv + Step 3 re-land):**
- Adds Step 3 — Plan structure requirements — to `plan-review/SKILL.md`: no placeholders, bite-sized steps, context-complete steps. A plan failing any returns to the author before gap-finding.
- Extracts the Reviewer roles table, Reconciliation logic, and Item ownership table (83 lines) to `plan-review/ROUTING.md` behind an unconditional Read directive. All spawn criteria live exclusively in ROUTING.md. Net: 312 → 239 lines.
- `REFERENCES.md`: routing architecture rationale, smoke test result (2026-05-05).

**Consultation enforcement:**
- `log-routing-read.sh` (PostToolUse/Read): writes a per-session marker when ROUTING.md is Read during an active plan-review session.
- `require-routing-read.sh` (PreToolUse/Agent): denies Agent spawns when the plan-review session is active but the routing-read marker is absent or stale (>60 min).
- `settings.json`: wires up both hooks (first PostToolUse hook in this repo).
- `SKILL.md` output format: requires a "Spawned" block attributing each specialist to the checklist item IDs it owns (from ROUTING.md's Item ownership table).
- `deactivate-gate` fixture updated to clean up both the active-session marker and the routing-read marker.

**Tests:**
- `test_log_routing_read.py`: 8 tests for the marker-writer hook.
- `test_require_routing_read.py`: 9 tests for the deny gate (staleness, cross-session isolation, deactivate-gate fixture alignment).
- `helpers.py`: new builders and marker helpers for both hooks.

## Test plan

- [x] `pytest claude/.claude/` — 514/514 pass
- [x] `ruff check claude/.claude/` — clean
- [x] `wc -l claude/.claude/skills/plan-review/SKILL.md` — 239 lines (was 312 on main; hook gate clears because 239 < 312)
- [x] Routing smoke test (2026-05-05): subagent on synthetic 3-domain plan read ROUTING.md, spawned 7 specialists, ROUTING.md informed 4 decisions the inline section couldn't have.
- [x] Hook enforcement smoke test: `require-routing-read.sh` deny path verified in test suite; `log-routing-read.sh` marker-write path verified with worktree and stowed path variants.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #140, opened 2026-05-05
feat: token-analyzer fidelity — per-record accumulation, windowed timestamps, tighter Opus→Sonnet heuristic

jcdendrite:
## Summary

Follow-up to #135 (F-19), addressing three design issues flagged in the post-merge review.

- **Per-record token accumulation**: each assistant record's tokens are credited to its own model family rather than the session's dominant family. Mixed-model sessions (e.g. Opus parent + Sonnet sub-agents) now contribute to both families' totals. Session `fam` in top-10 and candidate list still reflects the dominant non-sidechain family.
- **Per-record `--since` windowing**: `--since Nd` now sums only tokens from records whose ISO timestamp falls within the window — long-running sessions contribute only their in-window slice. mtime stays as a cheap file-level pre-filter. Section headers relabeled to `(activity in the last Nd)` to make the semantics explicit.
- **Tighter Opus→Sonnet candidate heuristic**: adds four new exclusion signals evaluated from non-sidechain (parent) records: `Task` tool calls (dispatcher orchestration), `thinking` blocks (extended reasoning), judgment-skill invocations (`code-review`, `plan-review`, `security-review`, `skill-review`, `respond-pr`, `ultrareview`), and presence of any sidechain record. A diagnostic line above the candidate list shows excluded session counts per signal for ongoing tuning.

## Semantic note

Exclusion signals (edits, task, thinking, judgment-skill, plan) are evaluated against all records for the session lifetime, not just the in-window portion. A session that used plan-mode before the `--since` window is still correctly excluded from candidates — the Opus routing decision applied to the whole session.

## Test plan

- [x] `pytest claude/.claude/scripts/tests/test_token_analyzer.py` — 14 passed (was 5)
- [x] `ruff check claude/.claude/` — clean
- [ ] Live smoke run: `~/.claude/scripts/token-analyzer.py` — per-model totals should reflect per-record attribution; candidate list diagnostic line should show exclusion signals
- [ ] Live smoke run: `~/.claude/scripts/token-analyzer.py --since 2d` — long-running sessions should contribute only in-window tokens

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #144, opened 2026-05-05
fix(ci): broaden path filter to skills/ and scripts/, harden detect step

jcdendrite:
## Summary

- **Regex broadened**: `claude/.claude/skills/` (whole tree) and `claude/.claude/scripts/` added alongside the existing `hooks/` entry. The prior regex only matched where test *files* live — it missed that `test_skills.py` reads frontmatter from every `skills/<name>/SKILL.md`, that hook-alignment tests like `test_require_plan_review.py` read `HOOK_TEST_FIXTURE` blocks directly from skill bodies, and that `test_token_analyzer.py` exercises `scripts/token-analyzer.py`. PRs #140, #141, and #142 all silently skipped CI as a result.
- **Inclusion-rule comment added**: States what belongs in the list so the next contributor knows when to extend it (not just copy a file path pattern).
- **Shell hardening**: `set -e` → `set -euo pipefail`; fail-closed on unset variables and broken pipes.
- **Drift visibility**: Both matched and unmatched file lists are now logged on every run. When tests are skipped on a future PR, the unmatched list in the Actions log answers "why" without anyone re-running the regex by hand.

## Architecture note (for future reference)

The in-job conditional shape is the correct GitHub Actions pattern here. Workflow-level `paths:` triggers leave required status checks in `Pending` indefinitely (GH community discussion #44490, in backlog since Jan 2023, no GitHub timeline). `dorny/paths-filter` solves the N-independent-jobs aggregator problem and carries solo-maintainer supply-chain risk (sibling `tj-actions/changed-files` was compromised March 2025). The fix here is hardening the existing shape, not switching architectures.

**If this filter goes stale again** (e.g., `agents/` or `CLAUDE.md` gains test coverage), the right move at that point is to delete the filter and run tests on every PR — the saved CI minutes do not justify another round of silent skips.

## Test plan

- [ ] Merge this PR — CI on this PR's own diff should trigger `Run tests` (`.github/workflows/hooks.yml` is in the filter).
- [ ] After merge, push an empty commit to re-trigger CI on PR #141 and confirm `pytest claude/.claude/ -v` actually runs against the `plan-review/SKILL.md` changes there.
- [ ] Confirm the Actions log now shows "Matched (will run tests):" and "Unmatched (skipped):" sections on every run.
- [ ] Confirm branch protection still passes — job id `hook-tests` is unchanged.

### DIFF
--- a/.github/workflows/hooks.yml
+++ b/.github/workflows/hooks.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Detect hook-relevant changes
         id: detect
         run: |
-          set -e
+          set -euo pipefail
           if [ "${{ github.event_name }}" = "pull_request" ]; then
             BASE="${{ github.event.pull_request.base.sha }}"
             HEAD="${{ github.event.pull_request.head.sha }}"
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #150, opened 2026-05-06
fix(skill): tighten cleanup-merged-branch against worktree constraint failures

jcdendrite:
## Summary

- **Upfront precondition added**: new third bullet at the top of the skill body states that Steps 4 and 5 require executing from inside a linked worktree. Previously the constraint was scattered and only surfaced mid-step.
- **Step 5 loophole removed**: the `# Call 1 — anchor CWD (skip if already in a worktree)` caveat gave small models an out to skip the cd and run `git merge --ff-only` from the main tree, where the worktree-required hook blocks it. Now the cd is mandatory and idempotent. The load-bearing role of `-C <REPO_ROOT>` (temp worktree is detached HEAD — bare `git merge` would target the wrong ref) is now explained inline.
- **Cleanup ordering fixed**: temp worktree cleanup has been moved from end of Step 4 to end of Step 5, so the worktree is guaranteed open during the fast-forward. The cleanup block now uses the same two-call CWD-anchoring pattern as other write ops, with `getcwd: cannot access parent directories` explained at the point of failure.

## Motivation

A Haiku session running `/cleanup-merged-branch` after a merged PR hit both failure modes: blocked by the worktree hook on `git merge --ff-only`, and a partial worktree cleanup due to `getcwd` failure. Haiku's self-assessment attributed both to "constraint awareness" but the skill body was the actual culprit — one loophole invited the skip, and the cleanup CWD gap was genuinely absent from the skill.

## Test plan

- [x] `pytest claude/.claude/ -q` — 514 passed
- [x] `/skill-review` — clean, behavioral-equivalence audit passed on all removed lines
- [x] `/code-review` — clean
- [ ] Manual smoke: next time `/cleanup-merged-branch` runs under Haiku, verify no hook denial during Step 5 fast-forward and no partial worktree cleanup

### DIFF
--- a/claude/.claude/skills/cleanup-merged-branch/SKILL.md
+++ b/claude/.claude/skills/cleanup-merged-branch/SKILL.md
@@ -15,7 +15,7 @@ argument-hint: "[branch-name]"
 
 # Cleanup merged branch
 
-This repo squash-merges PRs and enforces worktree discipline. Two
+This repo squash-merges PRs and enforces worktree discipline. Three
 constraints shape every step:
 
 - **Always use `git branch -D`** (force), not `-d`. Squash merges don't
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #153, opened 2026-05-06
feat: prose-pointer composition for project-layer skills + DRY hygiene item

jcdendrite:
## Summary

- Adds **Step 0.5** to `code-review` and **Step 2.5** to `plan-review`: each globs for `.claude/skills/<parent>-*/SKILL.md` from the repo root and invokes a single matching project layer via the Skill tool. Zero cost in projects without a layer. Resolves F-04 (description-based auto-trigger does not fire from inside a running skill).
- Adds **item 9** (Repeated in-house logic / DRY check with DAMP carve-out for tests) to `code-review` base checklist; renumbers items 9–34 → 10–35 throughout body and ownership table.
- Extends **B16** with a one-sentence calibration for solo-scale teams: default to surgical fix, because follow-up tickets rarely land without sprint ceremony.
- Adds **B17** (Plan and implementation in sync): plan files in `.claude/plans/` must ship in the same PR as their implementation.
- Updates `plan-review/ROUTING.md`: adds B17 to the judgment items row.
- Documents the composition design in `code-review/REFERENCES.md` (F-05).

## Design rationale

The global skills must stay project-agnostic — inlining project-specific content here would impose noise on projects that don't have a layer, and would publish project-shape fingerprints that the public-repo redaction rule prohibits. The prose-pointer mechanism (glob + Skill-tool invocation) is the only composition path that auto-fires reliably: Skill-tool invocation works mid-skill; description-based auto-trigger does not (F-04).

Token cost is conditional: the glob no-matches in projects without a layer; only projects that have adopted the `<skill>-<name>/` naming convention pay for their layer.

## Test plan

- [ ] In a project with a matching layer (`code-review-*/SKILL.md`), run `/code-review` and confirm Step 0.5 invokes the project layer via Skill tool
- [ ] In this repo (no matching layer), run `/code-review` and confirm the glob no-matches and the skill proceeds without a layer invocation
- [ ] Same verification for `/plan-review` + Step 2.5
- [ ] Confirm item 9 (DRY) fires on a diff with repeated logic ≥5 lines at 3+ sites; does NOT flag test-file duplication (DAMP carve-out)
- [ ] Plant a second `code-review-foo/SKILL.md` in a test project; confirm the multi-match path stops with a config-error message

### DIFF
--- a/claude/.claude/skills/code-review/REFERENCES.md
+++ b/claude/.claude/skills/code-review/REFERENCES.md
@@ -16,3 +16,13 @@ Reference material that informed this skill. Not loaded during skill execution �
 **Result:** No system-reminders for `claude-hook-review`, `skill-review`, or `review-permissions` appeared. Only `code-review` itself loaded. Auto-trigger did **not** fire from inside a running skill.
 
 **Policy decision:** The prose pointers at `code-review/SKILL.md` ("see the `skill-review` skill" and "see the `claude-hook-review` skill") are load-bearing. Without them, specialist skills are silently skipped when code-review runs on their surfaces. Keep the pointers. F-14 item 1 (standardize the voice of those pointers) is the next step.
+
+## F-05: Project-layer composition via prose-pointer + glob (2026-05-05)
+
+**Trigger:** A review shipped a 7-site DRY violation despite multiple `/code-review` passes. Investigation found two gaps: (1) no Hygiene item for in-house logic duplication, (2) a project-specific code-review layer (`user-invocable: false` + description-trigger) had never fired — F-04 explains why.
+
+**Decision:** Wire project layers via explicit prose pointer (Step 0.5 in code-review, Step 2.5 in plan-review) using a glob `git rev-parse --show-toplevel`/`.claude/skills/<parent>-*/SKILL.md`. Single match → invoke via Skill tool. Multiple matches → stop (config error in the project, not something the review skill resolves). Zero matches → proceed without a layer.
+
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

conversation in terminal with Sonnet:

> [Me] For the explanatory prose you removed, do you think it won't make sense to add that in
>   REFERENCES.md?
> 
> [Sonnet] ● Good question. Looking at each cut:
> 
>   Already in REFERENCES.md (no action needed):
>   The two "Why this is here" paragraphs (Step 0.5 in code-review, Step 2.5 in plan-review) both
>   pointed at F-04, which already documents why description-based auto-trigger doesn't fire from
>   inside a running skill. F-05 (added in the prior commit) captures the glob-pointer decision. The
>   rationale is already there — the SKILL.md prose was just a redundant in-file pointer to it.
> 
>   Not REFERENCES.md material:
>   - Core principle framing ("review the ripple effects, not just the change") — editorial
>   tone-setting, not a decision record or empirical finding
>   - Schema-touching diffs preview — a routing rule fully restated in the change-type table
>   downstream; nothing to preserve
>   - "Same escalation logic as plan-review's Reviewer roles..." — meta-commentary about why text is
>   duplicated; the duplication itself is the design, and the reason (skills load on demand) is
>   already in the general REFERENCES.md F-04 entry
>   - "After delivering the verdict, write the completion marker..." — structural preview of the next
>    two paragraphs; no decision to record
> 
>   Borderline: the closing "don't manually clean up" note — the 60-minute staleness cutoff is an
>   implementation detail of require-plan-review.sh. If that ever changes, a note in
>   plan-review/REFERENCES.md would help a future editor. But plan-review doesn't currently have a
>   REFERENCES.md, and this is a weak enough case that I'd leave it unless the cutoff becomes
>   load-bearing.
> 
>   Short answer: no. The "Why this is here" rationale (the only stuff with real decision content)
>   was already in REFERENCES.md before we cut it from the SKILL.md.

PR #158, opened 2026-05-06
feat(skills): require grep before asserting "canonical" pattern

jcdendrite:
## Summary

- **plan-it Step 3** — adds a load-bearing paragraph requiring `git grep` + call-site count before any plan asserts a code shape is "canonical," "the existing pattern," or "how the codebase does X." Surfaced during a feature plan where a single cited call site was promoted to "canonical" without verifying the population; the implementer followed the plan literally into seven wrong-shape functions that a PR reviewer then bounced.
- **plan-review B5** — expands the evidence-and-verification item to explicitly cover the pattern-claim case: verify the cited example represents the broader population, not just that the citation is accurate.
- **No auto-memory change** — the rule now lives in load-bearing skill bodies (hook-gated); a memory entry restating it would be wasted context budget per established convention.

The duplication across both skills is intentional: plan-it guards at write time, plan-review guards at review time. Each can fire without the other; defense-in-depth applies (skill-review §6 three-condition test holds).

## Test plan

- [x] `pytest claude/.claude/` — 512 passed
- [x] `/skill-review` — clean (additions-only diff; all new lines pass behavior test; duplication justified by §6)
- [x] `/code-review` — no issues found
- [x] Commit passed `deny-private-project-refs.sh`


claude/.claude/skills/plan-it/SKILL.md
```diff
@@ -35,6 +35,8 @@ Restate the problem, why now, and the intended outcome in one short paragraph. I
 
 Find similar features, the target subsystem, and integration points. Spawn `general-purpose` subagents in parallel when scope warrants — judge fan-out from surface area, do not default to a fixed count. Read the files each subagent flags before designing. Do not use `Explore` here; its read-excerpt window is wrong for design-context analysis.
 
+**Pattern claims require a grep, not a single example.** Before the plan asserts that any code shape is "canonical," "the existing pattern," "how the codebase does X," or equivalent, run `git grep` (or ripgrep) for the shape across the relevant directory and count call sites. Cite the count in the plan ("12 of 13 edge functions use form X; the one exception is `path/to/file.ts:NN`") rather than naming a single example. A single-call-site citation establishes that the shape compiles, not that it is canonical — outliers and one-offs read identically to canonical examples until you check the population.
```
jcdendrite: I think you can tighten this language.

Also, adding specific technologies like edge functions and a TS file gives me pause. Is that example helpful or does it anchor the agent to think in terms of one specific technology? Would it be better to make this more general?

claude/.claude/skills/plan-review/SKILL.md
```diff
@@ -96,7 +96,7 @@ B3. **Breaking intermediate states** — During phased migrations, is there a wi
...
+B5. **Evidence and verification** — Does the plan cite a source for each finding/assertion (file:line, tool that flagged it, or how it was discovered)? When the plan asserts a specific code shape (function signature, exact line number, type field, import path), verify the cited source actually matches — citation alone is insufficient if the assertion misquotes it. **When the plan asserts that a shape is "canonical," "the existing pattern," "the convention," or equivalent, verify the cited example represents the broader population — run `git grep` for the shape and check the count. A single-call-site citation does not establish a pattern; outliers and one-offs read identically to canonical examples until the population is checked.** Conclusions without evidence force reviewers to re-derive them; misquoted or unrepresentative citations produce phantom code-review findings when the implementer correctly diverges.
```
jcdendrite: I think you could tighten this language a bit.

claude/.claude/skills/plan-it/SKILL.md
jcdendrite: **[Claude Code]** Fixed in 85ea6cb. Two changes applied: (1) tightened the prose throughout ("Before asserting" vs "Before the plan asserts that any code shape is," removed "or equivalent" since it is implied, dropped the directory-scope qualifier); (2) replaced "edge functions" and `path/to/file.ts:NN` with "modules" and `path/to/file:NN` to keep the example tech-agnostic. `disposition: fixed-as-requested`

claude/.claude/skills/plan-review/SKILL.md
jcdendrite: **[Claude Code]** Fixed in 85ea6cb. Tightened the bolded addition: removed "that a" (redundant), dropped "or equivalent" (implied by the list), shortened "for the shape and check the count" to "and check the count," and cut the trailing rationale clause ("outliers and one-offs...") since the diagnostic "does not establish a pattern" carries the finding and the fuller rationale already lives in plan-it where the author needs it. `disposition: fixed-as-requested`

jcdendrite (commented):
lgtm

PR #161, opened 2026-05-06
chore: remove skill-creator from enabledPlugins + design doc

jcdendrite:
## Summary

- Removes `skill-creator@claude-plugins-official` from `enabledPlugins` entirely (not set to `false` — no quick-flip handle warranted; see rationale below)
- Creates `claude/.claude/skills/skill-review/REFERENCES.md` capturing why the eval/benchmark machinery doesn't fit this setup
- Updates README.md L11 to accurately describe the plugin stance (marketplace wired in, all official plugins disabled by default)

## Why this is a removal, not a `false` entry

In this repo, `enabledPlugins: false` entries are quick-flip handles — kept for plugins the user might want to re-enable for an occasional session. `skill-creator` has no foreseeable re-enable use case for a single-user skill catalog with deterministic `/skill-review` enforcement, so the entry is removed. `claude-md-management` and `claude-code-setup` retain their `false` handles.

## Stow blast radius

`claude/.claude/settings.json` is stow-distributed. This disables `skill-creator` for any user who clones and stows the repo. Downstream stow users who want it back can add `"skill-creator@claude-plugins-official": true` to their own `enabledPlugins`.

## Test plan

- [x] `python3 -c 'import json; json.load(open("claude/.claude/settings.json"))'` — JSON valid
- [x] `pytest claude/.claude/` — 512 passed (including `test_do_not_trigger_names_adjacent_skill[skill-review-skill-creator]` — test depends on naming not enablement, passes unchanged)
- [ ] Open a fresh Claude Code session, prompt "create a new skill for X" — `skill-creator` should not auto-invoke

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/claude/.claude/skills/skill-review/REFERENCES.md
+++ b/claude/.claude/skills/skill-review/REFERENCES.md
@@ -1,0 +1,55 @@
+# skill-review — References
+
+## Why skill-creator is disabled in this repo
+
+`skill-creator` (Anthropic's `claude-plugins-official` plugin) and
+`skill-review` (this repo's local skill) have a deliberate lane split:
+`skill-creator` owns the *authoring* loop — scaffolding a new skill
+directory, running eval/benchmark pipelines, and optimizing the
+`description` field via a `run_loop.py` search loop. `skill-review`
+owns the *audit* lane — the 11-item pre-commit checklist and the
+behavioral-equivalence table that `require-skill-review.sh` enforces
+at commit. The boundary was hand-tightened in commit `618c9ad`, which
+added `skill-creator` to `skill-review`'s DO NOT TRIGGER block to
+prevent dual-fire on every skill-edit turn.
+
+`skill-creator`'s eval/benchmark machinery is the right tool when a
+skill ships at scale: diverse phrasings from many users mean trigger
+accuracy is hard to verify by reading, mis-fires happen in sessions
+you don't observe, and a 20-query eval run pays for itself in reduced
+regressions. The `run_loop.py` description optimizer makes sense when
+you genuinely can't tell whether a wording change improves accuracy
+without measuring it.
+
+None of those conditions apply here. This repo is a single-user skill
+catalog — each stow clone is independent, mis-fires are observed in
+the same session, and the description can be tuned by reading and
+testing directly. The skill set is small and curated. `skill-review`'s
+deterministic checklist (`check-skill-length.sh` at 200 lines,
+`require-skill-review.sh` at commit) already provides stronger
+pre-commit signal than probabilistic evals for this workflow.
+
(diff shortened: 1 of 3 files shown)


claude/.claude/skills/skill-review/REFERENCES.md
```diff
@@ -0,0 +1,55 @@
...
+`skill-creator`'s body also conflicts with this repo's conventions in
+two concrete places: it advocates 500-line skill bodies (its L92,
+L96), while `check-skill-length.sh` caps at 200; and it uses a
+conversational, first-person voice ("Cool? Cool.") that `skill-review`
+explicitly rejects in favor of imperative second-person. Keeping it
+enabled would present contradictory authoring guidance every time a
+skill was edited.
+
+This decision mirrors the reasoning behind the no-CI-eval-harness
+policy recorded in `feedback_ci_eval_harness_tradeoffs.md` — the same
```
jcdendrite: Memory is individual-user scoped. This comment is going to make no sense outside of my setup. I recommend removing it unless there's content in the memory itself that is really relevant here.

claude/.claude/skills/skill-review/REFERENCES.md
```diff
@@ -0,0 +1,55 @@
...
+`claude -p` evaluation loop shape was rejected for CI on security,
+budget, and flakiness grounds. The local equivalent shares those
+tradeoffs.
+
+**Mechanism:** the `skill-creator@claude-plugins-official` entry is
+removed from `enabledPlugins` in `settings.json` entirely — not set to
+`false`. In this repo, `false` entries are quick-flip handles for
+plugins the user might want to enable for an occasional session;
+removing the entry means there's no foreseeable re-enable use case.
+`claude-md-management` and `claude-code-setup` retain their `false`
```
jcdendrite: Irrelevant skills to mention here, since they are unrelated to skill-review.

claude/.claude/skills/skill-review/REFERENCES.md
jcdendrite: **[Claude Code]** Removed the paragraph entirely — the surrounding text makes the case without needing the memory-file cross-reference. `disposition: fixed-as-requested` | `commit_sha: 03084d9`

claude/.claude/skills/skill-review/REFERENCES.md
jcdendrite: **[Claude Code]** Removed — kept only the re-enable instruction, which is the load-bearing part of the Mechanism paragraph. `disposition: fixed-as-requested` | `commit_sha: 03084d9`

jcdendrite (commented):
lgtm

PR #162, opened 2026-05-06
feat: replace inline marker recipes with marker.sh script

jcdendrite:
## Summary

- Adds `~/.claude/scripts/marker.sh` — a single script that centralises all 10 marker write/activate/deactivate operations across five workflow skills, replacing the repeated inline bash recipes in each SKILL.md
- Adds `enforce-marker-script-shape.sh` PreToolUse:Bash hook — fail-closed gate that denies any `marker.sh` invocation outside the 10 exact valid shapes (no chains, no env-var prefixes, no redirects, no bash wrappers, no path traversal)
- Adds 10 exact-shape `permissions.allow` entries to `settings.json` for silent auto-approval of all valid marker operations
- Drops `marker_chain_note_if_detected()` from `require-code-review.sh` and `require-skill-review.sh` — the strict-shape hook owns chain prevention; the legacy `>` redirect pattern is dead-letter post-migration

## Motivation

Agents were repeatedly prompted to approve long dynamic bash chains for marker operations, and occasionally ran wrong variants. The inline recipes also duplicated `tr -d '\n'` logic across 10 locations. One script + one allowlist entry replaces all of that.

## Security model (three-layer defense)

1. `settings.json` exact-shape `allow` entries — silent approval for the 10 valid shapes; nothing else is silently allowed
2. `enforce-marker-script-shape.sh` — fail-closed hook denies any command containing `marker.sh` that doesn't match the anchored regex; includes path-traversal pre-check (`..` blocked before regex)
3. `marker.sh` own arg validation — explicit `case` whitelisting; unknown subcommand/skill exits 2 with usage

## What changed

| Surface | Change |
|---|---|
| `claude/.claude/scripts/marker.sh` | New: central write-side wrapper |
| `claude/.claude/hooks/enforce-marker-script-shape.sh` | New: strict-shape gate |
| `claude/.claude/settings.json` | Hook wired to PreToolUse:Bash; 10 allow entries added |
| `claude/.claude/hooks/require-code-review.sh` | Dropped `marker_chain_note_if_detected` (~20 lines) |
| `claude/.claude/hooks/require-skill-review.sh` | Same |
| `claude/.claude/hooks/tests/helpers.py` | `SCRIPTS_DIR` constant; `run_skill_command` symlinks `marker.sh` into sandboxed `$HOME` |
| `claude/.claude/hooks/tests/test_enforce_marker_script_shape.py` | New: 28 tests covering all 10 valid shapes + all denial cases |
| `claude/.claude/hooks/tests/test_marker_script.py` | New: 15 tests for marker.sh own behaviour (missing session file exits 2, no stray markers written, happy path) |
| `test_require_code_review.py`, `test_require_skill_review.py` | Removed chain-detection tests (behaviour now owned by enforce hook) |
| 5 × SKILL.md | 10 fixture body replacements; 2 prose trims |
| `README.md` | `chmod +x` glob broadened; `marker.sh` entry added |

## Bugs found and fixed during review

**1. Dead unreachable `--help` branch** (`fdd54c0`): The `$# -ne 2` arg-count guard fired before the `--help|-h` case was reachable. Fixed by adding an explicit `--help`/`-h` check before the count guard.

**2. `exit 2` inside command-substitution subshell silently continued parent** (`228abd0`): `exit 2` inside `_resolve_session_id()` only exits the command-substitution subshell — without `set -e`, the parent script continued with `SESSION_ID=""`, writing a marker with an empty session-id suffix no hook can match. Fixed with `|| exit 2` on every `SESSION_ID=$(...)` and `REPO_HASH=$(...)` call. Added `test_marker_script.py` to pin the invariant.

**3. Ancestor-scan needed for production PPID** (`05a3c7e`): When marker.sh runs as an executable (via shebang), the OS execs a new bash whose `$PPID` is the Bash tool shell — one level above the Claude Code process that owns the session file. Fixed `_resolve_session_id` to walk up the ancestor chain with `ps -o ppid=` until finding a PID with a session file.

## Test plan

- [x] 558 tests pass (`pytest claude/.claude/`)
- [x] `ruff check` clean
- [x] `bash -n claude/.claude/scripts/marker.sh` syntax-clean
- [x] `marker.sh --help` exits 0
- [x] `marker.sh write nonsense` exits 2 + usage on stderr
- [x] End-to-end: `/ready-for-review` ran to completion, `git push` gate passed

## Follow-up

A separate PR will extract a shared `_lib.sh` from all five `require-*.sh` hooks and `marker.sh` to DRY up repo-hash and session-id computation.

### DIFF
--- a/claude/.claude/hooks/tests/helpers.py
+++ b/claude/.claude/hooks/tests/helpers.py
@@ -15,6 +15,7 @@ from pathlib import Path
 HOOKS_DIR = Path(__file__).resolve().parent.parent
 
 SKILLS_DIR = HOOKS_DIR.parent / "skills"
+SCRIPTS_DIR = HOOKS_DIR.parent / "scripts"
 
 # SKILL.md fences may be indented when the fixture sits inside a
 # numbered list (e.g. respond-pr's "0. **Enable hook bypass.**"). The
@@ -247,6 +248,11 @@ def write_plan_review_routing_read_marker(home: Path, session_id: str) -> Path:
 
 def run_skill_command(command: str, cwd: Path, isolated_home: Path) -> None:
     """Run a SKILL.md-extracted bash command in a sandboxed $HOME."""
+    scripts_dir = isolated_home / ".claude" / "scripts"
+    scripts_dir.mkdir(parents=True, exist_ok=True)
+    marker_link = scripts_dir / "marker.sh"
+    if not marker_link.exists():
+        marker_link.symlink_to(SCRIPTS_DIR / "marker.sh")
     subprocess.run(
         ["bash", "-c", command],
         cwd=cwd,
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Utility scripts in `claude/.claude/scripts/` (stowed to `~/.claude/scripts/`).
 **First-time setup:** after pulling this repo for the first time, make scripts executable:
 
 ```bash
-chmod +x ~/.claude/scripts/*.py
+chmod +x ~/.claude/scripts/*
 ```
 
 - **`analyze-context.py`** — inspect context window growth for a Claude Code session. Reads `~/.claude/projects/<project>/<session>.jsonl` and `~/.claude/usage-data/session-meta/` locally; no network calls, no writes.
@@ -195,6 +195,8 @@ chmod +x ~/.claude/scripts/*.py
   ~/.claude/scripts/token-analyzer.py             # all-time
   ~/.claude/scripts/token-analyzer.py --since 7d  # include token activity from the last N days (e.g. 2d, 7d)
   ```
+
+- **`marker.sh`** — write and remove review markers on behalf of workflow skills. `/code-review`, `/skill-review`, `/plan-review`, `/ready-for-review`, and `/respond-pr` write review markers via `~/.claude/scripts/marker.sh`. The 10 valid invocation shapes are allowlisted in `settings.json` for silent auto-approval. A companion `enforce-marker-script-shape.sh` hook denies any other invocation (chains, env-var prefixes, redirects) to prevent prompt-injection escalation via the allowlist.
 
 ## Worktree enforcement
 
--- a/claude/.claude/hooks/enforce-marker-script-shape.sh
+++ b/claude/.claude/hooks/enforce-marker-script-shape.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+# Gate: enforce strict invocation shape for ~/.claude/scripts/marker.sh.
+#
+# Fail posture: fail-closed — if jq cannot parse the input, deny.
+#
+# WARNING: Do NOT remove the internal marker.sh check below.
+# The "if" field in settings.json is unreliable — it has been observed
+# to fire this hook on ALL Bash commands. The internal grep is the actual
+# gate. The "if" field is a hint only.
+#
+# Any Bash command containing "marker.sh" must be one of the 10 valid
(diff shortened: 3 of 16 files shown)


claude/.claude/hooks/tests/helpers.py
```diff
@@ -15,6 +15,7 @@
 HOOKS_DIR = Path(__file__).resolve().parent.parent
 
 SKILLS_DIR = HOOKS_DIR.parent / "skills"
+SCRIPTS_DIR = HOOKS_DIR.parent / "scripts"
```
jcdendrite, line 18: I think this is follow-up for a new PR, but I'm realizing that this helpers file probably needs to live in a parent tests folder, not in the hooks/tests folder. Not sure what is canonical to Anthropic's file structure - that's probably worth some web research and documentation lookup.

claude/.claude/hooks/tests/helpers.py
```diff
@@ -247,6 +248,11 @@ def write_plan_review_routing_read_marker(home: Path, session_id: str) -> Path:
 
 def run_skill_command(command: str, cwd: Path, isolated_home: Path) -> None:
     """Run a SKILL.md-extracted bash command in a sandboxed $HOME."""
+    scripts_dir = isolated_home / ".claude" / "scripts"
+    scripts_dir.mkdir(parents=True, exist_ok=True)
+    marker_link = scripts_dir / "marker.sh"
+    if not marker_link.exists():
+        marker_link.symlink_to(SCRIPTS_DIR / "marker.sh")
```
jcdendrite, line 255: Why are you sym linking? That seems like it's a bandaid fix to an underlying design flaw. I could be wrong though. Please explain.

claude/.claude/hooks/tests/helpers.py
jcdendrite: **[Claude Code]** Agreed — `helpers.py` is now shared across hook and script test contexts (`SCRIPTS_DIR` is the tell). Moving it to a parent `tests/` directory (e.g., `claude/.claude/tests/helpers.py`) is the right fix; it would need ~17 import-site updates in the hook tests plus a conftest or path adjustment.

Out of scope for this PR — the move benefits from its own isolated commit history. Tracked in #164.

`disposition: DEFERRED` | `deferral-reason: test directory restructure is independent of the marker recipe migration and warrants its own PR` | `follow-up-ticket: https://github.com/jcdendrite/claude-config/issues/164`

claude/.claude/hooks/tests/helpers.py
jcdendrite: **[Claude Code]** Not a bandaid — the symlink is the right mechanism for the isolation model.

`run_skill_command` sandboxes `$HOME` to a temp directory so tests cannot write to the real `~/.claude/` state. The SKILL.md commands it runs use `~/.claude/scripts/marker.sh`; with `HOME=<sandboxed>`, `~` expands to the sandboxed path, and `<sandboxed>/.claude/scripts/marker.sh` does not exist without the symlink.

Alternatives and why they are worse:
- **Copy the file** — goes stale after any marker.sh edit; a test could pass against a stale binary of the script it is supposed to be testing.
- **Do not sandbox `$HOME`** — skill commands would write real markers to `~/.claude/`; parallel pytest runs would pollute each other.
- **Absolute path in SKILL.md** — bakes a machine-specific path into every skill file; breaks on any other machine.

The symlink exposes exactly one file from the real filesystem into the sandbox, preserves full HOME isolation, and never goes stale. `where-documented: helpers.py:251-255`

jcdendrite (commented):
lgtm

PR #165, opened 2026-05-06
feat(scripts): replace cleanup-merged-branch skill with no-arg discovery script

jcdendrite:
## Summary

- Adds `claude/.claude/scripts/cleanup-merged-branches.sh`: a no-arg discovery script that queries `gh pr list --head <branch> --state merged` for every local branch and cleans up any with a merged PR (worktree remove, `branch -D`, remote branch delete, ff-only default branch). Supports `--dry-run`. Cron-friendly (silent on no-op when non-TTY).
- Adds 21-case pytest suite at `claude/.claude/scripts/tests/test_cleanup_merged_branches.py` with a PATH-shim `fake_gh` fixture.
- Adds `permissions.allow` with two exact-string entries (`~/.claude/scripts/cleanup-merged-branches.sh` and `... --dry-run`) to `settings.json` alongside the existing marker.sh entries — no PreToolUse hook needed since no wildcards are in play.
- Deletes `claude/.claude/skills/cleanup-merged-branch/SKILL.md` and its directory.
- Drops `(e.g. /cleanup-merged-branch)` from the Haiku routing rule in `claude/.claude/CLAUDE.md`.
- Moves the script entry in `README.md` from the skills list to the scripts section.

## Why

The skill required a dedicated Haiku terminal kept open just for post-merge cleanup. The procedure is 100% deterministic: always the same steps, always on the merged branch. A discovery-based script eliminates the session-warmup cost, removes the argument-injection attack surface (no user-supplied branch name means nothing to validate against destructive ops), and makes the operation schedulable.

The worktree-required hook's word-boundary check for `git` doesn't fire on a single `bash <script>` invocation, collapsing the multi-Bash-call CWD dance the skill required into a single-process script with no worktree contortions.

## Behaviors preserved from the deleted SKILL.md

- `git branch -D` (force-delete handles squash merges) — hardcoded in script
- Worktree path lookup via `git worktree list --porcelain` matching `branch refs/heads/<BRANCH>` — not constructed from branch name (test case 6: slashed branch name)
- `--force` on `git worktree remove` for open/uncommitted files, `|| true` to swallow missing-path errors
- `git fetch --prune` then check `refs/remotes/origin/<BRANCH>` existence to detect auto-deleted remote refs (ref-existence check avoids substring match false-positives from sibling branches)
- Fast-forward default branch only when behind; uses `merge --ff-only` when HEAD is on the default branch, `fetch refspec` when invoked from a feature branch (avoids mutating the checked-out branch); handles non-`main` defaults (test case 10: `master`)

## Bugs fixed during code review

Two correctness issues found and fixed before merge:

1. **Remote-prune false positive** — `grep -qF "origin/${BRANCH}"` was a substring match: a fetch pruning `origin/feat/foo-v2` would set `REMOTE_AUTO_PRUNED=1` for `feat/foo`, silently skipping its remote deletion. Fixed by checking `git rev-parse --verify refs/remotes/origin/${BRANCH}` after the fetch — an exact ref lookup.

2. **ff-update on wrong branch** — `git merge --ff-only origin/main` runs against the currently checked-out branch, not the default branch. Invoking the script from a feature branch either produced a misleading "manual pull needed" or (if the feature branch is an ancestor of origin/main) silently advanced the wrong branch. Fixed with a conditional: `merge --ff-only` when HEAD is already on the default branch, `git fetch origin main:main` (refspec form) when on a feature branch.

## Test plan

- [ ] `pytest claude/.claude/scripts/tests/test_cleanup_merged_branches.py -v` — 21 cases, all green
- [ ] `ruff check claude/.claude/` — no violations
- [ ] Smoke: run `~/.claude/scripts/cleanup-merged-branches.sh` from a repo with a recently merged branch; verify worktree gone, local branch gone, remote gone, default ff'd
- [ ] Smoke: `--dry-run` prints candidates without acting
- [ ] Smoke: `!~/.claude/scripts/cleanup-merged-branches.sh` in Claude session — no permission prompt (exact-string allow match)
- [ ] Smoke: `!~/.claude/scripts/cleanup-merged-branches.sh foo` — allow rule doesn't match; Claude prompts or denies
- [ ] Confirm `/cleanup-merged-branch` no longer appears in slash-command catalog after harness reload


README.md
```diff
@@ -106,7 +106,7 @@ flowchart LR
...
 - **`/verify-primary-sources`** — when web research informs a code or design decision, read the primary documentation directly rather than trusting agent summaries or secondary sources.
 - **`/read-docx-comments`** — extract comments from `.docx` files with anchored text context.
-- **`/cleanup-merged-branch`** — removes the local worktree and branch, prunes remote tracking refs, and fast-forwards the default branch after a PR is merged. Handles squash-merge branch detection and CWD anchoring for worktree-enforced repos.
+- **`~/.claude/scripts/cleanup-merged-branches.sh`** — script that discovers all local branches whose PRs are merged (queried via `gh pr list --head <branch> --state merged`) and cleans them up: removes the worktree, force-deletes the local branch, prunes the remote tracking ref, deletes the remote branch if not auto-deleted, and fast-forwards the default branch. Run from any shell (or via `!` from any Claude session — auto-approved by the paired `permissions.allow` entries). Pass `--dry-run` to preview without acting.
```
jcdendrite: Whoops, you put this in the skill section. Should be in scripts now.

jcdendrite (commented):
You have merge conflicts with settings.json

README.md
jcdendrite: **[Claude Code]** Fixed. The entry has been moved from the skills list to the Scripts section of the README, where it belongs alongside `analyze-context.py`, `token-analyzer.py`, and `marker.sh`. `disposition: fixed-as-requested`

jcdendrite:
**[Claude Code]** Fixed. The branch was forked before the marker.sh PR (#162) landed, so `settings.json` diverged when that PR added 10 `marker.sh` allow entries and the `enforce-marker-script-shape.sh` hook. Resolved by rebasing onto main and keeping both sets of entries — the existing marker.sh entries from main plus the two new `cleanup-merged-branches.sh` entries. Tests (579 passed) and lint clean on the rebased branch. `disposition: fixed-as-requested`

jcdendrite (commented):
lgtm

PR #167, opened 2026-05-06
refactor(tests): move helpers.py to shared claude/.claude/tests/ directory

### LINKED ISSUE #164: refactor(tests): move helpers.py to a parent tests/ directory
## Background

`claude/.claude/hooks/tests/helpers.py` was originally scoped to hook tests only. PR #162 added `SCRIPTS_DIR` and extended `run_skill_command` to symlink `marker.sh` into the sandboxed HOME — making it a shared utility for both hook and script tests.

## Problem

The file is mislocated. Its current path (`hooks/tests/helpers.py`) implies it is hook-specific, but it now also supports tests under `scripts/`. This will become a recurring friction point as the `scripts/` directory grows.

## Proposed fix

Move `helpers.py` (and its fixtures/constants) to a shared parent, e.g., `claude/.claude/tests/helpers.py`. Update all import sites (~17 hook test files) and add a `conftest.py` or `pytest.ini` path adjustment so `from helpers import ...` continues to resolve.

## Scope note

Deliberately scoped as a separate PR so the rename has a clean commit history and doesn't tangle with functional changes.

jcdendrite:
## Summary

- Moves `claude/.claude/hooks/tests/helpers.py` → `claude/.claude/tests/helpers.py` via `git mv` (preserves blame history)
- Refactors path constants to anchor from `CLAUDE_DIR` (`.claude/` root) rather than the old `HOOKS_DIR.parent` chain — cleaner now that the file no longer lives under `hooks/`
- Adds `[tool.pytest.ini_options] pythonpath = ["claude/.claude/tests"]` to `pyproject.toml` so the 19 existing `from helpers import ...` statements in hook-test files resolve unchanged
- Extends the CI path-filter regex in `hooks.yml` to include `claude/.claude/tests/` so changes to the relocated file still trigger the test suite

No changes to any of the 19 hook-test import sites or to `conftest.py`. `pytest claude/.claude/ -v` → 558 passed.

Closes #164

## Test plan

- [x] `pytest claude/.claude/ -v` — 558 passed (verified locally before commit)
- [x] `git log --follow claude/.claude/tests/helpers.py` — blame preserved
- [ ] CI "Hook tests" check passes on this PR (path filter now covers `claude/.claude/tests/`)

### DIFF
--- a/.github/workflows/hooks.yml
+++ b/.github/workflows/hooks.yml
@@ -55,7 +55,7 @@ jobs:
           # Include any directory that contains test files or whose contents are
           # read by tests. Update this list when a new test or test-input
           # directory appears under claude/.claude/.
-          REGEX='^(claude/\.claude/(hooks|skills|scripts)/|\.github/workflows/hooks\.yml|pyproject\.toml)'
+          REGEX='^(claude/\.claude/(hooks|skills|scripts|tests)/|\.github/workflows/hooks\.yml|pyproject\.toml)'
 
           MATCHED=$(echo "$CHANGED" | grep -E "$REGEX" || true)
           UNMATCHED=$(echo "$CHANGED" | grep -vE "$REGEX" || true)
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #168, opened 2026-05-06
refactor: extract _marker_lib_repo_hash shared helper from require-* hooks and marker.sh

jcdendrite:
## Summary

- Adds `claude/.claude/hooks/_lib.sh` — a sourced bash library exporting `_marker_lib_repo_hash()`, the single source of truth for the marker repo-hash recipe
- Updates four hooks (`require-code-review.sh`, `require-skill-review.sh`, `require-plan-review.sh`, `require-ready-for-review.sh`) to source the lib and replace their inline one-liners
- Updates `marker.sh` to source the lib via `$HOME/.claude/hooks/_lib.sh` and call `_marker_lib_repo_hash` inside `_resolve_repo_hash()`
- Adds `test_marker_lib.py` with 3 tests pinning the function's output against Python `hashlib.sha256`, the original inline recipe, and a falsifiable newline-regression check

## What problem this solves

Before PR #162, both the read side (hooks) and write side (`marker.sh`) computed the repo-hash independently. PR #162 added a comment documenting the `tr -d '\n'` / `printf '%s'` equivalence, but a comment alone can't prevent drift. This PR makes drift structurally impossible: there is now one recipe and both sides reference it.

`require-respond-pr.sh` is untouched — it does not compute a repo-hash.

## Scope rationale

The predecessor plan also proposed sharing session-id resolution and marker-path construction. After reviewing the merged PR #162:

- Session-id resolution cannot be shared: hooks parse a JSON payload via `jq`; `marker.sh` walks the process ancestor chain via `ps -o ppid=`. Different mechanisms by necessity.
- Marker-path literals don't compute anything and can't drift.

This PR covers the one genuine drift surface and nothing else.

## Test plan

- [x] `bash -n claude/.claude/hooks/_lib.sh` — syntax clean
- [x] 561 tests pass (`pytest claude/.claude/ -x -q` — 3 new tests added)
- [x] `ruff check` clean
- [x] Code review passed
- [ ] End-to-end smoke: `/code-review` → marker written → `git commit` allowed (verified post-push by CI)

### DIFF
--- a/claude/.claude/hooks/_lib.sh
+++ b/claude/.claude/hooks/_lib.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Shared helper library sourced by require-*.sh hooks and scripts/marker.sh.
+# Keep this file the single source of truth for any recipe that must produce
+# byte-identical output on both the read side (hooks) and the write side
+# (marker.sh). Source it; do not invoke it directly.
+
+# Compute the marker repo-hash for an absolute repo-toplevel path.
+# Input must have no trailing newline -- printf '%s' omits one, so the SHA
+# covers exactly the bytes of $1.
+# Usage: hash=$(_marker_lib_repo_hash "$REPO_ROOT")
+_marker_lib_repo_hash() {
+  printf '%s' "$1" | sha256sum | awk '{print $1}'
+}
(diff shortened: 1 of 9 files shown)


jcdendrite (commented):
lgtm

PR #170, opened 2026-05-06
fix(memory-hook): replace per-turn UUID debounce with active-marker bypass

jcdendrite:
## Summary

- Fixes memory gate false-positives caused by system-reminder injections advancing the per-turn UUID — each injection creates a new `type=="user"` transcript entry, so the debounce's "same turn" window never opens and writes that should pass re-deny
- Replaces UUID debounce with the active-bypass pattern (mirrors `plan-review` and `respond-pr`): the `ai-instruction-and-memory-files` skill writes `~/.claude/.<name>-active.d/<session_id>` at Step 0, hook bypasses while fresh (<60 min, mtime refreshed on each pass), skill removes marker at final step
- Extends `scripts/marker.sh` and `enforce-marker-script-shape.sh` with `activate memory-skill` / `deactivate memory-skill` (12 valid shapes total, up from 10)
- Adds Step 0 and Final step with `HOOK_TEST_FIXTURE` markers to `ai-instruction-and-memory-files/SKILL.md`; moves primary source URLs to new `REFERENCES.md` (repo convention for load-on-demand reference material) to keep SKILL.md at 200 lines

## Test plan

- [ ] 568 tests pass (`pytest claude/.claude/`)
- [ ] Ruff lint passes (`ruff check claude/.claude/`)
- [ ] Live smoke test: memory write gates correctly before skill invoked, passes after activate marker is written, re-gates after deactivate removes it
- [ ] System-reminder injection no longer re-denies writes mid-skill (the regression case this fixes)

### DIFF
--- a/claude/.claude/hooks/enforce-marker-script-shape.sh
+++ b/claude/.claude/hooks/enforce-marker-script-shape.sh
@@ -8,7 +8,7 @@
 # to fire this hook on ALL Bash commands. The internal grep is the actual
 # gate. The "if" field is a hint only.
 #
-# Any Bash command containing "marker.sh" must be one of the 10 valid
+# Any Bash command containing "marker.sh" must be one of the 12 valid
 # invocation shapes exactly. No chains, no env-var prefixes, no redirects,
 # no bash wrappers, no extra args. This prevents prompt-injection attacks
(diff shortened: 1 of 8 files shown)


jcdendrite (commented):
lgtm

PR #171, opened 2026-05-06
feat(ready-for-review): route verify-class commands through a subagent

jcdendrite:
## Summary

- **`claude/.claude/skills/ready-for-review/SKILL.md`** — Step 2 routes suite-level verification commands through `Agent` with `subagent_type: general-purpose` instead of inline Bash. The subagent writes each command's full output to `${TMPDIR:-/tmp}/<slug>-<epoch-ms>.txt` (slug = command lowercased with non-alphanumeric runs collapsed to `-`; epoch-ms via `date +%s%3N`), then returns a structured verdict (per-command status, smallest failure excerpt, overall PASS or FAIL) plus the output file paths. The parent reads the file for more detail rather than re-running or spawning a follow-up subagent.
- **`claude/.claude/CLAUDE.md`** — New `### Heavy command output` subsection under Working Style formalises the same rule globally, so ad-hoc "run the tests" requests honour it even when the skill isn't triggered.
- Inline exception preserved in both locations: commands scoped to a single test file or test name during interactive debugging may run directly in the parent.

## Why

Verify-class commands produce high-volume single-pass output that displaces parent working state and invalidates the prompt cache — and this compounds when verification runs multiple times per gate iteration. Subagent isolation keeps the parent context clean; writing output to a temp file lets the parent read more detail without re-running commands or spawning a second subagent. `${TMPDIR:-/tmp}` respects user-scoped temp dirs on macOS; epoch-ms prevents same-second collisions when commands run in parallel.

## Notes

- **Hard precondition before merge**: Plan B's Verification Step 0 (lock-hook compatibility check on the consuming project) must pass before this lands. Plan B applies the same rule to a project-scoped skill and verifies that the project's test-lock hook handles subagent execution correctly.
- Rule is fully generic; no private-project references.

## Test plan

- [x] `pytest claude/.claude/` — 568 passed (HOOK_TEST_FIXTURE blocks in steps 0 and 7 are untouched)
- [x] `/skill-review` on SKILL.md diff — clean (behavioral-equivalence table has no N rows, both commits)
- [x] `/code-review` on full staged diff — no findings (both commits)
- [x] `ready-for-review/SKILL.md` line count: 200 (at limit, not over)
- [x] Temp file naming reviewed by `staff-platform-engineer` — F3 (epoch-ms), F1 (`${TMPDIR:-/tmp}`) addressed; F2/F4 accepted trade-offs

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -19,7 +19,7 @@
 
 ### Heavy command output
 
-Verify-class runs (full test suites, lint, typecheck, build) should run via the `Agent` tool with `subagent_type: general-purpose` — not directly via `Bash` in the parent. The subagent absorbs the output and returns a verdict plus any failure excerpts; the parent's prompt cache and working state stay intact. This applies to suite-level runs; commands scoped to a single test file or single test name during interactive debugging can stay inline.
+Verify-class runs (full test suites, lint, typecheck, build) should run via the `Agent` tool with `subagent_type: general-purpose` — not directly via `Bash` in the parent. The subagent writes full output to `${TMPDIR:-/tmp}/<command-slug>-<epoch-ms>.txt` and returns a structured verdict plus the file paths; the parent reads the file for more detail rather than re-running. This applies to suite-level runs; commands scoped to a single test file or single test name during interactive debugging can stay inline.
 
 ## Code Review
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #172, opened 2026-05-07
feat(context): proactive handoff + post-compact marker re-injection

jcdendrite:
## Summary

- **Phase 0 — marker re-injection after /compact and /clear:** `session-marker-dashboard.sh` now has `"matcher": "startup|clear|compact"` in settings.json, so it re-fires after compaction and clears. Output converted from plain-text stdout to `hookSpecificOutput.additionalContext` JSON (with `hookEventName: "SessionStart"` per the documented Claude Code hook contract). Fixes the core complaint: post-compact Claude now has fresh on-disk marker knowledge.

- **Phase 1 — state digest format (`COMPACT_INSTRUCTIONS.md`):** New `claude/.claude/COMPACT_INSTRUCTIONS.md` defines a **structured digest shape** (§1 Goal → §2 Status → §3 Next concrete step → §4 Files → §5 Markers → §6 Open questions) that both `/compact` and `/handoff` produce. Both modes deliver a state digest to a context-reset agent with the same information need — the carrier differs (inline summary vs. file) but the shape is canonical. `CLAUDE.md` references the file under `## State digest format`.

- **Phase 2 — `/handoff` slash command:** New `claude/.claude/commands/handoff.md` is a thin shell over the COMPACT_INSTRUCTIONS shape. It tells Claude to write a file at `/tmp/<slug>-handoff.md` using §1–§6 from `COMPACT_INSTRUCTIONS.md`, then append §7 Resume command plus a pre-write checklist. Slash command (not a skill) to avoid loading into the always-active skills catalog. `CLAUDE.md` nudge now suggests `/handoff` at ~60% context usage — matching Anthropic's recommended manual-/compact threshold (same rationale: cleaner context = better output, and every turn beyond 60% burns tokens destined for waste).

- **Phase 3 — README:** Documents all three layers, the decision table (compact vs clear vs handoff), and the threshold guide (60% / 83.5%).

## Post-merge: re-run install.sh

This PR adds two new top-level entries under `claude/.claude/`:
- `claude/.claude/COMPACT_INSTRUCTIONS.md`
- `claude/.claude/commands/`

Stow links each top-level child individually. These new entries **will not appear in `~/.claude/` on `git pull` alone** — symlinks are only created when stow runs. After merging, run:

```bash
./install.sh
```

Without this, `/handoff` won't be discoverable and `COMPACT_INSTRUCTIONS.md` won't be referenced by CLAUDE.md at runtime.

## Key implementation decision verified from official docs

- `"matcher": "startup|clear|compact"` for `SessionStart` is the exact documented syntax — `startup`, `clear`, `compact` are valid trigger values (plus `resume`).
- `hookSpecificOutput.hookEventName` is required alongside `additionalContext` in the JSON form.
- Plain-text stdout also works for `SessionStart` context injection per docs — but the JSON form is more explicit and forward-compatible.

## Phase 1 verification needed (open question)

CLAUDE.md says "See `~/.claude/COMPACT_INSTRUCTIONS.md`" but it's unverified whether the compaction summarizer follows file references. Test by: triggering `/compact` with active markers, then asking "what markers were active before?" If the summarizer ignores the reference, inline the file contents directly under `## State digest format` in CLAUDE.md.

## Test plan

- [x] 598/598 tests pass (`pytest claude/.claude/`)
- [x] 10 dashboard tests pass including new `test_output_is_valid_json_with_correct_shape` and updated assertions
- [ ] Manual: start a session, write an active marker, run `/compact`, ask about markers — should match `ls ~/.claude/*-markers/`
- [ ] Manual: repeat with `/clear` to confirm the matcher catches both
- [ ] Manual: type `/handoff` mid-task, open the produced file in a fresh session, verify clean resume
- [ ] Manual: verify Claude suggests `/handoff` around 60% context usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ flowchart LR
 - **`/ai-instruction-and-memory-files`** — how AI coding agents load instruction files (CLAUDE.md, AGENTS.md, Cursor rules, Lovable knowledge) and Claude Code auto-memory: precedence, duplication rules, length targets, import patterns.
 - **`/verify-primary-sources`** — when web research informs a code or design decision, read the primary documentation directly rather than trusting agent summaries or secondary sources.
 - **`/read-docx-comments`** — extract comments from `.docx` files with anchored text context.
+- **`/handoff`** — write a structured cross-session handoff file at `/tmp/<slug>-handoff.md` capturing goal, status, next step, modified files, active markers, open questions, and the resume incantation (`Read /tmp/<slug>-handoff.md and continue.`). User-invoked only — implemented as a slash command at `claude/.claude/commands/handoff.md`, not a skill, to keep it out of the always-loaded skills catalog. Claude proactively suggests `/handoff` when context usage exceeds ~60% with an incomplete task; you decide whether to invoke.
 
 Each skill lives in `claude/.claude/skills/<skill-name>/SKILL.md`. A skill directory may also contain a `REFERENCES.md` — canonical references (URLs, key quotes, framework notes) that informed the skill's rules. `REFERENCES.md` is not loaded during skill execution; consult it when editing a skill to verify a rule still holds or to add new guidance.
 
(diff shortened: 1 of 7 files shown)


jcdendrite (commented):
lgtm

PR #173, opened 2026-05-07
feat(hook): add 90-min hard ceiling to ready-for-review active-marker bypass

jcdendrite:
## Summary

- **Problem:** The `require-ready-for-review.sh` active-marker bypass used a sliding mtime window (60 min, refreshed on each bypass) with no absolute lifetime cap. Cross-skill iteration pushes (e.g. `/respond-pr` landing review fixes) refresh the mtime without explicit user re-authorization — so a stuck-but-not-cleaned-up active marker could bypass the gate indefinitely.
- **Fix (write side):** `scripts/activate ready-for-review` now writes a unix epoch timestamp instead of an empty file (`date +%s >`).
- **Fix (read side):** The hook reads `CREATED_TS` from the marker and requires `(now - created_ts) < 5400` (90 min). A `-ge 0` guard rejects future-dated markers (a negative delta would otherwise pass the `< 5400` check). Empty or non-numeric content fails closed, covering pre-upgrade markers.
- **90 min ceiling:** ~6x the observed typical run time (5-15 min); healthy skill runs are unaffected. At step 7 the marker is still removed by `/ready-for-review`, so the ceiling only matters for stuck/leaked markers.
- **Tests:** 5 new test cases; 2 updated existing tests (write valid timestamp so mtime-expiry failure is isolated from empty-content failure); 1 new `test_marker_script.py` test for epoch-content assertion on `activate ready-for-review`.

## Scope

Only `ready-for-review`. The same ceiling pattern applies to `require-respond-pr.sh`, `require-plan-review.sh`, `require-memory-skill.sh` — those hooks read mtime only, not content, so extending the pattern there would be a follow-up PR if/when warranted.

## Post-merge

`claude/.claude/plans` is a new top-level stow entry. Run `./install.sh` after merging so the symlink is created in `~/.claude/plans`.

## Test plan

- [ ] `pytest claude/.claude/` — 597 passed
- [ ] `ruff check claude/.claude/` — clean
- [ ] Manual smoke (positive path): run `/ready-for-review` on a branch with an open PR; `cat ~/.claude/.ready-for-review-active.d/<sid>` mid-run shows a unix epoch integer; iteration push is allowed; marker removed at step 7
- [ ] Manual smoke (hard ceiling): `echo "$(($(date +%s) - 5500))" > ~/.claude/.ready-for-review-active.d/<sid> && touch <path>`; attempt `git push` — hook denies
- [ ] Manual smoke (empty-content): `: > ~/.claude/.ready-for-review-active.d/<sid> && touch <path>`; attempt `git push` — hook denies

### DIFF
--- a/claude/.claude/hooks/require-ready-for-review.sh
+++ b/claude/.claude/hooks/require-ready-for-review.sh
@@ -13,10 +13,17 @@
 #
 # Two-marker pattern:
 # - Active marker (~/.claude/.ready-for-review-active.d/<session_id>):
-#   presence-only, fresh mtime <60 min. Written by /ready-for-review at
-#   step 0; removed at step 7 (completion). Bypasses the gate so the
+#   content = unix epoch written at activation; fresh mtime <60 min;
+#   activation age <90 min (hard ceiling). Written by /ready-for-review
+#   at step 0; removed at step 7 (completion). Bypasses the gate so the
 #   skill's own iteration pushes (step 3 fix → push → loop) don't
 #   self-deny. mtime refreshed on each bypass to handle long runs.
+#   The 90-min ceiling caps the sliding window: cross-skill iteration
+#   pushes (e.g. /respond-pr landing review fixes) refresh the mtime
+#   without explicit user re-authorization — without a ceiling, a
+#   stuck-but-not-cleaned-up active marker bypasses indefinitely.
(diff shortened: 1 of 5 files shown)


jcdendrite (commented):
lgtm

PR #174, opened 2026-05-07
fix(scripts): skip locked worktrees in cleanup-merged-branches.sh

jcdendrite:
## Summary

- `cleanup-merged-branches.sh` was calling `git worktree remove --force`, which overrides git's per-worktree lock. Active subagent worktrees carry a `locked claude agent ... (pid ...)` lock in porcelain output — so running cleanup while agents are still live removed their worktrees out from under them.
- New parser detects locked status using a deferred-commit pattern (the `locked` line appears *after* `branch` in each porcelain record, so we finalise at record boundaries rather than at the moment the branch line matches).
- When a worktree is locked, the entire branch is skipped — worktree removal, local-branch deletion, and remote cleanup all bypassed. Output: `worktree: locked (skipped)`.
- `--force` dropped from `git worktree remove` as defense-in-depth; the echo is now conditional on the exit status.
- `--dry-run` annotates locked candidates with `[locked — will skip]` to avoid surprising the user.

## Test plan

- [ ] `pytest claude/.claude/scripts/tests/test_cleanup_merged_branches.py -v` — 24/24 pass
- [ ] `TestLockedWorktreeSkipped` — locked worktree + branch survive; output contains `locked (skipped)`
- [ ] `TestLockedWorktreeSkipped::test_locked_annotated_in_dry_run` — dry-run shows `[locked — will skip]`, no destructive action
- [ ] `TestLockedWorktreeMixedWithUnlocked` — locked branch skipped, unlocked branch fully cleaned in same run
- [ ] Existing `TestBranchWithWorktree` still passes (unlocked path unchanged)

### DIFF
--- a/claude/.claude/scripts/cleanup-merged-branches.sh
+++ b/claude/.claude/scripts/cleanup-merged-branches.sh
@@ -142,7 +142,24 @@ if [ "$DRY_RUN" -eq 1 ]; then
   if [ "${#MERGED_BRANCHES[@]}" -gt 0 ]; then
     echo "Would clean up:"
     for BRANCH in "${MERGED_BRANCHES[@]}"; do
-      echo "  ${BRANCH} (${MERGED_PR_INFO[$BRANCH]})"
+      _DRY_LOCKED=0
+      _DRY_MATCHED=0
+      while IFS= read -r _line; do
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #176, opened 2026-05-07
refactor(handoff): inline format, drop /compact from docs

jcdendrite:
## Summary

- Smoke testing confirmed Anthropic's internal compaction engine uses its own 9-section format, ignoring CLAUDE.md instructions and external file references entirely — the §1–§6 shape has no effect on auto-compact output
- Only `/handoff` actually reads and follows the format definition (Claude manually reads it and writes the resume file)
- The marker re-injection hook restores gate state after auto-compact independently of the digest format — unaffected by this finding
- Since `COMPACT_INSTRUCTIONS.md` only served `/handoff`, the right fix is to inline the format there and delete the external file — not to rewrite it or inline it into CLAUDE.md

## Changes

- `claude/.claude/commands/handoff.md` — §1–§6 format inlined; file is now self-contained, no external pointer
- `claude/.claude/COMPACT_INSTRUCTIONS.md` — deleted (no remaining consumers after inline)
- `claude/.claude/CLAUDE.md` — remove dangling `## State digest format` pointer section; trim `/compact` citation-rationale from the 60%-threshold bullet (CLAUDE.md is loaded every session and should carry operational rules, not background reasoning — the Anthropic citation lives in the README)
- `README.md` — rewrite Context management section: drop all `/compact` recommendations, remove Compaction guidance bullet, clean decision table and threshold reference; Anthropic 60% citation kept but now scoped to `/handoff` with an honest analogical chain

## Test plan

- [ ] `cat claude/.claude/commands/handoff.md` — §1–§6 sections, pre-write checklist, and slug guidance all present inline; no external file reference
- [ ] `grep -r 'COMPACT_INSTRUCTIONS' .` — zero hits in tracked files
- [ ] `grep '/compact' README.md claude/.claude/CLAUDE.md claude/.claude/commands/handoff.md` — zero hits
- [ ] Read `README.md` Context management section — two layers (marker re-injection + /handoff), no `/compact` row in decision table, threshold reference scoped to `/handoff`
- [ ] Read `claude/.claude/CLAUDE.md` — `## State digest format` section gone; 60%-threshold bullet is operational-only with a README pointer for background


jcdendrite (commented):
lgtm

PR #183, opened 2026-05-08
feat(marker): guard against empty-staged write before git add

jcdendrite:
## Summary

- Adds `_guard_staged_vs_unstaged` helper to the marker script that exits 2 when the staged diff for the relevant pathspec is empty but unstaged tracked changes exist
- Prevents an empty-hash marker from being recorded before `git add` runs, which would cause a require-\* hook mismatch at commit time
- Calls the guard in both `code-review` (whole-tree) and `skill-review` (`claude/.claude/skills/**/SKILL.md` pathspec) write arms

## Test plan

- [x] 643 existing tests pass (verified locally)
- [x] `ruff check` clean (verified locally)
- [x] `TestMarkerScriptEmptyStagedGuard` covers 6 new cases:
  - code-review: staged + unstaged → marker written (guard does not fire)
  - code-review: empty staged + unstaged tracked → exits 2, no marker written
  - code-review: empty staged + no unstaged → marker written (review-of-nothing escape hatch)
  - skill-review: unstaged non-SKILL.md file → marker written (pathspec scoping works)
  - skill-review: unstaged SKILL.md → exits 2, no marker written
  - skill-review: staged SKILL.md → marker written

### DIFF
--- a/claude/.claude/hooks/tests/test_marker_script.py
+++ b/claude/.claude/hooks/tests/test_marker_script.py
@@ -131,3 +131,124 @@ class TestMarkerScriptHappyPath:
     def test_help_exits_0(self, isolated_home, git_repo):
         result = _run(["--help"], cwd=git_repo, home=isolated_home)
         assert result.returncode == 0
+
+
+class TestMarkerScriptEmptyStagedGuard:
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #184, opened 2026-05-08
feat(scripts): add ~/.local/bin wrappers for short-form script invocation

jcdendrite:
## Summary

- Ships three thin bash wrappers in `claude/.local/bin/` (stowed to `~/.local/bin/`): `cleanup-merged-branches`, `analyze-context`, `token-analyzer` — replacing the need to type full absolute paths like `~/.claude/scripts/cleanup-merged-branches.sh`
- Adds `mkdir -p "$HOME/.local/bin"` to `install.sh` before the stow invocation to prevent stow tree-folding on fresh-clone systems where the dir doesn't yet exist
- Commits exec bit on `analyze-context.py` (was `-rw-rw-r--`), removing the need for the `chmod +x ~/.claude/scripts/*` README setup step
- Adds exact-match `Bash(cleanup-merged-branches)` and `Bash(cleanup-merged-branches --dry-run)` allowlist entries in `settings.json`; keeps absolute-path entries during transition
- Updates README with per-platform PATH setup one-liners (Linux/WSL/macOS zsh/fish), rollout note with `--adopt` collision check, and short-form invocation examples
- Updates `CLAUDE.md`: `analyze-context.py` → `analyze-context` in the handoff threshold hint

The shape-enforcement hook pins the marker script to its absolute path — that script is intentionally excluded from the wrappers.

## Follow-ups

- [ ] Remove the absolute-path `cleanup-merged-branches.sh` allowlist entries once the short form is the default in practice (and the auto-memory reference has been live for a few weeks)
- [ ] Generalize `review-permissions` skill to flag bare-name `Bash(...)` allowlist entries as a category requiring justification

## Test plan

- [ ] After merging and re-running `./install.sh`, `command -v cleanup-merged-branches` prints `~/.local/bin/cleanup-merged-branches`
- [ ] `cleanup-merged-branches --dry-run` produces identical output to `~/.claude/scripts/cleanup-merged-branches.sh --dry-run`
- [ ] `git status` in repo after re-stowing shows no unexpected modifications under `claude/.local/bin/`
- [ ] `readlink ~/.local/bin` returns nothing (real directory, not a symlink — confirms no tree-folding)
- [ ] Old absolute-path invocation still works (backward compat)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #187, opened 2026-05-08
fix(hook): fix two false-positives in enforce-marker-script-shape

jcdendrite:
## Summary

- **Heredoc false-positive:** The activation guard used `grep -qF 'marker.sh'` — a bare substring match — so a `git commit` with a heredoc body mentioning `marker.sh` was inspected and denied. Fixed with a two-stage check: cheap substring fast-reject (Stage 1) followed by a bash `[[ =~ ]]` anchored leading-path check (Stage 2). Bash `=~` anchors at start-of-subject rather than per-line, so heredoc body lines don't activate the deep validator. Root cause behind PR #184's temp-file workaround.
- **Dot-range false-positive:** `grep -qF '..'` matched any two adjacent dots — `Bash(...)`, `git log a..b`, ellipses in commit messages. Replaced with a regex that matches `..` only as a path segment (preceded or followed by `/` or string boundary).
- **Traversal check ordering:** Moved before Stage 2 so tilde-form invocation paths containing a traversal segment are caught before Stage 2 can fast-exit them.
- **`$HOME` literal in Stage 2 alternation:** A literal `$HOME/.claude/scripts/marker.sh` invocation previously bypassed deep shape validation (fast-exit at Stage 2). It now reaches VALID_PATTERN and is correctly denied.
- **Header and deny message corrected:** Both previously claimed env-var prefix and bash wrapper forms were denied here — they're not. Those wrapper executables are not in `permissions.allow`, so the permission layer denies them before this hook's deep validation matters.

## Test plan

- [x] 653 tests passing (`pytest claude/.claude/ -q`)
- [ ] Manual: `git commit` with a heredoc body mentioning `marker.sh` — must succeed without temp-file workaround
- [ ] Manual: `gh pr create --body "marker.sh inline mention"` — must succeed
- [ ] Manual: `~/.claude/scripts/marker.sh bogus-verb skill` — must still deny
- [ ] Manual: confirm `bash -c "~/.claude/scripts/marker.sh activate code-review"` is denied at the permissions layer (not the hook)

## Security surface

Reviewed by CISO and staff-platform-engineer during planning. The narrowed activation guard creates a silent-allow path for wrapped/chained forms (`bash -c`, env-var prefix, semicolons, subshells) — previously caught by substring match and denied. Accepted: those wrapper executables are not in `permissions.allow`, so the permission layer denies them first. Bypass-shape tests pin this assumption with docstrings warning future editors to verify the permission-layer gate before widening.

## Follow-up

The traversal check still fires on any Bash command that (a) mentions `marker.sh` in its body AND (b) contains a traversal segment anywhere in the same command text. This PR's own body triggered it (temp-file workaround required here). Fixing this properly requires restructuring so the traversal check runs only after Stage 2 establishes it's an actual marker invocation, while still catching traversal-prefixed invocations that would otherwise slip past Stage 2 — filed as a separate follow-up.

### DIFF
--- a/claude/.claude/hooks/enforce-marker-script-shape.sh
+++ b/claude/.claude/hooks/enforce-marker-script-shape.sh
@@ -8,11 +8,13 @@
 # to fire this hook on ALL Bash commands. The internal grep is the actual
 # gate. The "if" field is a hint only.
 #
-# Any Bash command containing "marker.sh" must be one of the 12 valid
-# invocation shapes exactly. No chains, no env-var prefixes, no redirects,
-# no bash wrappers, no extra args. This prevents prompt-injection attacks
-# that chain marker.sh invocations with malicious commands and rely on the
-# settings.json allowlist to approve the first segment.
+# Commands that start directly with the marker.sh path (~/ or absolute) must
+# match one of the 12 valid invocation shapes exactly: no chains, no redirects,
+# no extra args. Wrapped forms (env-var prefix, bash wrapper, relative path,
+# subshell) are not gated here — they fast-exit at Stage 2 and are denied by
+# the permissions.allow layer, which does not list their wrapper executables.
+# Removing the permissions.allow gate without updating this hook would leave
+# those forms ungated.
 set -uo pipefail
 
 INPUT=$(cat)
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #190, opened 2026-05-08
feat(workflow): add foundation-critique gates to plan-review, plan-it, agents, and CLAUDE.md

jcdendrite:
## Summary

Adds workflow infrastructure so Claude can catch wrong-foundation designs itself, rather than relying on user pushback after elaboration has already happened.

Motivated by a case study where a session built PreToolUse validator + Write hook with symlink-race checks + mktemp anchoring + cross-platform fixtures + staggered two-PR rollout to harden a `bypassPermissions: true` foundation that was itself unnecessary. A ciso-reviewer pass surfaced 3 Criticals — all real, all only present because the bypass design existed.

Three structural gaps corrected:

- **plan-review Step 4**: adds a third gate question — a concrete tripwire block — before domain checklists. Three tripwires fire on observable plan text (over-powered primitive, compounding layers, self-referential findings), not on judgment calls, so they catch wrong-foundation designs even when the AI's internal reasoning is coherent. When any fires, specialists are not spawned until the foundation question is resolved.

- **All 8 specialist agents**: each now has a domain-adapted "foundation question first" instruction and a "Foundation concern (or N/A)" output slot above per-finding output. Previously all 8 were structurally defense-auditors on a fixed foundation — no slot existed for "this artifact/control/contract/test shouldn't exist."

- **plan-it Step 5**: requires (a) literal source-line quotes for external-doc pattern claims (extends the existing Step 3 grep rule to external docs), and (b) a "Lighter alternatives considered" subsection when the chosen approach is heavier than the task needs.

Two ambient heuristics added to CLAUDE.md ("Default-suspect over-powered primitives," "Compounding defensive layers are a wrong-foundation tell") for coverage in all sessions independent of /plan-review or /plan-it.

Surfacing incident documented in plan-review/REFERENCES.md (load-on-demand, not in the skill body).

## Test plan

- [ ] Run `pytest claude/.claude/` — Step 4 changes are prose-only; hook-alignment tests for plan-review fixtures should pass unchanged
- [ ] Smoke test 1: Write a stub plan in `.claude/plans/` that proposes an over-powered approach (e.g., `bypassPermissions` where the source doc has a non-bypass alternative). Run `/plan-review`. Verify output leads with "Foundation concern" and no specialists are spawned.
- [ ] Smoke test 2: Trigger `ciso-reviewer` directly on a diff with a deliberate over-privileged scaffold. Verify output begins with "Foundation concern (or N/A):" line.
- [ ] Smoke test 3: Ask `/plan-it` to plan something whose chosen approach is heavier than needed. Verify plan includes "Lighter alternatives considered" subsection.

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -7,12 +7,14 @@
 - Avoid duplicating managed values across files where they can drift out of sync, but use judgment — sometimes a simple hardcoded value is better than an over-engineered abstraction.
 - Before taking any action that is destructive, irreversible, or has blast radius beyond the immediate change (data loss, breaking API changes, infrastructure modifications), flag the risk and confirm the approach.
 - When uncertain about a CLI flag, tool behavior, or API detail, verify rather than guessing.
+- **Default-suspect over-powered primitives.** When a design adopts a more powerful, invasive, or wider-scope mechanism than the task requires — a heavier abstraction, a more privileged execution context, a more complex coordination pattern, a more invasive integration — identify the lighter primitive in the source documentation/system that could solve the problem before adopting the heavier one. Re-read the source with the specific question "what mechanisms exist that do NOT require this heavier choice?" — first-pass reads miss the answer routinely.
 
 ## Working Style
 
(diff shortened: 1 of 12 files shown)


jcdendrite (commented):
lgtm

PR #192, opened 2026-05-09
feat(hook): block escaped backticks in gh pr create/edit bodies

jcdendrite:
## Summary

- Adds `deny-escaped-backticks-in-pr-body.sh`, a PreToolUse gate on `gh pr create` and `gh pr edit` that rejects bodies containing literal `\`` sequences.
- Wires two new entries into `settings.json` (`Bash(gh pr create *)`, `Bash(gh pr edit *)`) adjacent to the existing PR-body gate group.
- 12 new tests; 665-test full suite green.

**Why a hook, not just the skill rule:** The "Backtick hygiene" prose already exists in `ready-for-review/SKILL.md` lines 133–140. Sessions that edit a PR body outside the ready-for-review flow (mid-iteration `gh pr edit`, respond-pr flows) never read that rule. The hook fires deterministically regardless of which skill is active.

**What it catches:** `\`` survivng from a `<<'EOF'` single-quoted heredoc where the session unnecessarily escaped the backtick. The fix named in the deny message: drop the backslash — no escape is needed inside a single-quoted heredoc.

**Scope limits (documented in hook header):** does not cover `git commit` or `gh api`; does not resolve `--body "$(cat file)"` shell-substitution; does not carve out `\`` inside fenced code blocks (the fix is always to drop the escape).

## Test plan

- [ ] `pytest claude/.claude/hooks/tests/test_deny_escaped_backticks_in_pr_body.py -v` — 12/12 pass
- [ ] `pytest claude/.claude/` — full suite green (665 tests)
- [ ] `ruff check claude/.claude/` — clean
- [ ] Live smoke: `gh pr create --body 'test \`backtick\`'` → hook denies; `gh pr create --body 'test \`backtick\`'` with backslash removed → allows

### DIFF
--- a/claude/.claude/hooks/deny-escaped-backticks-in-pr-body.sh
+++ b/claude/.claude/hooks/deny-escaped-backticks-in-pr-body.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+# Gate: reject `gh pr create` and `gh pr edit` commands whose body
+# content (inline --body "..." or body-source file) contains literal
+# backslash-backtick sequences (\`). Those sequences appear when a
+# session uses a <<'EOF' single-quoted heredoc but writes \` to
+# "escape" the backtick — the backslash survives into the body GitHub
+# renders, breaking code-span formatting.
+#
+# NOTE — `if`-dispatch in settings.json is advisory; the authoritative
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #195, opened 2026-05-09
feat(hooks): replace TTL-based marker liveness with PID liveness

jcdendrite:
## Summary

- **Root fix**: Active-bypass markers now store the originating Claude session PID; hooks check liveness with `kill -0` instead of a 60-minute mtime window. Dead PID → orphan evicted on next gate hit. Alive PID → bypass.
- **`--continue` edge case**: `capture-session-id.sh` sweeps `.*-active.d` entries for the current session on `SessionStart` and rewrites their PID body, so resumed sessions don't lose bypass on the new liveness check.
- **Recovery utility**: `marker.sh clear-stale [--dry-run]` walks all `.*-active.d` dirs and evicts dead-PID entries manually.
- **Authorial fix**: `plan-review` SKILL.md now calls `deactivate` unconditionally before the conditional `write`, covering Request-changes / error / user-cancel exits.
- **Removed**: The entire TTL compensation stack — 60-min mtime check, mtime-refresh on bypass, 90-min hard ceiling, future-timestamp guard, empty/non-numeric guards in `require-ready-for-review.sh`.

## What this dissolves

Five compounding layers existed because `marker.sh activate` had no termination signal. Each layer closed a gap created by the prior layer. Replacing the TTL guess with PID liveness detection makes all five layers redundant and removes them.

## Test plan

- [x] 668 tests pass (`pytest claude/.claude/`)
- [ ] Manual smoke: `marker.sh activate plan-review` → check active.d entry has numeric PID → kill that PID → trigger any gated Write → confirm orphan evicted and gate denies
- [ ] Manual smoke: `marker.sh clear-stale` with a live session → reports kept, not evicted
- [ ] Manual smoke: `claude --continue` resumes session → `capture-session-id.sh` rewrites PID in existing active.d entries → bypass still works


jcdendrite (commented):
lgtm

I always thought the TTL was an odd choice, but I okayed it - this is significantly better

PR #200, opened 2026-05-12
fix(hooks): close 7 command-shape gaps in ready-for-review gate

jcdendrite:
## Summary

- Replaces the regex-based push detector in `require-ready-for-review.sh` with the same tokenized fragment parser already used in `require-worktree-for-git-writes.sh`, shared via `_lib.sh`. Closes seven command forms the old regex missed silently.
- Reads `.cwd` from the tool-input JSON for all git/gh operations instead of relying on the inherited hook process CWD, matching the existing pattern in the worktree hook.
- Guards the new `_lib.sh` source in `require-worktree-for-git-writes.sh` so a missing `_lib.sh` emits a deny rather than exiting non-zero with no output.

**Command shapes now gated (all were silent-allow before):**

| Shape | Why it escaped the old regex |
|---|---|
| `git -C /wt push origin feature` | flags break adjacency between `git` and `push` |
| `git --git-dir=... --work-tree=... push` | same |
| `GIT_DIR=... git push` | env-var prefix has no leading separator |
| `eval git push` / `xargs git push` | wrapper lacks separator |
| `(cd /wt; git push)` | trailing `)` defeated the `(\s\|$)` anchor |
| `git push;` | `;` is not whitespace and not `$` |

## Test plan

- [x] `pytest claude/.claude/` — 676 tests, all passing
- [ ] New parametrized test `test_command_shapes_that_escaped_old_regex_are_denied` covers all 7 gap shapes
- [ ] New test `test_cwd_json_field_used_for_branch_resolution` verifies `.cwd` JSON field is used for branch detection when hook process CWD differs

### DIFF
--- a/claude/.claude/hooks/_lib.sh
+++ b/claude/.claude/hooks/_lib.sh
@@ -11,3 +11,67 @@
 _marker_lib_repo_hash() {
   printf '%s' "$1" | sha256sum | awk '{print $1}'
 }
+
+# Decide whether a shell fragment actually invokes `git`, not just mentions it
+# as a substring of a path or URL. Walks whitespace-separated words; returns
+# success iff any word equals `git` or ends in `/git`. Env-var prefixes
+# (GIT_DIR=... git ...), wrapper commands (eval, sudo, xargs), and `git` as a
+# non-first word are all handled by scanning every word, not just the first.
+#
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

PR #201, opened 2026-05-12
fix(skills): add pre-write destination check to ai-instruction-and-memory-files

jcdendrite:
## Summary

- **Gap closed:** When `require-memory-skill.sh` forced the `ai-instruction-and-memory-files` skill, the session only checked whether an existing memory entry needed updating — it didn't first ask whether the content belonged in CLAUDE.md, AGENTS.md, or a specific SKILL.md instead. The routing logic existed in §4/§5 tables but was reached only after the writer had already committed to a destination.

- **Fix:** Added `## Step 1 — Decide where this rule belongs before writing` immediately after Step 0 (gate activation) and before the architecture overview. Step 1 mirrors Step 0's procedural form — which the transcript confirmed runs reliably before any downstream decision. It's an imperative 3-step checklist: grep instruction files first, check if the rule is workflow-scoped to a skill, then confirm memory is the right type.

- **Workflow gap closed:** `ready-for-review` step 6 verified pushed state but was silent on what to do when the branch was ahead post-iteration (steps 2/3 produced fix commits). Updated step 6 to instruct: push if ahead, then re-verify. Updated `require-ready-for-review.sh`'s denial message (both branches) to name this specific scenario explicitly.

## What changed

| File | Change |
|---|---|
| `claude/.claude/skills/ai-instruction-and-memory-files/SKILL.md` | Added Step 1 pre-write destination check (+10 net lines; stays at 200) |
| `claude/.claude/skills/ready-for-review/SKILL.md` | Step 6 pushed-commits bullet: verify-only → verify-and-recover |
| `claude/.claude/hooks/require-ready-for-review.sh` | Both denial-message branches name the post-iteration-commit scenario |

Local-only (not in this PR): deleted `~/.claude/projects/.../memory/feedback_push_without_asking.md` and removed its MEMORY.md index line — the rule now lives in the workflow surfaces it governs.

## Test plan

- [x] `pytest claude/.claude/` — 676 tests, all pass (pre-commit confirmed)
- [ ] Skill-review: both SKILL.md changes clean
- [ ] claude-hook-review: denial-message string change clean, no structural changes
- [ ] Code-review: no issues found


jcdendrite (commented):
lgtm

PR #206, opened 2026-05-13
feat(agents): treat PreToolUse hook blocks the same as permission denials in check-runner

jcdendrite:
## Summary

- Extends the check-runner agent to cover PreToolUse hook blocks (exit 2) with the same stop-and-report behavior as the existing "Permission denials" guidance.
- Prevents the check-runner subagent from acting on recovery commands embedded in a hook's stderr — applies to any project whose PreToolUse hooks emit diagnostic prose alongside the block decision.
- Wording is project-generic; no private-project references.

## Test plan

- [ ] Read the resulting file end-to-end and confirm the new paragraph reads cleanly after the existing Permission denials paragraph.
- [ ] No automated test coverage in claude-config for agent prose.

## Context

Pairs with a private-project hook-message rewrite that updates the hook's own stderr to direct diagnosis rather than instruct a recovery command. The same dynamic — a hook's stderr containing recovery prose that a subagent could mistake for instruction — generalizes to any project with PreToolUse hooks that emit diagnostics; this PR addresses the agent-side half.


jcdendrite (commented):
lgtm

PR #207, opened 2026-05-13
perf(skills): move skill-review and claude-hook-review to plugins, read-docx-comments to command

jcdendrite:
## Summary

- **`skill-review` and `claude-hook-review`** moved from always-loaded stowed skills (`claude/.claude/skills/`) to project-scoped marketplace plugins (`plugins/skill-review/`, `plugins/claude-hook-review/`). Auto-enabled in claude-config sessions via `enabledPlugins` in `.claude/settings.json`; downstream projects that stow claude-config but don't author SKILL.md files or hooks no longer pay the description-budget cost.
- **`require-skill-review.sh` hook** moved into the plugin (`plugins/skill-review/hooks/`) and wired via `hooks.json`. Hook glob expanded from `claude/.claude/skills/**/SKILL.md` to also cover `plugins/*/skills/**/SKILL.md`; same expansion applied in `check-skill-length.sh` and `marker.sh write skill-review` to keep read-side and write-side hashes in sync.
- **`read-docx-comments`** converted from a stowed skill to a slash command at `claude/.claude/commands/read-docx-comments.md`. Run irregularly enough that always-loaded description budget is not justified.
- **`marketplace.json`** updated with both new plugin entries.
- **Tests**: `test_require_skill_review.py` redirected to the plugin hook path; four new tests added (plugin-path deny, plugin-path allow, mixed stowed+plugin stale-marker deny, `_lib.sh` byte-identity guard). `test_skills.py` extended to discover plugin skills with dedup guard.
- **`docs/skills.md`**: new "Project-scoped plugins" and "Tuning the skill-listing budget" sections.

## Budget impact

Before this PR, three always-loaded skill descriptions (`skill-review`, `claude-hook-review`, `read-docx-comments`) consumed description budget in every session regardless of whether the user was touching those surfaces. After this PR, only sessions in repos that install the plugins or invoke the command pay the cost.

## Test plan

- [x] `pytest claude/.claude/` — 682 tests pass
- [x] `ruff check claude/.claude/` — clean
- [x] `/code-review` run on staged state (marker written)
- [x] `/skill-review` run on staged SKILL.md changes (marker written)
- [x] `require-skill-review.sh` plugin hook: deny without marker, allow with correct marker, deny on stale marker after re-staging, deny when session-B tries to use session-A's marker, deny when plugin-path SKILL.md staged without marker, deny when stowed+plugin staged after stowed-only marker
- [x] `_lib.sh` byte-identity test guards against hash-recipe drift between plugin hook and marker.sh

[...]

PR #210, opened 2026-05-13
fix(install): wire claude-config marketplace registration for contributor onboarding

jcdendrite:
## Summary

- **`install.sh`**: fixes the marketplace-registration loop — was reading `.value.source.repo` from `extraKnownMarketplaces`, which is absent for directory-source entries (jq emitted `null` → `claude plugin marketplace add null` failed silently). Replaces the single-field jq query with a three-column query (name, source type, identifier) and branches on source type: `directory` sources use `$REPO_DIR` (the checkout path of `install.sh`); `github` sources preserve existing behavior.
- **`claude/.claude/settings.json`**: adds a `claude-config` entry to `extraKnownMarketplaces` with `source.source: "directory"`. Was missing entirely — no auto-registration path existed for new contributors. `path: ""` satisfies the schema's required field; install.sh overrides it with `$REPO_DIR` at registration time.
- **`CONTRIBUTING.md`**: adds one-line pointer in "Skill edits" noting that `/skill-review` is provided by the `skill-review@claude-config` plugin (auto-enabled by `./install.sh`), with a link to docs/skills.md for the marketplace setup details.

## Root cause

PR #207 moved `skill-review` and `claude-hook-review` to project-scoped plugins auto-enabled via the project-local `.claude/settings.json`. That resolution requires the `claude-config` marketplace to be registered at user scope first. Without this fix, `./install.sh` silently failed to register it, contributors had no `/skill-review` available, and the first SKILL.md commit was blocked by `require-skill-review.sh` with no obvious recovery path.

## Upgrade notes

If you previously ran `./install.sh` and the `claude-config` marketplace is absent or pointing at a stale path, re-run: `claude plugin marketplace remove claude-config && ./install.sh`.

## Verification note

Before merging: confirm `path: ""` in the stowed `extraKnownMarketplaces.claude-config` entry does not produce a session-start warning in non-claude-config repos. Exposure is scoped (the matching `enabledPlugins` is project-local to the claude-config checkout, so other projects won't trigger a plugin-load attempt), but a smoke test on a fresh stow is recommended.

## Test plan

- [ ] Clone repo to a path other than the primary checkout (e.g., `/tmp/claude-config-test`). Run `./install.sh`. Confirm stdout includes `→ adding claude-config (/tmp/claude-config-test)` and no `null` failures.
- [ ] `claude plugin marketplace list` shows `claude-config` registered at the correct path.
- [ ] Idempotency: re-run `./install.sh` → `✓ claude-config (already registered)`.
- [ ] Open a Claude Code session in the test checkout. `/skill-review` appears in `/skills` and the commit gate fires correctly on a SKILL.md change.
- [ ] Open a session in an unrelated repo (after stow). No session-start warning about `claude-config` marketplace.
- [ ] `pytest claude/.claude/` passes.
- [ ] `ruff check claude/.claude/` clean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ CI runs both on every PR and main push. Both must pass before the PR can be revi
 
 ## Skill edits
 
-When editing a `SKILL.md`, invoke the skill on its own diff before committing. A skill edit can violate the rules it enforces — the skill is the reviewer of its own changes. Run `/skill-review` via the Skill tool and check the diff against its output before staging the file.
+When editing a `SKILL.md`, invoke the skill on its own diff before committing. A skill edit can violate the rules it enforces — the skill is the reviewer of its own changes. Run `/skill-review` via the Skill tool and check the diff against its output before staging the file. (`/skill-review` is provided by the `skill-review@claude-config` plugin, auto-enabled after `./install.sh` — see [docs/skills.md — Project-scoped plugins](./docs/skills.md#project-scoped-plugins).)
 
 ## Contact
 
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -209,6 +209,12 @@
         "source": "github",
         "repo": "anthropics/claude-plugins-official"
       }
+    },
+    "claude-config": {
+      "source": {
+        "source": "directory",
+        "path": ""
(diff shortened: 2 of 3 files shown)


jcdendrite (commented):
lgtm

PR #208, opened 2026-05-13
fix(scripts): add reachability-based merge detection to cleanup-merged-branches

jcdendrite:
## Summary

- `gh pr list --head <branch>` misses merged branches whose commits reached `main` via a differently-named PR head (renamed branch, agent worktree prefix, closed-and-reshipped PR). This caused the local branch scan list to grow over time.
- Adds `git merge-base --is-ancestor` as a second detection signal. Tier A (gh-confirmed) deletes silently as before; Tier B (reachable-only) prompts for confirmation or auto-confirms with `--yes`.
- Non-TTY stdin without `--yes` skips Tier B with a warning — prevents hangs when invoked via piped/cron/Claude Code Bash tool.
- Locked worktrees now check the pid embedded in the lock reason before unlocking: live agent worktrees are skipped rather than torn down.

## Test plan

- [x] 37/37 tests pass (`pytest claude/.claude/scripts/tests/test_cleanup_merged_branches.py`)
- [x] `ruff check claude/.claude/scripts/` clean
- [ ] `cleanup-merged-branches --dry-run` on this repo shows current stale local branches (e.g. `decompose-hook-tests`, `worktree-skill-footprint-restructure`) under "Probable merges" section
- [ ] `cleanup-merged-branches --yes` cleans them without prompting
- [ ] `cleanup-merged-branches` (no flags, from a terminal) prompts per Tier B branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #211, opened 2026-05-13
feat(skills): add stripped-WHY-comments check to code-review

jcdendrite:
## Summary

- Agents applying the "default to no comments" rule during edits were stripping **all** comments rather than evaluating each against the WHY test, deleting comments that documented non-obvious constraints, subtle invariants, and bug workarounds (observed in PR #208).
- Adds item **12 "Stripped WHY comments"** to the Clarity section of the `code-review` checklist. It fires when a diff deletes comments that meet the WHY standard and explicitly names the misapplication: the system default applies to *adding* comments, not to bulk deletion during unrelated edits.
- Subsequent items (old 12–35) renumbered to 13–36; Item Ownership table and inline "see item N" reference updated to match.

## Why /code-review rather than CLAUDE.md

The "default to no comments" rule lives in the Claude Code system prompt (not in CLAUDE.md — confirmed). A CLAUDE.md override addition would be loaded into every session; the preservation check only matters at review time. Placing it in /code-review fires at exactly the right moment.

## Test plan

- [x] 694/694 tests pass (`pytest claude/.claude/ -x -q`)
- [x] `/skill-review` passed on the SKILL.md diff (no behavioral regressions, all items pass behavior test)
- [x] `/code-review` passed (no issues; ripple-effect triage: renumbering is copy-edit-only for existing rows, new row has no named persona to spawn)
- [ ] Verify next time an agent edits a comment-heavy file, it flags stripped WHY comments rather than silently allowing deletion

### DIFF
--- a/claude/.claude/skills/code-review/SKILL.md
+++ b/claude/.claude/skills/code-review/SKILL.md
@@ -74,67 +74,69 @@ Evaluate the code against each item. Only flag items where there is a concrete i
 
 11. **Misleading names** — Do function or variable names promise more or less than they deliver? A `validateUser` that only checks one field, an `allItems` that holds a filtered subset.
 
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #212, opened 2026-05-13
docs(readme): surface skill-review and claude-hook-review as installable plugins

jcdendrite:
## Summary

- The marketplace exposes three plugins (`lovable-knowledge`, `skill-review`, `claude-hook-review`) but the README "Current plugins" list named only `lovable-knowledge` — downstream users had no discoverable install path for the other two.
- Added `skill-review` and `claude-hook-review` bullets to the "Current plugins" list with their install commands.
- Pulled the one-time `claude plugin marketplace add` prerequisite out of the `lovable-knowledge` bullet into a shared intro sentence above the list — avoids repeating it three times now that there are three plugins.
- Updated the directory-tree comment at L102 to enumerate all three plugin names instead of `lovable-knowledge, etc.`.

## Test plan

- [ ] Verify the "Current plugins" block in the rendered README lists all three plugins with install commands.
- [ ] Verify `marketplace add` appears once (in the intro sentence) and not inside any bullet.
- [ ] Verify the directory-tree comment at `## What this installs` names all three plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Verify: `command -v cleanup-merged-branches` should print the wrapper path.
 
 ```
 claude/        # stow package — claude/.claude/ → ~/.claude/
-plugins/       # marketplace plugins (lovable-knowledge, etc.)
+plugins/       # marketplace plugins (lovable-knowledge, skill-review, claude-hook-review)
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #209, opened 2026-05-13
fix(plugins): remove duplicate hooks entry from skill-review plugin manifest

jcdendrite:
## Summary

- The `skill-review` plugin manifest explicitly declared `"hooks": "./hooks/hooks.json"`, but Claude Code auto-loads `hooks/hooks.json` from the standard path for any plugin that contains one — causing a duplicate-load failure that prevented the plugin from initializing entirely.
- Removed the redundant `manifest.hooks` entry from `plugins/skill-review/.claude-plugin/plugin.json`; the file continues to be loaded via standard auto-loading.
- Bumped `version` to `1.0.1`.

`claude-hook-review`'s manifest was already correct (no `hooks` key) and is unchanged.

## Test plan

- [ ] Run `/doctor` and confirm the `skill-review@claude-config` plugin error is gone.
- [ ] Confirm `/skill-review` appears in the loaded skills list (not in the deficit listing).
- [ ] Smoke-test the gate: stage a no-op edit to any `SKILL.md` and attempt a commit — the `require-skill-review.sh` PreToolUse hook should block with the `/skill-review` instruction.
- [ ] `pytest claude/.claude/` — 682 tests pass (verified before PR).
- [ ] `ruff check claude/.claude/` — clean (verified before PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #213, opened 2026-05-13
feat(hooks): add chain-detection hint to deny-private-project-refs deny messages

jcdendrite:
## Summary

- Adds `chain_split_hint_if_chained()` helper to `deny-private-project-refs.sh` that detects `&&`/`||` chains in the Bash command and appends a corrective hint to both deny messages (tracker-ID branch and private-projects branch)
- The hint suggests splitting the chain into two separate Bash calls so a blocklisted token in a `cd /home/<name>/...` setup prefix doesn't appear in the gated command text
- Hint uses `<name>` placeholder — the \"matched entry NOT named\" invariant is preserved and explicitly tested in the new chained-blocklist test
- Mirrors the existing `cwd_anchor_note_if_chained` pattern in `require-worktree-for-git-writes.sh`
- 4 new tests: chained+tracker hint, chained+blocklist hint (with invariant assertion), unchained no-hint, semicolon-in-message no-hint

## Background

During PR #209, the redaction gate fired when `cd /home/<name>/... && gh pr create ...` was used and the user's name in the path matched their `~/.claude/private-projects.md` blocklist. The correct fix is not to weaken the match-anywhere foundation but to detect the chain shape and guide the agent toward the split-call workaround.

## Design decisions

- `&&`/`||` only — `;` excluded because it appears too often in English prose (commit messages, PR descriptions) to be a reliable chain signal
- Detection is best-effort grep (not a shell parser) — `&&` inside a quoted string body is a known false positive, documented in the helper comment
- Deny conditions unchanged — this is purely a UX improvement to the deny message

## Test plan

- [x] 698 tests pass (`pytest claude/.claude/`)
- [x] Ruff clean (`ruff check claude/.claude/`)
- [ ] Smoke test: from a worktree, run `cd /home/<your-name>/... && gh pr create --base main --head <branch> --title "..." --body "..."` with a blocklisted token in the path — deny message should include the chain-split hint
- [ ] Negative: `gh pr create` with a blocklisted token in the body and no chain operator — deny message should NOT include the hint

### DIFF
--- a/claude/.claude/hooks/deny-private-project-refs.sh
+++ b/claude/.claude/hooks/deny-private-project-refs.sh
@@ -108,6 +108,19 @@
 # Allowlist extension: append to OSS_ALLOWLIST below if a legitimate
 # open-source prefix is blocked. Do NOT add private-project-specific
 # prefixes.
+#
+# Shape-aware deny hint.
+# ----------------------
+# When the command chains operations with && or ||, the deny messages
+# (both tracker-ID and private-projects branches) append a hint suggesting
+# the agent split the chain into separate Bash calls. Detection is
+# best-effort grep (not a real shell parser) — && and || in prose are
+# rare enough to justify detection; ; is excluded to avoid false positives
+# on prose semicolons. The hint is informational only — the deny condition
+# is unchanged. The "matched entry is intentionally NOT named" invariant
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #215, opened 2026-05-14
feat(skills): lovable-cloud plugin, ready-for-review PR-creation, test-conventions discovery

jcdendrite:
## Summary

- A1: `ready-for-review` gains step 5.5 — opens a PR when none exists, using the `<TICKET-ID>/<topic-slug>` branch-creation convention. Previously deferred to a project-specific pre-merge skill; the global skill now handles it directly.
- A2: `test-conventions` adopts the Glob+load pattern for project-specific layers (matching `code-review` and `plan-review`), replacing the DO NOT TRIGGER suppression approach.
- A3: `lovable-knowledge` plugin renamed to `lovable-cloud` (v2.0.0) and joined by two new genericized skills: `lovable-cloud-edge-functions` (ES256 gateway constraint, four-tier auth model, `config.toml` rules) and `lovable-cloud-migration-sync` (identify unsynced migrations, diff duplicates, db reset, delete originals, open PR).

### Amendments (review feedback)

- C1: `lovable-cloud-edge-functions` — ES256/HS256 background section extracted to co-located `REFERENCES.md`; SKILL.md now carries a one-line pointer. Moves reference material out of the runtime-loaded body per `REFERENCES.md` convention.
- C2: `test-conventions` Step 0 rewritten to match the canonical Glob+load wording used by `plan-review` and `code-review` (adds "invoke it now and apply its conventions alongside the items below", "resolved via `git rev-parse --show-toplevel`", exact error/proceed phrasing).
- C3: `skill-review` SKILL.md and README updated with `disable-model-invocation: true` guidance for globbed project-specific add-on skills (`.claude/skills/<parent>-<project>/SKILL.md`). Cites `test-conventions`, `plan-review`, and `code-review` as parent skills using the pattern.
- C4: `skill-review` SKILL.md — removed stale `skill-creator` references from the DO NOT TRIGGER clause and body intro. `skill-creator` is not in this repo's `enabledPlugins`; REFERENCES.md documents why.
- C5: `skill-review` TRIGGER expanded to include authoring turns — `skill-creator` is disabled (`disable-model-invocation: true`), so the dual-fire concern no longer applies; `skill-review` is now the active skill for both reviewing and authoring `.claude/skills/**/SKILL.md` changes.

### Post-review fixes (CI + review-pass findings)

[...]

PR #217, opened 2026-05-14
docs(skills): document disable-model-invocation for project-layer add-on skills

jcdendrite:
## Summary

Adds the `disable-model-invocation: true` rule to the **Frontmatter** bullet in `docs/skills.md` → "Project-specific layers".

Project-layer add-on skills (`.claude/skills/<parent>-<project>/SKILL.md`) are invoked by the parent via Glob+Skill-tool, not via description-based auto-trigger — so their description does not need to appear in the always-loaded skill-listing budget. Setting `disable-model-invocation: true` removes the listing entry while leaving Skill-tool invocation intact.

One additional note lands alongside the rule: suppression is only observable from a fresh CLI session — subagents inherit the parent session's listing, so an O1-style listing check via a subagent will always show the entry even with the flag set.

## Why here, not `code-review/REFERENCES.md`

The empirical test for this rule was originally drafted as an F-07 entry in `code-review/REFERENCES.md`. That file is per-skill scoped ("reference material that informed *this* skill"); the finding applies to any parent skill using the Glob+Skill-tool dispatch pattern (`code-review`, `plan-review`, and `test-conventions` once PR #215 lands). Moving it to `docs/skills.md` "Project-specific layers" — the existing home for add-on skill guidance — keeps scope accurate and removes stale PR references that would not survive the PR description being lost.

## Test plan

- [ ] `git diff main` shows one-line edit to `docs/skills.md`, zero changes to `code-review/REFERENCES.md`
- [ ] `pytest claude/.claude/` passes (no test coverage for doc text; hook tests unaffected)
- [ ] `ruff check claude/.claude/` clean


jcdendrite (commented):
lgtm

PR #216, opened 2026-05-14
feat(respond-pr): avoid SHA-pinned commit references in replies

jcdendrite:
## Summary

One targeted skill addition for `respond-pr`, keeping acknowledgment text stable across rebases.

**respond-pr:**
- Avoid SHA-pinned commit references; prefer branch-tip language ("addressed in the latest commit on this branch") so acknowledgment text survives rebases and force-pushes without going stale. Also clarifies that stale SHAs from prior replies should be corrected with a new reply before requesting reviewer re-verification.

## What changed from the original draft

The original PR included four `plan-it` additions (design-doc grep, UI-touch scope check, billing-domain grounding note, domain-triggered mid-plan-review). These were reverted: all four encoded stack-specific paths or private-project domain names from one project's review, which is inappropriate for a globally stow-distributed skill. Those rules belong in a project-scoped skill in the consuming project's repo.

## Test plan

- [x] `pytest claude/.claude/` — 698 tests passed
- [x] `ruff check claude/.claude/` — clean
- [x] `/skill-review` on modified SKILL.md — clean
- [x] `/code-review` on staged diff — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #218, opened 2026-05-14
feat(skills): add hygiene items 9a/9b/34a to code-review checklist

jcdendrite:
## Summary

- **Item 9a** — flags repeated string literal domain discriminants (tier, role, status, event type) at ≥3 sites without a shared named type; suggests the language's idiomatic named-type construct (TypeScript string-literal union, Java/Kotlin/C# enum, Python `Literal[...]`/`Enum`, Rust enum, Go typed-string constant set) at the canonical module.
- **Item 9b** — flags bare numeric literals used as semantic bounds (prefix length for redaction, maximum field width, ID segment length) in slicing, truncating, or windowing operations in non-test code; requires a named constant in the language's idiomatic form.
- **Item 7 extension** — extends the unnecessary-wrapper rule to module level: a file whose entire post-refactor content is re-exports from one other module is a shim; suggests deleting it and updating importers directly.
- **Item 34a** — flags untyped/generic errors raised in files that already dispatch on typed error variants (catch-arm narrowing, `instanceof` / `isinstance` / `errors.As`, sum-type pattern match); the generic raise bypasses the typed arm and falls to the generic handler; fix is a typed variant, not widening the dispatch.
- **Ownership table** — adds rows for 9a, 9b, and 34a; 34a is co-owned by `staff-backend-engineer` (server-side paths) and `staff-frontend-engineer` (client-side paths), consistent with item 2's split.

Closes gaps 2, 3, 5, and 7 from the PR #201 gap-analysis plan (`claude/.claude/plans/gap-analysis-pr201-skill-updates.md`). The respond-pr + plan-it items from the same gap analysis landed in #216.

## Test plan

- [x] `pytest claude/.claude/` — 698 tests passed (prior session; SKILL.md-only diff, no test impact)
- [x] `ruff check claude/.claude/` — clean (prior session)
- [x] `/skill-review` on `code-review/SKILL.md` — clean (both sessions)
- [x] `/code-review` on cumulative diff — prior session fixed 34a ownership row missing `staff-frontend-engineer` co-owner; this session fixed stale JS-specific titles in 9b and 34a ownership table rows after paradigm-agnostic rewrites
- [x] Reviewer feedback addressed: 9a, 9b, 34a rewritten to be paradigm/language-agnostic


claude/.claude/skills/code-review/SKILL.md
```diff
@@ -62,12 +62,16 @@ Evaluate the code against each item. Only flag items where there is a concrete i
...
 
+9a. **Repeated domain discriminants without a shared type** — Is the same string literal representing a domain concept (tier, role, status, event type) used as a type annotation at 3 or more sites without a shared `type` alias? Flag with: suggest `export type <Name> = '<value1>' | '<value2>'` at the canonical module for that domain, and replace inline literals at call sites. String literal unions are preferred over `enum`; do not suggest `enum` as the fix.
```
jcdendrite: Are you sure this is canonical across all languages: "string literal unions are preferred over 'enum'"? That doesn't work for all languages (hint, hint Java).

claude/.claude/skills/code-review/SKILL.md
```diff
@@ -62,12 +62,16 @@ Evaluate the code against each item. Only flag items where there is a concrete i
...
+
+9b. **Unnamed bounds in slice/substring calls** — Does the code pass a numeric literal directly as a `substring`, `slice`, `substr`, or `truncate` argument in non-test code? If the number represents a semantic bound (prefix length for redaction, maximum field width, ID segment length), it should be a named constant expressing the intent, not a bare number. Suggest: `const <DESCRIPTIVE_NAME> = <N>;` at the shared-utility level if the bound is used across files, or at the top of the file if local.
```
jcdendrite: I don't think you need to type out all types here `substring`, `slice`, etc. - those seem language-specific. Be more general here. Remember this is the generic code review file that applies to all projects.

claude/.claude/skills/code-review/SKILL.md
```diff
@@ -132,6 +136,8 @@ Evaluate the code against each item. Only flag items where there is a concrete i
 
 34. **Performance-sensitive code paths** — Does the change modify a hot path (queries in loops, N+1, cache read/write, large list ops)? Verify with representative data volumes, not just test fixtures.
 
+34a. **Bare `new Error()` in a file with typed Error subclasses** — If the file defines or imports a custom `Error` subclass and uses `instanceof` dispatch in a catch handler, any `throw new Error(string)` for a parallel failure path will fall through the typed catch arm and land in the generic handler. Flag any bare `new Error(string)` throw in the same file that shares the logical shape of an existing typed subclass. The fix is a typed subclass — not try/catch widening.
```
jcdendrite: I feel like this is too specific. Some paradigms/languages don't use error throwing. Can you make this more paradigm/language-agnostic without losing actionable specificity?

jcdendrite (commented):
Does the skill length still pass the required check?

claude/.claude/skills/code-review/SKILL.md
jcdendrite: **[Claude Code]** Rewrote 9a to drop the TS-specific prescription. The new wording suggests the language's idiomatic named-type construct — TS string-literal union, Java/Kotlin/C# enum, Python `Literal[...]`/`Enum`, Rust enum, Go typed-string constant set — declared at the canonical module. `enum` is now explicitly listed as the right answer for Java/Kotlin/C#/Rust rather than being prohibited. Addressed in the latest commit on this branch.

claude/.claude/skills/code-review/SKILL.md
jcdendrite: **[Claude Code]** Rewrote 9b to drop the `substring`/`slice`/`substr`/`truncate` enumeration and the JS `const` declaration template. The rule now describes the behavior shape ("slicing, truncating, or windowing a string or collection") and recommends a named constant in the language's idiomatic form. Addressed in the latest commit on this branch.

claude/.claude/skills/code-review/SKILL.md
jcdendrite: **[Claude Code]** Rewrote 34a to generalize vocabulary across paradigms: "typed error variants" covers exception classes, Rust enum variants, Go typed-error values, etc.; dispatch-by-type covers `instanceof` / `isinstance` / `errors.As` / sum-type pattern match; "generic-error raise, throw, or error-return" covers languages that return errors rather than throwing. Substance preserved: typed variant, not widening the dispatch. Addressed in the latest commit on this branch.

jcdendrite:
**[Claude Code]** Length check: the file is at 291 lines (main baseline 283, +8 net for the four new items). The skill-review target is under 200 lines, with 200–300 flagged and past 300 a harder threshold — so this file was already past target on main; this PR adds to it but stays under the harder bound. The paradigm-agnostic rewrites are roughly line-flat, so they do not make the length worse. A separate compression pass on the whole skill is the right lever to bring it back under 200 and is out of scope for this PR.

jcdendrite (commented):
lgtm

PR #219, opened 2026-05-14
feat(skills): mark command-style skills as disable-model-invocation

jcdendrite:
## Summary

- `plan-it`: replaces `user-invocable: true` (default, was always model-invocable) with `disable-model-invocation: true`; updates description to user-facing summary since TRIGGER routing text is dead weight when the description is no longer loaded into Claude's context
- `commands/handoff.md`: adds frontmatter with `disable-model-invocation: true` (was missing any frontmatter, making it model-invocable)
- `commands/read-docx-comments.md`: adds `disable-model-invocation: true` to existing frontmatter

## Why

Per the Claude Code skills docs, skills without `disable-model-invocation: true` have their descriptions loaded into every session so Claude can auto-invoke them. These three are user-controlled workflows — Claude should not decide to run them; the user must type `/skill-name` explicitly.

### DIFF
--- a/claude/.claude/commands/handoff.md
+++ b/claude/.claude/commands/handoff.md
@@ -1,3 +1,8 @@
+---
+description: Write a cross-session handoff file at /tmp/<descriptive-slug>-handoff.md.
+disable-model-invocation: true
+---
+
 Write a cross-session handoff file at `/tmp/<descriptive-slug>-handoff.md` using the structure below.
 
 ## §1 Goal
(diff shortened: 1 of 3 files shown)


jcdendrite:
Closing — PR 221 (#221) landed first and already ships both of the commands→skills migrations (handoff and read-docx-comments) with `disable-model-invocation: true`. The two path edits here are dead (the files moved). The plan-it hunk is not stale — it's a deliberate disagreement with PR 221's choice to keep plan-it model-invocable (with TRIGGER discipline, enforced by the new COMMAND_SKILLS test registry). That position deserves a fresh focused PR against current main if it's still worth pursuing.

PR #221, opened 2026-05-14
feat(skills): migrate commands to skills, add TRIGGER discipline and test invariants

jcdendrite:
## Summary

- **Migrate `commands/` → `skills/` with `disable-model-invocation: true`**: `handoff` and `read-docx-comments` were in `commands/` without frontmatter, causing their descriptions to appear in the model's always-loaded skill listing despite being purely user-typed slash commands. Moved to `skills/<name>/SKILL.md` with the flag, which suppresses the listing-budget entry while preserving `/handoff` and `/read-docx-comments` slash invocation (empirically confirmed safe by F-07).
- **Add TRIGGER/DO NOT TRIGGER discipline to `code-review` and `respond-pr`**: These were the only model-invokable workflow skills lacking the standard trigger-discipline shape. `respond-pr`'s existing body TRIGGER text is lifted verbatim; `code-review` adds adjacent-skill callouts (`skill-review`, `plan-review`). Also standardizes colon syntax on `plan-it` and `ready-for-review` (`TRIGGER when:` / `DO NOT TRIGGER when:` matching what the new tests check for).
- **Add `/ready-for-review` to `claude/.claude/CLAUDE.md`**: The pre-push gate lacked the always-loaded CLAUDE.md anchor that `/code-review` and `/plan-review` already have, making it the most fragile discovery path in the workflow skill set.
- **Extend `test_skills.py` with two new invariants**: (a) `_model_invokable_skills()` + tests enforce TRIGGER discipline on every model-invokable skill without per-skill registration; (b) `COMMAND_SKILLS` registry enforces `disable-model-invocation: true` on user-only commands. Docstring calls out the three invocation paths and explains why `user-invocable: false` + `disable-model-invocation: true` is a valid pattern (parent-glob-dispatch add-on layers — not a bug).
- **Update `docs/skills.md` and `README.md`** for new paths and the accurate suppression-mechanism explanation (`disable-model-invocation: true` vs the former "not a skill" framing).

## Test plan

- [x] `pytest claude/.claude/` — 736 tests pass
- [x] `/skill-review` on all changed `SKILL.md` files — passed; one finding fixed (code-review DO NOT TRIGGER adjacent-skill callouts)
- [x] `/code-review` on full diff — clean
- [x] **Owner-driven manual smoke test**: In a fresh CLI session on this branch, verify (a) `/handoff` and `/read-docx-comments` still autocomplete and execute correctly; (b) the model's available-skills system-reminder no longer lists descriptions for `handoff` or `read-docx-comments` (F-07 O1 — only observable from a human-driven fresh session); (c) run `/doctor` to confirm listing budget pressure dropped.


jcdendrite (commented):
lgtm

PR #223, opened 2026-05-14
feat: disable update-config built-in, promote two memories to durable surfaces

jcdendrite:
## Summary

- Disable the `update-config` built-in skill via `skillOverrides: "off"` — its ~1,100-char description was the largest single contributor to the skill-listing budget overflow that caused `git-state-safety` to appear with no description in the available-skills system reminder
- `update-config`'s high-value territory is already covered by focused skills: `review-permissions` (permissions.allow), `claude-hook-review` (hooks), `skill-review` (skill bodies); remaining env/model/theme edits are trivial direct file changes
- Update `docs/skills.md`: bump disabled-skill count to eight, add rationale row for `update-config`
- Promote permission-grammar + three-layer strict-shape pattern from memory into `review-permissions/REFERENCES.md` (full pattern) + a one-line gloss in `review-permissions/SKILL.md` (cross-reference)
- Promote `settings.json` vs `settings.local.json` scoping rule from memory into `claude/.claude/CLAUDE.md` Safety section (applies to all projects, not just claude-config)
- Deleted the two promoted memory files + removed their MEMORY.md index lines (memory is for personal preferences, not load-bearing engineering rules per existing `feedback_memory_vs_claude_md_classification` guidance)

## Test plan

- [ ] JSON sanity: `python3 -c "import json; json.load(open('claude/.claude/settings.json'))"`
- [ ] Docs row count: `grep -c '^| /' docs/skills.md` → 8
- [ ] Promoted content: `grep -n "three-layer strict-shape" claude/.claude/skills/review-permissions/SKILL.md claude/.claude/skills/review-permissions/REFERENCES.md` → hits in both
- [ ] Settings split rule: `grep -n "settings.local.json" claude/.claude/CLAUDE.md` → at least one hit
- [ ] Fresh session: `update-config` absent from available-skills list; `loop` and `schedule` still present
- [ ] Fresh session: `git-state-safety` shows its full TRIGGER / DO NOT TRIGGER description (the truncation symptom this PR fixes)
- [ ] `/doctor` quiet on skill-listing budget in fresh session

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #224, opened 2026-05-14
fix(review-permissions): remove stale update-config reference

jcdendrite:
`update-config` was disabled in #223; the `(use `update-config`)` parenthetical in the DO NOT TRIGGER description pointed to a skill that no longer fires.

One-word removal — no behavioral change to the skill's logic.


jcdendrite (commented):
lgtm

PR #227, opened 2026-05-14
fix(skills): project layers load via Read+glob, not Skill()

jcdendrite:
## Summary

- `disable-model-invocation: true` blocks `Skill()` invocation in current Claude Code; two sessions hit `"Skill <name> cannot be used with Skill tool due to disable-model-invocation"` when parent skills tried to load a project layer
- Switches all four parent skills (`code-review`, `plan-it`, `plan-review`, `test-conventions`) from `Skill()` dispatch to `Read`+glob — the right primitive since the parent needs the layer's _content_, not an invocation context
- Updates `docs/skills.md`, `docs/design-decisions.md` §8, and `tests/test_skills.py` to reflect the corrected mechanism; adds 8 parametrized regression assertions

## Test plan

- [x] `pytest claude/.claude/` — 750 tests pass (includes 8 new `TestProjectLayerUsesReadNotSkill` assertions)
- [x] `/skill-review` on all four SKILL.md diffs — behavioral-equivalence table all Y, no blockers
- [ ] Manual smoke: run `/code-review` or `/plan-review` in a project that has a `code-review-*` or `plan-review-*` layer; confirm parent globs and Reads the layer without error

**Follow-up (not blocking):** `plugins/skill-review/skills/skill-review/SKILL.md` §1 still documents the old "Glob+Skill-tool" pattern for `disable-model-invocation: true` add-on skills — will need a separate PR.


jcdendrite (commented):
lgtm

PR #233, opened 2026-05-15
feat(check-runner): hard guardrails against rogue investigation runs

jcdendrite:
## Summary

- Drop `Write` from `tools`; spool files now use Bash redirect to `${TMPDIR:-/tmp}/` — no other write path is legitimate for check-runner
- Add `maxTurns: 20` (~2× typical dispatch depth) to hard-stop runaway iteration via the canonical Anthropic primitive
- Add reviewer-style "You do not modify project files" prose, single-shot per-command protocol, and require `timeout: 600000` on every Bash call
- Hoist the 32-entry read-only git subcommand allowlist into `_lib.sh` (`_lib_readonly_git_subcmds`); both `require-worktree-for-git-writes.sh` and `check-runner-bash-guard.sh` source from it — eliminates the duplicated-literal anti-pattern
- Add `check-runner-bash-guard.sh` as a correct reference implementation for the git-write guard; agent-frontmatter hooks do not fire when a subagent is spawned via the Agent tool (verified pre-merge — see "What I did NOT do")
- Address review: dedupe §10 paragraph (two near-duplicate drafts collapsed into one), rename "Reviewer subagents" section to "Agents" and split into reviewer-persona + check-runner blocks
- Restore `claude-config` marketplace entry in `settings.json` accidentally dropped when #232 collapsed the duplicate-permissions block

## Background

check-runner went off-script: 89 min / 121 tool calls, created an unauthorized SQL migration, and modified production source in a downstream project while attempting to "fix" a failing test. Three gaps enabled it: `tools: Bash, Write` with no path scoping; prose that forbade recommending fixes but not applying them; no turn cap.

Non-git mutation vectors (`rm`/`mv`/stray redirects) are not separately fenced — `maxTurns` + dropped `Write` + prose are sufficient. Stacking more pattern-regex would compound defensive layers without a foundational reason.

Incident and rationale documented in `docs/design-decisions.md` §10.

## What I did NOT do (and why)

- Custom marker-file tool-call counter → `maxTurns` is the canonical primitive
- Wire `check-runner-bash-guard.sh` via agent frontmatter → pre-merge testing confirmed that `hooks:` in agent YAML frontmatter does not fire when a subagent is spawned via the Agent tool (tested with bare tilde path and absolute path; both silently no-op). Settings.json wiring would fire globally for all sessions, defeating the check-runner-only scope intent. The script is kept as a reference; wire via settings.json only after verifying the global-deny posture is compatible with normal git operations.
- Regex for `rm`/`mv`/redirects → compounding layers over a solved foundation
- Create `docs/agents.md` → reviewer conditioned this on a third agent type being added

## Test plan

- [x] 801 existing tests pass (`pytest claude/.claude/ -x -q`)
- [x] 53 new guard-hook tests pass (`pytest claude/.claude/hooks/tests/test_check_runner_bash_guard.py -v`)
- [x] Ruff clean (`ruff check claude/.claude/`)
- [x] `grep 'ALLOWED_SUBCMDS=(blame' claude/.claude/hooks/*.sh` — no matches (allowlist deduped to `_lib.sh`)
- [x] `bash -c '. claude/.claude/hooks/_lib.sh && _lib_readonly_git_subcmds | wc -l'` — 32
- [x] Agent-frontmatter hook wiring verified non-functional pre-merge; frontmatter `hooks:` block removed; `check-runner-bash-guard.sh` header updated to document this
- [x] Dispatch check-runner on real project — verdict returned in ≤ 6 turns with no deny in transcript (verified during ready-for-review gate: `pytest claude/.claude/` + `ruff check claude/.claude/`)


docs/design-decisions.md
```diff
@@ -49,3 +49,15 @@ Three operations on the persona roster, with different decision criteria. **Bias
...
+
+Non-git mutation vectors (`rm`/`mv`/stray redirects) are not separately fenced — `maxTurns` plus dropped `Write` plus prose are sufficient. Git mutations get a dedicated harness block because commits land in history and pushes can't be undone; `require-worktree-for-git-writes.sh` in `settings.json` covers worktree-required repos for all sessions including check-runner.
+
+Non-git mutation vectors (`rm`/`mv`/stray redirects) are not separately fenced — `maxTurns` plus dropped Write plus prose are sufficient. Stacking pattern-regex for those would compound defensive layers without a foundational reason: every such layer exists because a prior layer left a gap, which is the tell that the foundation needs changing, not more hardening. Git mutations get a dedicated harness block because commits land in history and pushes can't be undone.
```
jcdendrite: What happened here? These are almost redundant word for word and look like different versions of drafts. Which paragraph version are you keeping?

README.md
```diff
@@ -174,7 +174,7 @@ One-time setup per machine: register the marketplace with `claude plugin marketp
...
 ### Reviewer subagents
 
-Eight stack-agnostic reviewer personas in `claude/.claude/agents/`, spawned by `/plan-review` and `/code-review` based on the **Item ownership** tables in those skills. Each runs in its own context with read-only tools (`Read`, `Grep`, `Glob`, `Bash`). The directory also ships `check-runner`, a non-reviewer Haiku agent that runs test suites and returns structured pass/fail verdicts — dispatched by the parent via the `Agent` tool; see `claude/.claude/CLAUDE.md` "Heavy command output".
+Eight stack-agnostic reviewer personas in `claude/.claude/agents/`, spawned by `/plan-review` and `/code-review` based on the **Item ownership** tables in those skills. Each runs in its own context with read-only tools (`Read`, `Grep`, `Glob`, `Bash`). The directory also ships `check-runner`, a non-reviewer Haiku agent that runs test suites and returns structured pass/fail verdicts — dispatched by the parent via the `Agent` tool; see `claude/.claude/CLAUDE.md` "Heavy command output". `check-runner` is constrained to `tools: Bash` only, capped at `maxTurns: 20`, and guarded by a `hooks.PreToolUse` script that denies git write operations — see `docs/design-decisions.md` §10.
```
jcdendrite: `check-runner` is not a reviewer subagent. I recommend having a section on agents. If we add another type of agent beyond reviewer and check-runner we can handle this like hooks and skills and surface the interesting ones in the README but put full descriptions in a dedicated doc.

docs/design-decisions.md
jcdendrite: **[Claude Code]** Fixed. The two paragraphs were near-duplicate drafts that both survived the prior edit — each carried one unique claim (the compounding-layers rationale and the `require-worktree-for-git-writes.sh` session-coverage note). Collapsed into a single paragraph that preserves both. `disposition: fixed-as-requested` | `commit_sha: c9b0448`

README.md
jcdendrite: **[Claude Code]** Fixed. Renamed `### Reviewer subagents` → `### Agents`; added a one-line intro distinguishing the two agent types; split the body into two clearly-labelled blocks — **Reviewer subagents** (the 8-persona list, routing notes, §9 link) and **`check-runner`** (its own paragraph with dispatch reference and §10 constraint rationale). `docs/agents.md` intentionally not created — your framing conditions it on a third agent type appearing. `disposition: fixed-as-requested` | `commit_sha: c9b0448`

jcdendrite (commented):
lgtm

PR #237, opened 2026-05-15
feat(plan-review): add misordered observe-then-mutate tripwire to Step 4

jcdendrite:
## Summary

- Adds a fourth foundation-correctness tripwire to `plan-review` Step 4: when a plan caveats a step's output as possibly-stale because a later step changes the state that output reads, the steps are misordered — the fix is to re-sequence, not to annotate.
- Folds the first two over-elaboration marker bullets into one (all four clauses survive verbatim) to hold the file at 239 lines and pass the `check-skill-length` gate.
- Appends a one-line origin note to `plan-review/REFERENCES.md` so future editors applying the behavior test have the rationale on file.

## Why

During planning of the PR 235 CI-status fix, the planning session repeatedly proposed a compensating caveat ("this result may be stale, re-check it") instead of re-sequencing the steps. Step 4's three existing tripwires (over-powered primitive, compounding layers, self-referential findings) are all calibrated for too-much-machinery — a single misplaced step plus a single caveat trips none of them. This is the orthogonal bug class they were blind to.

## Test plan

- [x] `/skill-review` on staged diff — behavioral-equivalence audit confirmed all folded clauses survive verbatim; marker written
- [x] `/code-review` on staged diff — clean; marker written
- [x] `pytest claude/.claude/` — 801 tests pass (hook-alignment tests confirm HOOK_TEST_FIXTURE blocks in `plan-review/SKILL.md` are byte-identical)
- [x] File is 239 lines (same as HEAD) — `check-skill-length` gate passes

### DIFF
--- a/claude/.claude/skills/plan-review/REFERENCES.md
+++ b/claude/.claude/skills/plan-review/REFERENCES.md
@@ -31,6 +31,8 @@ Step 4's foundation tripwires (over-powered primitive, compounding layers, self-
 
 The lesson the rules encode: gap-finding on a wrong foundation elaborates the wrong foundation. The tripwires anchor Step 4 to observable surface features (layer counts, self-references, heavier-than-needed mechanism) so they fire even when the AI's internal reasoning is coherent. See also `claude/.claude/CLAUDE.md` Engineering Judgment and Working Style for the global-level statement of the same heuristics.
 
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #238, opened 2026-05-15
fix(install): remove empty-path directory marketplace from extraKnownMarketplaces

jcdendrite:
## Summary

- Removes `claude-config` from `extraKnownMarketplaces` in `claude/.claude/settings.json` — the entry had `path: ""` as a placeholder for `install.sh`, but Claude Code reads this key natively. An empty `directory`-source path resolves relative to the current project directory and produces "Marketplace file not found at `<cwd>/.claude-plugin/marketplace.json`" in every non-claude-config project.
- Updates `install.sh` to self-register the `claude-config` marketplace from `$REPO_DIR` before the loop. The jq query now emits all `extraKnownMarketplaces` entries with source type; the bash loop skips non-github entries with a visible warning (`! <name>: non-github source '<type>' — skipping`) and registers github entries as before. The old `directory`/`case` arm is removed.
- Updates `README.md` to note that `./install.sh` auto-registers the `claude-config` marketplace, with the manual `claude plugin marketplace add` command as a fallback for setups that skip `install.sh`.

## Root cause

PR #210 added the `claude-config` entry so `install.sh` could auto-register the marketplace. The `path: ""` was deliberate — `install.sh` substitutes `$REPO_DIR` at runtime — but PR #210's own verification note flagged the exact risk and the smoke test missed it. The invariant that fails: `extraKnownMarketplaces` is consumed by both `install.sh` (overrides directory paths) and Claude Code's native parser (does not). Only portable values (github sources) belong in committed config.

## Test plan

- [x] `jq -e . claude/.claude/settings.json` passes
- [x] `bash -n install.sh` passes
- [ ] `/plugin` in a non-claude-config project opens cleanly (no "Marketplace file not found")
- [ ] `lovable-cloud@claude-config` skills still load (resolves via `known_marketplaces.json`, untouched)
- [ ] `./install.sh` re-registers `claude-config` from `$REPO_DIR` on a fresh machine; idempotent on second run

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ This repo exposes a marketplace via `.claude-plugin/marketplace.json`. Each plug
 
 **Current plugins:**
 
-One-time setup per machine: register the marketplace with `claude plugin marketplace add <path-or-URL-to-claude-config>`. Then install any of the plugins below at project scope:
+`./install.sh` registers the `claude-config` marketplace automatically. To register it manually (without running `install.sh`), run `claude plugin marketplace add <path-to-claude-config>`. Then install any of the plugins below at project scope:
 
 - **`lovable-cloud`** — Skills for Lovable Cloud projects: Project/Workspace Knowledge fields, edge-function auth model (ES256 gateway constraint, two-tier auth), and migration-sync workflow. `claude plugin install lovable-cloud@claude-config --scope project`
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #239, opened 2026-05-15
feat(check-runner): narrow charter, anchor cwd, and stop invented verdict counts

jcdendrite:
## Summary

- Narrows `check-runner.md` charter to read-only verification commands (test, lint, typecheck, build); explicitly refuses setup/state-mutating commands (db resets, migration apply, container lifecycle, DB seed, package installs) with a structured `NOT RUN — out of charter` per-command verdict
- Adds cwd-anchoring requirement: check-runner must `cd` to a parent-supplied absolute path as its first standalone Bash call; returns `FAIL` if none is provided
- Adds umbrella-command discipline: a single enumerated command that internally runs multiple sub-suites returns exactly one verdict entry (name, exit code, pass/fail) — no per-sub-suite breakdown, no test counts; parent reads the spool file if detail is needed
- Updates `CLAUDE.md` "Heavy command output" dispatch instructions to require an absolute working directory and prohibit setup commands in the command list
- Appends `docs/design-decisions.md` §11 recording the wrong-directory db reset incident and §12 recording the invented-test-count incident; both resolve by foundational prose scoping rather than layered heuristic hardening (consistent with §10)
- Two plan files committed (`check-runner-might-need-more-scalable-zebra.md`, `read-claude-research-check-runner-test-c-composed-graham.md`) per repo convention

## Root causes addressed

**Charter/cwd issue (§11):** A dispatch that included a db reset alongside the test command had check-runner run the reset from the wrong worktree's directory, applying a different worktree's migrations to the shared local DB. The test command then failed with a schema-cache miss that read like a genuine test failure — costing ~18 min of agent runtime plus a long diagnostic session.

**Verdict over-reporting issue (§12):** An umbrella command that internally runs multiple sub-suites produced a 1.1 MB spool file. The haiku model free-form-summarized it and grabbed two salient-looking counts with no grounding in which runner emitted each line; those wrong counts were propagated into a PR body before discovery.

## Test plan
- [x] `pytest claude/.claude/` — passing
- [ ] Manual: re-read `check-runner.md` and confirm a dispatch list containing a db-reset command would be refused with a `NOT RUN — out of charter` named-command verdict
- [ ] Manual: confirm `CLAUDE.md` "Heavy command output" and agent body agree on the same contract (parent passes absolute cwd; setup commands stay out of the dispatch list)
- [ ] Manual: confirm a dispatched umbrella command returns one verdict entry with no per-suite numbers, and the failure-excerpt rule is still unambiguous on a failed run


claude/.claude/agents/check-runner.md
```diff
@@ -8,6 +8,10 @@ maxTurns: 20
...
 
+**Scope: checks only.** check-runner runs read-only verification commands — test suites, lint, typecheck, and build. It does not run environment or fixture setup, or any state-mutating command. If the enumerated list contains any of the following, do not run it: `supabase db reset`, any `<tool> db reset` / `<tool> db push`, migration generation or application, `docker` container start/stop/restart, database seeding, or package installs (`npm install`, `pip install`, etc.). These commands mutate shared state and are directory-sensitive; running them from a subagent with no guaranteed cwd can apply the wrong state and produce misleading check failures. Report each refused command in the structured verdict as a per-command entry: name the command, status `NOT RUN — out of charter (setup/state-mutating command)`, exit code `null`. If any command was refused, set the overall verdict to FAIL. Do not silently skip it.
+
```
jcdendrite: The tone of this (third-person, passive) does not match the direct, second-person voice of other paragraphs. Need to standardize this.

The skill-review skill makes it clear that language should be direct.

I think we need to start running skill-review on agent files. Those probably should be in scope. Check Anthropic docs.

claude/.claude/CLAUDE.md
```diff
@@ -24,6 +24,8 @@
...
 
+**The dispatch prompt must include the absolute working directory** (e.g. `Working directory: /absolute/path/to/worktree`). check-runner has no guaranteed cwd of its own; a directory-sensitive command run from the wrong tree produces misleading failures. Do not enumerate environment- or fixture-setup commands in the check-runner command list — `supabase db reset`, migration apply, `docker` start/stop, dependency install, DB seed — check-runner runs checks only; perform setup in the correct directory yourself before dispatching. (check-runner is a plain `Agent`, not `isolation: worktree`, so passing an explicit working directory does not conflict with the Agent Briefing rule that forbids a `Working directory:` line for worktree-isolated agents.)
```
jcdendrite: Wait wait wait - why wouldn't we use isolation: worktree on the check-runner agent?! I didn't know that was an option. Wouldn't that solve the issue?

claude/.claude/agents/check-runner.md
jcdendrite: **[Claude Code]** Fixed (voice). Converted "Scope: checks only." to imperative second person — "Run read-only verification commands only…", "Do not run environment or fixture setup…", "Refuse these —…" — matching the rest of the file. `disposition: fixed-as-requested`

Skill-review on agent files: agreed it is the right direction, but the change is larger than this PR — the skill body and gate hooks are SKILL.md-specific (frontmatter fields, marker recipe, require-skill-review.sh pathspecs all need updating), and there is a design choice between extending skill-review vs. a new agent-review skill. Tracked as a separate follow-up: #241. `disposition: deferred`

claude/.claude/CLAUDE.md
jcdendrite: **[Claude Code]** Short answer: it breaks check-runner's actual job. An isolated worktree is a fresh git checkout at a committed ref — it does not carry the parent session's uncommitted working-tree changes, and a bare checkout has no installed dependencies, build artifacts, or local environment. check-runner's whole purpose is to verify the code the parent is *currently editing* (pre-commit); running in an isolated worktree would silently verify stale committed state against a dependency-free environment. It also would not have prevented the original incident — the wrong-directory db reset hit the shared local Supabase DB, which a git-worktree boundary does not isolate.

Documented this reasoning in docs/design-decisions.md §11 ("isolation: worktree rejected" paragraph) and tightened the CLAUDE.md parenthetical to name the reason. `where-documented: docs/design-decisions.md §11`

claude/.claude/agents/check-runner.md
```diff
@@ -8,6 +8,10 @@ maxTurns: 20
...
 
+**Scope: checks only.** Run read-only verification commands only — test suites, lint, typecheck, and build. Do not run environment or fixture setup, or any state-mutating command. If the enumerated list contains any of the following, do not run it: `supabase db reset`, any `<tool> db reset` / `<tool> db push`, migration generation or application, `docker` container start/stop/restart, database seeding, or package installs (`npm install`, `pip install`, etc.). Refuse these — they mutate shared state and are directory-sensitive; run from a subagent with no guaranteed cwd, they can apply the wrong state and produce misleading check failures. Report each refused command in the structured verdict as a per-command entry: name the command, status `NOT RUN — out of charter (setup/state-mutating command)`, exit code `null`. If any command was refused, set the overall verdict to FAIL. Do not silently skip it.
```
jcdendrite: Remember this is a generic repo applying to multiple types of projects. I'd be hesitant to name specific commands/technologies like `supabase` unless they are canonical. The placeholders seem better.

claude/.claude/CLAUDE.md
```diff
@@ -24,6 +24,8 @@
...
 
+**The dispatch prompt must include the absolute working directory** (e.g. `Working directory: /absolute/path/to/worktree`). check-runner has no guaranteed cwd of its own; a directory-sensitive command run from the wrong tree produces misleading failures. Do not enumerate environment- or fixture-setup commands in the check-runner command list — `supabase db reset`, migration apply, `docker` start/stop, dependency install, DB seed — check-runner runs checks only; perform setup in the correct directory yourself before dispatching. (check-runner is deliberately a plain `Agent`, not `isolation: worktree` — an isolated worktree is a fresh checkout missing the parent's uncommitted changes and prepared environment, so checks would verify the wrong tree; see `docs/design-decisions.md` §11. Passing an explicit working directory therefore does not conflict with the Agent Briefing rule that forbids a `Working directory:` line for worktree-isolated agents.)
```
jcdendrite: This is extremely verbose. I think you can trim it. Remember that CLAUDE.md applies to ALL projects. Spending this many words/lines on check-runner seems wastesful.

claude/.claude/agents/check-runner.md
jcdendrite: **[Claude Code]** Fixed. Replaced tool-specific names (`supabase db reset`, `docker`) with generic placeholders (`<tool> db reset`, `container start/stop/restart`, `<package-manager> install`) — matching the pattern already used in the rest of the paragraph. Addressed in the latest commit on this branch. `disposition: fixed-as-requested`

claude/.claude/CLAUDE.md
jcdendrite: **[Claude Code]** Trimmed. Cut the worktree-rationale parenthetical (that reasoning lives in `docs/design-decisions.md` §11) and replaced the tool-specific setup command list with generic categories. Paragraph is now ~40 words vs ~100. Addressed in the latest commit on this branch. `disposition: fixed-as-requested`

jcdendrite (commented):
lgtm

PR #240, opened 2026-05-15
fix(staff-backend-engineer): wire Write tool and harden file-based review output

jcdendrite:
## Summary
- Adds `Write` to `staff-backend-engineer` agent tools (root cause of all write failures)
- Rewrites the `### File-based output` section: mandates the `Write` tool explicitly, pre-empts the stock "don't create .md docs" instruction conflict, forbids false success reporting, and hardens the terse-return rule
- Updates `/code-review` SKILL.md line 215: requires synchronous spawn; adds conditional read-back (success → read file; failure → use inline findings)

## Background
Audit of conversation logs (2026-04-28 → 2026-05-14) found the technique was exercised 3 times and failed every time across two projects.

## Test plan
- [ ] Manual smoke test: run `/code-review` on a backend diff; confirm the spawned `staff-backend-engineer` uses the `Write` tool (not `cat`), returns only the pointer line, file exists on disk, parent reads it in the same turn
- [ ] `pytest claude/.claude/` passes (no hook/skill structure regressions)
- [ ] `ruff check claude/.claude/` passes


jcdendrite (commented):
lgtm

PR #228, opened 2026-05-14
perf(lovable-cloud): trim knowledge skill description to restore edge-functions to listing

jcdendrite:
## Summary

- `lovable-cloud:lovable-cloud-edge-functions` was being dropped from the skill listing (name-only, no description) because `lovable-cloud-knowledge`'s 762-char description exceeded the 1% default `skillListingBudgetFraction`
- Trimmed the description from 762 → ~361 chars; savings (~401) exceed the dropped skill's description (338), so the listing is net smaller
- Moved SKILL.md §1 ("The four sources Lovable loads") into REFERENCES.md — it's architectural background, not actionable guidance; its source quotes were already there
- Renumbered §2–§5 → §1–§4; added one-line cross-ref from §1 to REFERENCES.md for the full four-source priority list
- Bumped plugin to 2.1.1

## Activation

After merge, run in each consuming repo to pull the updated cache:

```
claude plugin install lovable-cloud@claude-config --scope project
```

## Test plan

- [ ] `claude plugin install lovable-cloud@claude-config --scope project` in DayTag
- [ ] Open a fresh DayTag session, run `/doctor` — confirm no skill descriptions are dropped
- [ ] Confirm `lovable-cloud:lovable-cloud-edge-functions` appears in the system-reminder skill listing with its full description
- [ ] Confirm `lovable-cloud:lovable-cloud-knowledge` still auto-routes on `.lovable/*.md` edits in a DayTag session

## Notes

skill-review skipped: changes are non-behavioral — all TRIGGER/DO NOT TRIGGER routing signals are preserved; §1–§4 (formerly §2–§5) functional guidance is unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #242, opened 2026-05-16
fix(lovable-cloud-edge-functions): extend TRIGGER to cover pg_cron and invocation files

jcdendrite:
## Summary

- The prior TRIGGER clause named only edge function code and `supabase/config.toml`, missing the case where a pg_cron migration adds a job that *invokes* an edge function without touching the function code or config.toml.
- New clause generalizes from enumerated files to a category — "modifying a file that configures or invokes an edge function" — with concrete examples (`supabase/config.toml`, `pg_cron job`) so the router can match on literal diff tokens without re-deriving the inference.
- DO NOT TRIGGER tightened from the weak "not working on edge functions or config.toml" to a positive-negative: "the change neither configures, invokes, nor modifies an edge function."
- No body edit needed — Tier 2 already documents pg_cron/pg_net callers as first-class; this aligns the routing signal to the body's existing scope.
- Bumps plugin to `2.1.2` and syncs `marketplace.json` from `2.1.0` to `2.1.2`, reconciling the drift left over from #228.

## Test plan

- [ ] Verify YAML frontmatter parses correctly (validated via `python3 -c "import yaml; ..."` — passes)
- [ ] `/skill-review` run on staged diff — clean (behavioral-equivalence audit: all removed lines preserved in new text)
- [ ] `/code-review` run on staged diff — clean
- [ ] Smoke test: apply to a session that sees a pg_cron migration touching an edge-function invocation and confirm the skill fires
- [ ] After merge: run `claude plugin install lovable-cloud@claude-config --scope project` in consuming repos to pull 2.1.2 into the plugin cache

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
     {
       "name": "lovable-cloud",
       "description": "Skills for Lovable Cloud projects — Project/Workspace Knowledge fields, edge-function auth model, and migration-sync workflow.",
-      "version": "2.1.0",
+      "version": "2.1.2",
       "author": {
         "name": "Cordova Strategy"
       },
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #243, opened 2026-05-16
fix(staff-backend-engineer): remove mkdir priming step and correct heredoc rationale

jcdendrite:
## Summary

- Removes the `mkdir -p agent-reviews` step from `staff-backend-engineer`'s file-based-output section — the step was both redundant (the `Write` tool creates parent directories) and harmful (primed the agent into Bash-mode, causing it to continue with a shell heredoc for the actual file write).
- Corrects the rationale for avoiding heredocs: the original said "permission-gated and will be denied" — inaccurate; the observed failure was a shell command-length parser abort on a multi-KB findings block. An agent that believes the risk is only "denial" may try a small heredoc anyway; the correct failure-mode description is what makes the prohibition credible.
- Adds a canary-scope note to `code-review/SKILL.md:215` documenting that `findings_path` may only go to `Write`-capable agents (currently `staff-backend-engineer` alone), so a future routing edit can't silently re-arm the heredoc-abort bug for a non-`Write` reviewer.

## Root cause

In a live `code-review` session, `staff-backend-engineer` was observed running:
1. `Bash: Create agent-reviews directory` (the now-removed step 1)
2. `Bash: Write findings to agent-reviews directory` — a heredoc carrying the full review

Step 2 overran the Bash command-length limit → parser abort → truncated output → unexpected permission prompt. `ciso-reviewer` and `staff-sdet`, both inline-only, finished cleanly — confirming the bug was scoped to the file-writing agent.

## Test plan

- [x] `pytest claude/.claude/` — 801 tests pass
- [x] `/code-review` run (Claude Code config domain) — clean; `/skill-review` invoked for the SKILL.md edit
- [ ] End-to-end: run `/code-review` on a backend-touching diff and confirm `staff-backend-engineer` writes its findings via the `Write` tool, returns the one-line pointer, and the parent reads the file back (this is the canary's pass criterion before any rollout to the other agents)


jcdendrite (commented):
lgtm

PR #244, opened 2026-05-16
feat(evals): skill trigger-fidelity harness + 3 pilot skills

jcdendrite:
## Summary

- Adds `evals/run_trigger_evals.py` — a local-only batch harness (adapted from Anthropic's `run_eval.py` mechanism) that runs `claude -p` against per-skill `trigger-cases.json` files and reports trigger fidelity as a human-read pass-rate table
- Ships 3 pilot case files (`code-review`, `test-conventions`, `test-evaluation`) covering positive, negative, and cross-skill confusion cases via `also_not_triggered`
- Adds CI-safe unit tests: a static shape validator for all `trigger-cases.json` files (discovery-based, no model call), and a deterministic detector unit test fed by committed stream-json fixtures
- Updates docs (`CONTRIBUTING.md`, `docs/skills.md`, `skill-review/SKILL.md`) and `REFERENCES.md` to reflect the adopted harness

## Why local-only, never CI

Three constraints rule out CI: (1) `--dangerously-skip-permissions` on a public repo is a footgun; (2) CI would require `ANTHROPIC_API_KEY` — local runs use Max-plan OAuth, no per-token charge; (3) skill auto-trigger is a probabilistic model classification, not a deterministic computation — binary pass/fail produces a flaky signal. All three constraints apply only to CI. Local runs: `python evals/run_trigger_evals.py --skill <name> --verbose`.

## Detection mechanism

Spawns `claude -p <query> --output-format stream-json --verbose --include-partial-messages` from a throwaway temp project whose `.claude/skills/` is symlinked to the working-tree skills (approach A — real skills, real mutual competition). Parses stream-json for `content_block_start → tool_use` with `name == "Skill"` (not `Read` — `Read`'s input is a file path, not a skill name; including it caused false-positive fires on any `Read` whose path contained the skill name). Accumulates `input_json_delta`, then at `content_block_stop` parses the accumulated JSON and compares `input["skill"] == skill_name` exactly (not substring). Early-terminates the subprocess once the target fires and no `also_not` guards remain; reads to completion/timeout when misfire guards are active so every block is observed. K=3 samples per case, pass threshold 50%.

## Implementation notes

- `evals/` is on pytest's `pythonpath` (via `pyproject.toml`) so `test_trigger_detector.py` can import the harness module; `evals/` itself is not a test root and is not collected by `pytest claude/.claude/`
- No additional dependencies needed: PyYAML was removed along with the unused `parse_skill_frontmatter` function (skill filtering is handled by case-file presence, not frontmatter)
- Fixture JSONL files carry synthetic identifiers (session UUIDs, request IDs, thinking-block signatures replaced with obvious placeholders); no real session artifacts in the public repo

## Test plan

- [x] `pytest claude/.claude/` — 808 tests pass; `test_trigger_cases_files_well_formed` and all `test_trigger_detector.py` tests pass (including `test_read_of_skill_named_path_no_match` regression guard); `evals/` not collected (verified locally and in CI)
- [x] `ruff check claude/.claude/` — clean (verified locally and in CI)
- [ ] `python evals/run_trigger_evals.py --skill code-review --verbose` — confirm harness spawns `claude -p`, detection fires, and report shows per-case fidelity
- [ ] Run the `test-conventions` / `test-evaluation` pair and verify `also_not_triggered` cases name which sibling fired on a mis-fire


claude/.claude/skills/test-evaluation/evals/trigger-cases.json
```diff
@@ -0,0 +1,31 @@
...
+      "id": "evaluating-existing-suite-quality",
+      "query": "Look at our existing test suite and assess the coverage quality and test pyramid shape.",
+      "should_trigger": true
+    },
+    {
+      "id": "debugging-flaky-tests",
+      "query": "Our CI is failing intermittently. Some tests pass locally but fail in CI. Help me debug these flaky tests.",
+      "should_trigger": true
+    },
+    {
```
jcdendrite, line 14: What should get triggered here? Shouldn't `test-conventions` get triggered when asking about testing patterns? Or is the intention to spawn staff-sdet agent?

evals/fixtures/no-trigger-typo.jsonl
```diff
@@ -0,0 +1,34 @@
...
+{"type":"assistant","message":{"model":"claude-sonnet-4-6","id":"msg_01NtayPpoCC5uKhSgB8zEWy4","type":"message","role":"assistant","content":[{"type":"thinking","thinking":"The user wants me to fix a typo in a README file. Let me find it first.","signature":"EoQCClsIDRgCKkCSh6z1rYty5KpqY1pR+HuZgCasAwYZIx+3GsxdNXllQKcLlf8qDjvDTRDsbTtldycZ4a6LNblU6ZcEYgNnIT7gMhFjbGF1ZGUtc29ubmV0LTQtNjgAEgxHMbf4hn4eHcmSWGsaDLL0SgEKvKOS4dJWPSIwQHQ/lEnc/6ReHD6OVgrsXREXb55qppUhEGQJRAxzoiJjU3yTTXk4WlHruxC9fl3CKlfd2J/EEP7lZYTZK823V5yUjYk1sxMVftBHXVccytM6uxAB0Vi6UNoru3CY/n16UAbFULQLe1Oxl7LnUuu9y4v72ZKdvFs0migHhRtw6RZLaq4DO08r6n4YAQ=="}],"stop_reason":null,"stop_sequence":null,"stop_details":null,"usage":{"input_tokens":3,"cache_creation_input_tokens":8686,"cache_read_input_tokens":14878,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":8686},"output_tokens":6,"service_tier":"standard","inference_geo":"not_available"},"diagnostics":null,"context_management":null},"parent_tool_use_id":null,"session_id":"471300e7-7df7-4610-afe1-c228ff9bc54d","uuid":"d6112a94-1ee8-4dd6-8958-caeba30ed4a9","request_id":"req_011Cb5HCZBBZnxsJXRQyC6oZ"}
```
jcdendrite: I hope all these secrets are synthetic. Can you confirm?

claude/.claude/skills/test-evaluation/evals/trigger-cases.json
jcdendrite: **[Claude Code]** Good catch — `also_not_triggered: ["test-conventions"]` was wrong. The query ("I'm about to write tests for a new OAuth flow… What testing patterns should I use?") is squarely `test-conventions`' TRIGGER ("planning tests for new code… test strategy for new features"), so test-conventions *should* fire here — the field was asserting the opposite.

The harness has no positive cross-skill assertion field; `also_not_triggered` is only for skills that must *stay silent*. The fix is to drop the field; the case still correctly asserts that `test-evaluation` itself stays silent for a new-tests query. The positive direction is already covered by `test-conventions`' own `should_trigger: true` cases.

The same bug existed in the mirror case: `vs-test-evaluation-debugging-flaky` in `test-conventions/evals/trigger-cases.json` had `also_not_triggered: ["test-evaluation"]` on a flaky/slow-test debug query — exactly `test-evaluation`'s TRIGGER. Fixed both. `disposition: fixed-with-modification | rationale: the sibling case in test-conventions had the identical wrong assertion; both removed | commit: addressed in the latest commit on this branch`

evals/fixtures/no-trigger-typo.jsonl
jcdendrite: **[Claude Code]** No API keys, tokens, or credentials — `grep` for key/token/secret/password/bearer patterns is clean across all fixtures, and every transcript shows `apiKeySource: "none"`.

That said, to be straight: the `signature` (thinking-block integrity hashes), `session_id`, `request_id`, `uuid`, and message-`id` values in the original fixtures were *real* artifacts captured from an actual Claude Code session, not synthetic. None of them are credentials — the signatures are integrity hashes bound to specific thinking content, and the IDs are opaque/non-replayable — so there is nothing to rotate. But a public repo should not carry real session artifacts even if harmless.

The detector (`detect_trigger_in_lines`) only reads `event`/`content_block`/`delta` fields, so all of those values were inert ballast. Fixed by scrubbing every identifier to obvious synthetic placeholders (`00000000-0000-0000-0000-000000000001`, `msg_SYNTHETIC0001`, `SYNTHETIC_SIGNATURE`, etc.) across all three fixtures, preserving JSON structure and line count. Unit tests still pass. `disposition: fixed-with-modification | rationale: truthful answer was the values were real; scrubbed rather than just confirming | commit: addressed in the latest commit on this branch`

jcdendrite (commented):
lgtm

PR #250, opened 2026-05-16
fix(check-runner): prevent stale spool contamination via one-call discipline

jcdendrite:
## Summary

- **Root cause:** The previous per-command procedure described two steps — run the command, then write/read the spool. This opened a window where a separate Bash call could locate the spool via a glob (`${TMPDIR:-/tmp}/<slug>-*.txt`), matching all stale spool files from prior sessions sharing the same slug prefix. In one incident, ~85 stale files matched and produced a 67 MB read; the agent confabulated a verdict from the mixture.
- **Fix (check-runner.md):** Each command now runs in one self-contained Bash call. The spool path is echoed *before* the command runs so the path survives a Bash-tool timeout kill — if the harness terminates the call mid-command, the pre-run echo has already landed and the TIMEOUT verdict can reference the partial spool. The call then redirects command output to the spool, captures the exit code, and emits a bounded `tail` inline. An explicit prohibition added: never locate a spool file with a glob or `ls`. Also: the failure-excerpt verdict bullet now reflects that the subagent sees only the inline tail (not the full spool), and a note clarifies that `<command>` must be a single simple command with no embedded `;` or shell control operators.
- **Fix (docs/design-decisions.md):** New §13 documents the incident, the two contributing factors, and the resolution — including the echo-before-command ordering and the timeout-recovery rationale. Consistent with the §10–§12 pattern of foundational scoping over layered detection heuristics.

## Test plan

- [x] `pytest claude/.claude/` and `ruff check claude/.claude/` — skipped; diff touches only agent config (`.claude/agents/check-runner.md`) and docs, no executable code

### DIFF
--- a/claude/.claude/agents/check-runner.md
+++ b/claude/.claude/agents/check-runner.md
@@ -18,19 +18,21 @@ Run each enumerated command exactly once. On non-zero exit, capture and move to 
 
 Every Bash call must include the tool's `timeout` parameter set to 600000 (10 minutes). Do not omit it. A command that exceeds 10 minutes is treated as a TIMEOUT verdict for that command — capture whatever output the spool file holds and proceed to the next command.
 
-For each command, run it in one self-contained Bash call — not split across multiple calls. The single call computes the epoch, runs the command, captures the exit code, and emits the result inline:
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #251, opened 2026-05-16
docs(engineering-judgment): add failing-check attribution rule

jcdendrite:
## What

Adds one Engineering Judgment bullet to `claude/.claude/CLAUDE.md`: before treating a failing check as in-scope work, prove the current change caused it by reproducing the failure on the pre-change baseline.

> **Prove your change caused a failing check before treating it as in-scope.** When a check fails, reproduce it on the pre-change baseline — a throwaway worktree at the merge-base (`git worktree add <tmp> $(git merge-base HEAD origin/main)`) — before assuming the failure is yours. Failing there too means pre-existing drift, not this task's work: sync the branch if the base already fixed it, but never hand-reimplement a merged fix or edit an unrelated file. Passing at the baseline but failing on your branch means your change caused it — in scope even if you never touched the failing file, since a change breaks dependents through their imports.

## Why

A real scope-explosion failure mode: a one-line task ran the full check suite on a branch that was an ancestor of the default branch, saw the suite's pre-existing failures, and attributed them to itself — ballooning to 6 file edits, including an approval-gated security-invariant test. The root cause was treating a failing check as in-scope without proving the change caused it.

The rule makes the attribution signal **differential reproduction against the baseline**, not file-overlap. File-overlap is only a hint about where to look: a change can break a dependent through its imports without touching that file, so "didn't edit the file" cannot prove the failure isn't yours.

## Notes

- Stack-agnostic and judgment-altitude — matches the length of its neighboring `Engineering Judgment` bullets.
- No skill files touched: the worktree mechanic is a one-line git primitive, and the downstream "sync the branch" action routes to `git-feature-branch-sync` on its own.
- Reviewed via `/ai-instruction-and-memory-files` and `/code-review`: correct destination, no PR/ticket refs, no duplication of an existing rule.


claude/.claude/CLAUDE.md
```diff
@@ -9,6 +9,7 @@
...
 - **Audit structural siblings before scoping a fix narrowly.** When a fix lands in one arm of a multi-arm structure (case statement, switch, parallel subcommands, sibling functions sharing a shape), check the other arms for the same bug shape before finalizing scope. If the fix is identical, apply to every affected site; abstract when two or more share it. Scope is set by the bug, not by where the symptom surfaced.
+- **Prove your change caused a failing check before treating it as in-scope.** When a test/lint/verify failure appears, reproduce it on the pre-change baseline — a throwaway worktree at the merge-base (`git worktree add <tmp> $(git merge-base HEAD origin/main)`) — before assuming the failure is yours. Failing there too means pre-existing drift, not this task's work: sync the branch if the base already fixed it, but never hand-reimplement a merged fix or edit an unrelated file. Passing at the baseline but failing on your branch means your change caused it — in scope even if you never touched the failing file, since a change breaks dependents through their imports.
```
jcdendrite: `test/lint/verify` - careful about using non-standard names for test suites. You may want to consider using the broader, more accepted term for testing and linting, maybe just those - testing and linting, or "checking" or all 3. Feel free to do some web research to see what is standard language that is paradigm and language agnostic.

jcdendrite: **[Claude Code]** Fixed — replaced `test/lint/verify` with `check`. Quick research confirmed "check" / "CI checks" is the paradigm- and language-agnostic umbrella term (GitHub formalizes it in its Checks API); "verify" was a stack-specific script name (a Maven lifecycle phase / npm-script convention). Went with the single umbrella term rather than enumerating "testing and linting" — that would omit type-check and build, and the bullet's lead clause already says "failing check", so the body now mirrors it. `disposition: fixed-as-requested` — addressed in the latest commit on this branch.

jcdendrite (commented):
lgtm

PR #252, opened 2026-05-16
feat(claude-hook-review): spawn staff-platform-engineer on hook reviews

jcdendrite:
## Summary

- **Section 10 (new):** `claude-hook-review` now spawns `staff-platform-engineer` synchronously after the section-9 checklist, with four specific latency/failure-mode questions. Also conditionally spawns `ciso-reviewer` when the hook gates a security boundary (auth checks, secrets, env-var reads, private-data redaction, credential handling). This fires on both entry paths: `code-review` → `claude-hook-review`, and `claude-hook-review` auto-triggered directly.
- **Section 7 (extended):** Added deterministic checklist line: external commands that contact a daemon, socket, or network (`docker`, `systemctl`, `curl`, package managers) must run under an explicit timeout.
- **REFERENCES.md:** Records the motivating failure mode as edit-time rationale for both additions (properly abstracted per repo redaction rules).
- **plugin.json:** Version bumped `1.0.0` → `1.1.0` (additive capability, backward-compatible per `plugin-semver`).

## Context

`code-review` item 36 names `staff-platform-engineer` as primary owner of "Hook correctness," but the ownership was declared and never actuated — neither `code-review` nor the `claude-hook-review` delegate it fires had a step that escalated to the platform persona. A hook PR that shelled out to an external daemon command exposed this: it passed `/code-review` clean, and a follow-up manual review found an unbounded-latency bug.

The fix lands at the delegate (`claude-hook-review`) rather than at `code-review`'s triage table, so it covers both invocation paths without editing `code-review`'s routing rows (which would itself trigger a full multi-persona review per the "Reshapes reviewer ownership" triage row). See plan file at `claude/.claude/plans/code-review-hook-platform-routing.md` for the full Gap A / Gap B analysis and mechanism notes.

## Test plan

- [x] `pytest claude/.claude/` — 807 tests pass
- [x] `ruff check claude/.claude/` — clean
- [x] `/code-review` — no findings
- [x] `/skill-review` — no findings
- [x] `claude-hook-review/SKILL.md` line count: 125 (≤200 gate)
- [ ] Manual smoke test: invoke `claude-hook-review` on a sample hook diff that shells to an external command; confirm `staff-platform-engineer` is spawned and its findings surface

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #253, opened 2026-05-16
docs(claude-hook-review): surface agent escalation in plugin description

jcdendrite:
## What

Follow-up to #252. v1.1.0 added Section 10 (operational-footprint escalation to `staff-platform-engineer`), but the `claude-hook-review` plugin description still advertised only "the 9-item review checklist" — understating the headline capability of that release, and reading ambiguously now that the skill has a literal Section 10.

## Change

- Update the description in **both** `plugins/claude-hook-review/.claude-plugin/plugin.json` and the root `.claude-plugin/marketplace.json` to name the escalation capability and drop the now-ambiguous "9-item" count. The two strings are kept verbatim-identical to avoid drift.
- Bump `plugin.json` version `1.1.0 → 1.1.1` — a documentation clarification correcting metadata to match already-shipped behavior (semver patch). No `version` field added to the marketplace entry.

No skill behavior changes.

## Plugin semver checklist

- [x] `plugin.json` `version` bumped (1.1.0 → 1.1.1)
- [x] Bump magnitude matches the change (patch — clarification)
- [x] No `version` field added to the marketplace.json entry
- [x] Post-merge install command below

## Post-merge install

Consuming repos refresh with:

```
claude plugin install claude-hook-review@claude-config --scope project
```

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -26,7 +26,7 @@
     },
     {
       "name": "claude-hook-review",
-      "description": "Review playbook for Claude Code hooks: event/matcher selection, path resolution, script skeleton, fail-open/fail-closed posture, dispatch drift, and the 9-item review checklist. Install at project scope in repos that author .claude/hooks/*.sh scripts.",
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #254, opened 2026-05-17
docs(claude-md): add codebase-discovery dispatch guidance

jcdendrite:
## Summary

- Adds `### Codebase discovery` as a sibling H3 under `### Heavy command output` in `claude/.claude/CLAUDE.md`
- Generalizes the existing "parent context inhales the output" dispatch principle from check-suite commands to broad locate-style reads and grep/glob sweeps
- Uses harness-builtin `Explore`/`general-purpose` subagent types — no new agent files (adding agents costs Opus description-load, which was explicitly rejected)
- Includes a comprehension-read carve-out so the rule doesn't contradict existing design (code-review reads every changed file inline — that read is the review; plan-review project-layer reads are intentionally inline per design-decisions §8)

## Motivation

In auto mode on a Max plan, the parent session is locked to Opus. An audit of current dispatch habits found: check-runner handles suite output (mandatory); reviewer personas run in isolated Sonnet/Haiku context; `/plan-it` and `/ready-for-review` dispatch discovery. One gap: general codebase discovery outside a skill — broad grep/glob sweeps, locate-style reads — lands inline on the Opus parent with no guidance to do otherwise.

## Test plan

- [ ] Re-read the new section in context: sits cleanly next to `### Heavy command output`, consistent "parent context inhales the output" terminology, no contradiction with `## Model Routing`
- [ ] Confirm instruction reads stack-agnostically (auto-mode/Opus clause framed as sharpest case, not universal claim — Team/Enterprise/API users can run `ANTHROPIC_MODEL=sonnet claude-auto`)

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -28,6 +28,12 @@ Use the `Agent` tool with `subagent_type: check-runner` to run the checks (full 
 
 If the subagent reports a check was blocked by permissions, do NOT fall back to running it directly in the parent — that defeats the dispatch (parent context inhales the output) and silently bypasses the permission gate. Stop, surface the exact missing allow-rule (e.g. `Bash(npm run verify)`), and wait for the user to either pre-approve it in the appropriate scope's `permissions.allow` or run the command themselves.
 
+### Codebase discovery
+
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #255, opened 2026-05-17
docs(claude-md): fix heading misplacement and plugin/skill mislabels

jcdendrite:
## Summary

- Re-home the review-pipeline paragraph under its own `## Review pipeline` heading — it was orphaned under `## When editing a skill, run the skill on its own diff`, a different topic.
- Move `## Review pipeline` to immediately after `## Working in this repo` — the section is gate-defining (hook-enforced) and belongs before the process sub-rules that follow from it, not after them.
- Tighten the pipeline enumeration: `/plan-review` and `/code-review` are called out as unconditionally mandatory; `/skill-review` is noted as conditional (fires when SKILL.md is staged) and invoked automatically by `/code-review`.

Note: an earlier commit changed `claude-hook-review` from "skill" to "plugin"; this was reverted per reviewer feedback — `claude-hook-review` is a skill that happens to ship inside a plugin, and the skill identity is the right call-site label.

Found via independent audit of the root `CLAUDE.md` prompted by a `/claude-md-management:claude-md-improver` run.

## Test plan
- [ ] Re-read `CLAUDE.md` — `## Review pipeline` appears immediately after the `## Working in this repo` section ends, before `## Plans in this repo affect all stow users`.
- [ ] `grep -n "skill-review" CLAUDE.md` — conditional note reads "When staged changes include a SKILL.md, `/skill-review` is also required; `/code-review` invokes it automatically."
- [ ] `grep -n "claude-hook-review" CLAUDE.md` — still reads "skill" (not "plugin").
- [ ] File stays under 200 lines.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,7 +45,7 @@ X…", "before/after X…" — the answer is a hook configured in
 `.claude/settings.json`, not a memory or a skill instruction. The
 harness executes hooks; nothing in memory or a CLAUDE.md prose rule can
 fulfill an automatic-trigger request. Route to the `claude-hook-review`
-plugin for hook design and review.
+skill for hook design and review.
 
 **Plugin config:** `enabledPlugins` only takes effect in
 `settings.json`, not `settings.local.json`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,7 +71,8 @@ not want to extend. Mirror existing entries when in doubt.
 `/plan-it` is the prescribed entry for plan creation. `/plan-review` and
 `/code-review` are mandatory pipeline steps before PR handoff — both are
 hook-enforced (see README "Workflow" for the full skill invocation order and
-which hook gates each transition).
+which hook gates each transition). When staged changes include a SKILL.md,
+`/skill-review` is also required; `/code-review` invokes it automatically.
 
 ## Plans in this repo affect all stow users
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,11 +68,10 @@ not want to extend. Mirror existing entries when in doubt.
 
 ## Review pipeline
 
-`/plan-it` is the prescribed entry for plan creation. `/plan-review`,
-`/code-review`, and `/skill-review` are mandatory pipeline steps before PR
-handoff — all three are hook-enforced (see README "Workflow" for the full
-skill invocation order and which hook gates each transition).
-`/skill-review`'s gate fires only when staged changes include a SKILL.md; it ships in the bundled `skill-review` plugin (not under `claude/.claude/skills/`).
+`/plan-it` is the prescribed entry for plan creation. `/plan-review` and
+`/code-review` are mandatory pipeline steps before PR handoff — both are
+hook-enforced (see README "Workflow" for the full skill invocation order and
+which hook gates each transition).
 
 ## Plans in this repo affect all stow users
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,7 +45,7 @@ X…", "before/after X…" — the answer is a hook configured in
 `.claude/settings.json`, not a memory or a skill instruction. The
 harness executes hooks; nothing in memory or a CLAUDE.md prose rule can
 fulfill an automatic-trigger request. Route to the `claude-hook-review`
-skill for hook design and review.
+plugin for hook design and review.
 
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -45,7 +45,7 @@ X…", "before/after X…" — the answer is a hook configured in
 `.claude/settings.json`, not a memory or a skill instruction. The
 harness executes hooks; nothing in memory or a CLAUDE.md prose rule can
 fulfill an automatic-trigger request. Route to the `claude-hook-review`
-skill for hook design and review.
+plugin for hook design and review.
```
jcdendrite: You've got the dichotomy wrong. It's not plugin vs. skill. It's that the skill happens to be part of a plugin. Both claude-hook-review and skill-review are still skills.

CLAUDE.md
```diff
@@ -87,11 +87,13 @@ check the diff against its output — e.g. an edit adding prose to a
 skill that argues for brevity is the kind of thing the skill would
 flag against itself.
 
+## Review pipeline
```
jcdendrite: Would it be helpful for this section to be moved earlier in CLAUDE.md since it's a pivotal part of engineering judgment / workflow? Explain why or why not.

CLAUDE.md
jcdendrite: **[Claude Code]** You are right — the dichotomy in my original commit was wrong. `claude-hook-review` is a skill that ships inside a plugin; the skill identity is the correct label at the call site. Reverted to "skill" in the latest commit on this branch.

CLAUDE.md
jcdendrite: **[Claude Code]** Slight lean toward leaving it in place, for two reasons:

1. The `## Review pipeline` heading (added in this PR) already addresses the discoverability concern — it gives the paragraph a navigable anchor even at line 90.
2. The pipeline section references `/plan-review`, `/code-review`, and `/skill-review` by name. If it moved to immediately after `## Commands`, readers would encounter those tool names before the "Working in this repo" section explains what skills, plugins, and hooks are. The current position lets the infrastructure context land first.

That said, the concern is real — it is the most operationally important constraint in the file. If you want it moved, the right destination would be just after `## Working in this repo` (so readers have the vocabulary) rather than right after `## Commands`. Happy to make that move if you decide it is worth it.

jcdendrite: **[Claude Code]** Moved. `## Review pipeline` now sits immediately after `## Working in this repo` ends, before `## Plans in this repo affect all stow users`. The section is now in front of the process sub-rules that follow from it rather than after them.

CLAUDE.md
```diff
@@ -66,6 +66,14 @@ not want to extend. Mirror existing entries when in doubt.
 
 **Global skill bodies stay platform-agnostic.** Skills under `claude/.claude/skills/` are stowed to every user who clones this repo — their bodies must read cleanly regardless of stack. Encode the generic concept; do not hardcode engine/platform-specific tokens (`pg_cron`, `net.http_post`, vendor API names). Stack-specific examples and checks belong in a project-layer skill (`<skill>-<project>/SKILL.md`) that the base skill loads via its project-layer glob — e.g. `/code-review`'s Step 0.5 globs `.claude/skills/code-review-*/SKILL.md`.
 
+## Review pipeline
+
+`/plan-it` is the prescribed entry for plan creation. `/plan-review`,
+`/code-review`, and `/skill-review` are mandatory pipeline steps before PR
```
jcdendrite: Wait, wait, `skill-review` is conditional based on what type of file is being reviewed. I think skill-review may be invoked from code-review, so I think it's fine to omit. You added a caveat to say this later, but I worry that it's verbose and potentially confusing to the agent.

jcdendrite: **[Claude Code]** You are right that calling it unconditionally mandatory was wrong. However, omitting it entirely creates a contributor blind spot — the commit hook fires when SKILL.md is staged and an agent with no CLAUDE.md signal hits it unexpectedly.

Fixed with a conditional framing: "When staged changes include a SKILL.md, `/skill-review` is also required; `/code-review` invokes it automatically." This addresses both concerns — accurate conditionality, and no verbose standalone caveat sentence. `disposition: fixed-with-modification` | `rationale: conditional note preserves hook discoverability without misrepresenting it as universally mandatory`

jcdendrite (commented):
lgtm

PR #257, opened 2026-05-17
fix(plan-review): gate ignores committed-clean plans; remove stow-leaked plan files

jcdendrite:
## Summary

- `require-plan-review.sh` replaced bare file-existence check with a per-file git-status check: a plan file only arms the gate if it is untracked or modified vs HEAD. Committed, unmodified plan files are treated as historical and pass through.
- Fixes a latent `find(1)` precedence bug where `-maxdepth 1` was not applied to the `.txt` branch of the `-o` expression.
- Removes 14 scratch plan files that `stow --adopt` had leaked into the tracked `claude/.claude/plans/` package directory. Adds `claude/.claude/plans/` to `.gitignore`; the existing root-anchored `.claude/plans/` entry does not cover this path.
- Adds 3 tests covering the new allow/deny paths (committed-clean → allow; modified-committed → deny; mixed committed + untracked → deny).
- Updates `docs/hooks.md` gate-deadlock section and `README.md` gate table row.

**Root cause:** The gate's signal was session- and content-agnostic — it armed whenever any plan file existed in `.claude/plans/`, whether that file was from a session 6 months ago or the current one. A completion marker dying with the session meant every subsequent session re-evaluated the same stale plan.

**Accepted limitation (documented in hook):** committing a plan file before running `/plan-review` makes the gate treat it as historical. The prescribed `/plan-it` → `/plan-review` → commit-with-implementation flow does not hit this path.

## Test plan

- [x] `pytest claude/.claude/` — 836 tests (833 existing + 3 new), all pass
- [ ] Manual: create a scratch repo, confirm the three manual verification steps in the plan's Verification section produce expected deny/allow/deny sequence
- [ ] Confirm `git ls-files claude/.claude/plans/` returns nothing after merge


jcdendrite (commented):
lgtm

PR #258, opened 2026-05-17
feat(transcript-analysis): add transcript analysis toolkit

jcdendrite:
## Summary

- Adds `transcript-analysis.py` — a read-only CLI with six subcommands for analyzing Claude Code session transcripts stored in `~/.claude/projects/`
- Adds `test_transcript_analysis.py` — full test suite covering helpers, regex patterns, and all six subcommands (862 tests pass)
- Adds `claude/.claude/skills/transcript-analysis/SKILL.md` — command-skill with subcommand routing table, fail-seq interpretation guide, caveats, and examples; description uses plain usage language (no TRIGGER/DO NOT TRIGGER — correct for user-invoked command skills)
- Registers `transcript-analysis` in `COMMAND_SKILLS` so it doesn't consume the model's always-loaded skill-listing budget
- `STRUGGLE_PHRASES` sourced from transcript analysis — `"hold on"` is attested; predicted phrases retained alongside

## Subcommands

| Subcommand | Question answered |
|------------|-------------------|
| `buckets` | Which branches exist? What model family was used on each? |
| `fail-seq` | Did a debugging loop converge or thrash? |
| `struggle` | Did the user express frustration more with one model? |
| `duration` | How much time was active vs idle? |
| `subagents` | How much work ran in subagents vs the main thread? |
| `pr-link` | Which PRs map to which branches? How many review comments? |

## Test plan

- [x] `pytest claude/.claude/` — 862/862 pass
- [ ] Run `python3 ~/.claude/scripts/transcript-analysis.py buckets` against your own `~/.claude/projects/` and verify output shape matches the SKILL.md table

### DIFF
--- a/claude/.claude/scripts/transcript-analysis.py
+++ b/claude/.claude/scripts/transcript-analysis.py
@@ -21,24 +21,36 @@ TEST_RUNNER_RE = re.compile(
 FAILED_RE = re.compile(r"\b(\d+)\s+failed\b")
 
 STRUGGLE_PHRASES: list[str] = [
+    # attested in transcripts
+    "hold on",
+    "why did you",
+    "try again",
+    "no not that",
+    # predicted patterns
     "no, that",
     "that's wrong",
     "not right",
     "you're wrong",
     "stop doing",
     "don't do that",
-    "why did you",
     "still broken",
     "still failing",
-    "try again",
     "you missed",
     "incorrect",
     "not what i asked",
     "wrong approach",
-    "no not that",
     "that doesn't work",
     "please don't",
 ]
+
+
+def _projects_glob(args: argparse.Namespace) -> str:
+    return getattr(args, "projects", None) or "*"
+
+
+def _branch_filter(args: argparse.Namespace) -> set[str] | None:
+    raw: str | None = getattr(args, "branches", None)
+    return {b for b in raw.split(",") if b} if raw else None
 
 
 def _fam(model: str) -> str:
@@ -104,10 +116,8 @@ def _longest_fail_streak(failed_flags: list[bool]) -> int:
 
 
 def cmd_buckets(args: argparse.Namespace) -> None:
-    projects_glob: str = getattr(args, "projects", None) or "*"
-    branch_filter: set[str] | None = (
-        {b for b in args.branches.split(",") if b} if getattr(args, "branches", None) else None
-    )
+    projects_glob = _projects_glob(args)
+    branch_filter = _branch_filter(args)
 
     branch_data: dict[str, dict] = defaultdict(
         lambda: {"sessions": 0, "opus": 0, "sonnet": 0, "haiku": 0, "other": 0, "ts_min": float("inf"), "ts_max": float("-inf")}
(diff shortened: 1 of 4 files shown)


claude/.claude/scripts/transcript-analysis.py
```diff
@@ -0,0 +1,468 @@
...
+    "still failing",
+    "try again",
+    "you missed",
+    "incorrect",
+    "not what i asked",
+    "wrong approach",
+    "no not that",
+    "that doesn't work",
+    "please don't",
+]
```
jcdendrite, line 44: Did you come up with your own list of struggle phrases or did you actually semantically analyze the wording I use when I corrected Claude's actions? It's fine to start with your best judgment of phrases that would fit, but please also include the ones I actually use.

claude/.claude/skills/transcript-analysis/SKILL.md
```diff
@@ -0,0 +1,54 @@
+---
+name: transcript-analysis
+description: Run the transcript-analysis.py toolkit to analyze Claude Code transcripts — model comparison by branch, test-failure convergence sequences, correction-signal frequency, active-vs-idle duration decomposition, subagent-vs-main turn split, or PR-to-branch mapping. TRIGGER when: the user asks which model was used on a branch, wants to check whether a debugging loop converged or thrashed, wants to scan for frustration signals, wants to see how active vs idle time breaks down on a branch, wants to know how much work ran in subagents vs the main thread, or wants to link branches to PRs with comment counts. DO NOT TRIGGER when: the user wants token-cost or cache-efficiency analysis (use token-analyzer.py directly), or when the question is about a specific file's content rather than transcript-wide patterns.
```
jcdendrite: TRIGGER and DO NOT TRIGGER don't really make sense for command-types skills, i.e. when the model can't invoke it. What does skill-review skill say to do?

jcdendrite (commented):
I also noticed in the conversation context in terminal you mentioned these as non-blockers -- why defer them?

> Code review: Two non-blocking hygiene notes (repeated projects_glob/branch_filter patterns — post-merge cleanup candidate).

claude/.claude/scripts/transcript-analysis.py
jcdendrite: **[Claude Code]** Fixed with modification. Added `"hold on"` — attested in transcripts (`"Hold on - [criticism]"` pattern appears consistently). Moved `"why did you"`, `"try again"`, `"no not that"` to an explicit `# attested in transcripts` section; original predicted phrases are kept under `# predicted patterns`. Dropped `"ugh"` — it is a substring of common words like "tough" and "through", making it a false positive risk. `disposition: fixed-with-modification` | `rationale: "ugh" substring collision; "hold on" is the main new attested phrase` | latest commit on this branch

claude/.claude/skills/transcript-analysis/SKILL.md
jcdendrite: **[Claude Code]** Fixed. The skill-review checklist requires TRIGGER/DO NOT TRIGGER for model-invokable skills, but has no exception for command skills — and the test suite enforces `disable-model-invocation: true` on everything in `COMMAND_SKILLS`. The correct fix was to drop the TRIGGER/DO NOT TRIGGER language (meaningless when the model cannot auto-invoke) and replace with a plain usage description, keeping `disable-model-invocation: true`. Latest commit on this branch. `disposition: fixed-as-requested`

jcdendrite:
**[Claude Code]** Fair push-back — extracted both now. `_projects_glob(args)` (6 callsites) and `_branch_filter(args)` (4 callsites) are in the latest commit. The code-review at handoff time labeled them non-blocking, but that was the wrong call; they were in-scope by the "three or more instances" rule. `disposition: fixed-as-requested`

jcdendrite (commented):
lgtm

PR #259, opened 2026-05-18
fix(evals): split skill eval harness into runtime + description-fidelity modes

jcdendrite:
## Summary

A smoke test of the skill trigger-fidelity harness found it measures `code-review` correctly (5/5) but returns a uniform 0/10 for the advisory skills `test-conventions` and `test-evaluation`. That zero is a measurement artifact of a confirmed won't-fix Claude Code limitation (anthropics/claude-code#34648 — headless `claude -p` does not reliably auto-trigger skills, especially advisory `user-invocable: false` ones), not a skill defect.

This PR splits the harness measurement model so it stops emitting false FAILs, and ships the verified `code-review` fix. A follow-up PR adds the `description-fidelity` runner.

## What shipped

- **Schema** — each `trigger-cases.json` declares a required `method`: `runtime` (measured here via `claude -p` auto-dispatch) or `description-fidelity` (measured by a separate runner, not yet built). `load_case_file()` rejects a missing or unknown method.
- **Harness** — `discover_case_files()` results are partitioned by method; only `runtime` skills run. Non-runtime skills are reported with a one-line skip notice instead of a false FAIL.
- **code-review** — marked `runtime`; keeps the fixture-seeding helper and the K=10 / 90s-timeout changes that make its real-diff measurement reliable.
- **test-conventions / test-evaluation** — marked `description-fidelity`.
- **Tests** — method validation, mode partitioning, skip-notice text, and an end-to-end skip-path test (no `claude -p`).
- **Docs** — `evals/README.md` documents the two measurement methods and the K=10 sampling noise floor.

## Test plan

- [ ] Reviewer judgment: confirm the two-mode split is the right seam, vs. converting every skill to description-fidelity. Rationale: runtime measurement observes real auto-dispatch and is strictly more faithful when a skill is runtime-triggerable; collapsing to one mode would discard that signal for `code-review`.

### DIFF
--- a/claude/.claude/skills/code-review/evals/trigger-cases.json
+++ b/claude/.claude/skills/code-review/evals/trigger-cases.json
@@ -1,5 +1,6 @@
 {
   "skill_name": "code-review",
+  "method": "runtime",
   "cases": [
     {
       "id": "code-written-commit-pending",
@@ -8,7 +9,7 @@
     },
     {
       "id": "explicit-user-code-review-request",
-      "query": "Please do a code review of this function: def fetch_user(user_id): return db.query(User).filter_by(id=user_id).first()",
+      "query": "I've just finished implementing a new feature and I'm ready to commit. Can you do a formal code review of my changes?",
       "should_trigger": true
     },
     {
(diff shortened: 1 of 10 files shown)


jcdendrite (commented):
lgtm

PR #262, opened 2026-05-18
feat(scripts): skip removing worktrees that a live process is using

jcdendrite:
## Summary

`cleanup-merged-branches.sh` removed merged branches' worktrees with bare `git worktree remove`. A worktree removed while a live process was working inside it — a Claude Code session, a shell, a dev server, or a container bind-mounted into it — left that process serving a deleted directory.

This adds an OS-level guard:

- **`collect_process_cwds`** snapshots every readable process working directory once per run — Linux via `/proc/[0-9]*/cwd`, otherwise via `lsof -d cwd`.
- **`worktree_in_use`** matches a worktree path (canonicalized) against that snapshot. The per-branch cleanup loop and the `--dry-run` preview both skip a worktree that a live process is cwd'd inside; `--dry-run` tags it `[worktree in use — would skip]`.
- An empty snapshot is treated as undetermined, so worktrees are skipped conservatively rather than deleted on the strength of a scan that did not work.

The check sees process working directories only, and only the invoking user's — it does not catch a process holding the worktree by an open file descriptor without cwd'ing in, nor a container whose owning session has already exited. The script header comment and `docs/scripts.md` keep the "run when other sessions are idle" operational rule as a backstop.

## What shipped

- **Script** (`cleanup-merged-branches.sh`) — `collect_process_cwds` + `worktree_in_use` helpers; in-use guard in the per-branch loop with a new `Skipped (worktree in use…)` summary line; dry-run annotation; rewritten header comment.
- **Tests** (`test_cleanup_merged_branches.py`) — new `TestWorktreeInUseGuard`: in-use skip, in-use-via-subdirectory skip, idle-still-removed (no false positive), dry-run annotation.
- **Docs** (`docs/scripts.md`) — describes the guard and its residual gaps.

## Test plan

- [x] `pytest claude/.claude/` and `ruff check claude/.claude/` pass
- [ ] In a repo with a merged branch's worktree, run a process with cwd inside that worktree and confirm `cleanup-merged-branches` skips it; confirm `--dry-run` tags it
- [ ] Confirm an idle merged worktree is still removed


jcdendrite (commented):
lgtm

PR #265, opened 2026-05-18
Add opt-in PII/PHI guard hooks for commits and data-file reads

jcdendrite:
## Summary

Adds two opt-in PreToolUse hooks plus a security-hardening guide, for a claude-config adoption that must pass a strict PII/PHI security review. Both hooks are dormant for every stow user until a user-local config file is created — nothing changes behaviour by default.

**What shipped — hooks (`claude/.claude/hooks/`)**

- `deny-pii-in-commits.sh` — gates `git commit`. Armed by `~/.claude/pii-patterns.md`. Scans the added lines of the staged diff, the commit message, and `-F`/`--file` message files. Built-in generic patterns (US SSN, Luhn-checked credit-card runs) plus user `<label>: <regex>` patterns. Fires on every Bash call and detects the commit via `_lib` word-walking, so globally-flagged forms (`git -c key=val commit`, `git -C <path> commit`) and `&&`/`;`/`|` chains cannot evade the gate. `exclude: <glob>` lines drop synthetic-fixture paths; `--no-verify` does not disable it. Deny messages name the matched pattern by label only — never the value, never the user regex.
- `deny-data-file-reads.sh` — gates the `Read` tool. Armed by `~/.claude/data-file-read-guard.md`. Refuses data-shaped files (data extensions, `Downloads/`, files over 5 MB, configured path globs) before content enters context.

Both fail closed on malformed input and require their config file to be a readable regular file (a FIFO/special file is treated as not-armed).

**What shipped — docs & wiring**

- `docs/security-hardening.md` — arming guide, MCP-connector / telemetry-env-var / `permissions.allow` audit checklist, managed-settings rollout path, and a cloud-sync warning for the config files.
- `claude/.claude/settings.json` — wires both hooks; `docs/hooks.md` and the README Hook table get entries; `.gitignore` gets belt-and-suspenders entries for the two config files.
- Test suites: 76 tests across the two hooks (armed/unarmed, every pattern and rule, fail-closed paths, evasion forms, boundary conditions).

## Review notes for the human reviewer

- The commit hook deliberately has **no `if`-condition** in `settings.json` (fires on every Bash call, filters internally) — a narrowing `if: \"Bash(git commit *)\"` let `git -c key=val commit` slip past the dispatch. This matches the `require-worktree-for-git-writes.sh` pattern.
- Known gaps are documented in each hook header (editor-flow amend, chained `git add` staging timing, `-C` into a different repo, non-`Read` ingestion paths). Confirm these are acceptable for the threat model.
- `docs/security-hardening.md` cites telemetry env-var names and managed-settings paths verified against the current Claude Code settings docs at implementation time.

## Test plan

- [x] `pytest claude/.claude/` — 976 tests pass
- [ ] Manual smoke test on a machine with the hooks armed (arm via the two config files, confirm a synthetic-SSN commit is denied and a clean commit proceeds; confirm a `.csv` Read is denied and a source-file Read is allowed)
- [ ] Confirm the documented known-gaps are acceptable for the adopter's threat model


jcdendrite (commented):
lgtm

PR #266, opened 2026-05-19
docs: guide general-purpose subagent model routing under auto mode

jcdendrite:
## Problem

Auto mode anchors the session to a single model for its entire lifetime — there is no plan-mode-to-execution model switch the way `opusplan` provides. On Max, that anchor is Opus (the only auto-eligible session model). The `general-purpose` subagent is the one routinely-dispatched built-in with no `model:` of its own, so it falls through to the parent and runs delegated execution / whole-file discovery on Opus — roughly 5x the per-token cost of Sonnet.

A rule gated on "in auto mode" is unfollowable: a session cannot detect its own permission mode (no env var exposes it). So the fix is unconditional.

## Change

- **`claude/.claude/CLAUDE.md`** — one bullet under `## Model Routing`: always dispatch `general-purpose` with an explicit `model`, defaulting to `sonnet` (`opus` only when the delegated task needs Opus-level reasoning). Unconditional because passing `model: sonnet` is a harmless no-op when the parent is already Sonnet and the needed fix when it is Opus. Names the agents that must *not* get a `model` override (`Explore` is Haiku-pinned; `staff-*` / `ciso-reviewer` / `check-runner` carry their own frontmatter).
- **`docs/auto-mode.md`** — new "Subagent delegation under auto mode" section: the four-step subagent model resolution order, a per-built-in resolution table, and why `CLAUDE_CODE_SUBAGENT_MODEL` is the wrong instrument (it overrides every subagent, including the deliberate pins).
- **`claude/.claude/skills/transcript-analysis/SKILL.md`** — one-line correction in the Caveats section: the `fail-seq`/`struggle` caveat claimed reviewer and Explore subagents run on Sonnet; Explore runs on Haiku. Dropped the non-load-bearing model attribution and listed the excluded agent types (reviewer, `Explore`, `check-runner`) accurately.

## Verification

All factual claims in the doc table were checked:
- `staff-*` / `ciso-reviewer` → `sonnet`, `check-runner` → `haiku` — confirmed against `claude/.claude/agents/*.md` frontmatter.
- `Explore` → Haiku — confirmed empirically against three `agentType: "Explore"` subagent transcripts, all `claude-haiku-4-5-20251001`.
- The per-dispatch `model` param overriding auto-mode parent inheritance, and the auto-mode classifier covering subagent Bash calls, were verified in earlier standalone test sessions.

## Out of scope

- `/plan-it` and `/plan-review` dispatch `general-purpose` internally and would run it on Opus under an auto-mode session — a larger skill-body change, not bundled here.

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -63,6 +63,7 @@ This does not apply to *comprehension* reads: when you need a file's content in 
 - **Opus:** judgment-heavy reasoning, plan-mode planning, and parent-dispatcher orchestration.
 - **Sonnet (default):** all code reading, code writing, and specialist reviewer agents. Enforced via `model: sonnet` frontmatter in each agent file.
 - **Haiku:** narrow, deterministic skills only. Never for code authoring or judgment.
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

jcdendrite (commented):
lgtm

PR #267, opened 2026-05-19
feat(hooks): detect git/gh commands past global flags in redaction gate

jcdendrite:
## What

Harden `deny-private-project-refs.sh` — the redaction gate that blocks
private-project tracker IDs from reaching this public repo — so evasive
`git commit` / `gh pr` / `gh api` invocations can no longer slip past it.

Regex command-detection is replaced with word-walking fragment detection:
the gate now sees through `git -c` / `-C` / `--git-dir`, env-var prefixes,
absolute paths, and `$()` / backtick command chains. The four `if`-gated
`settings.json` hook entries collapse into one unconditional entry, so the
hook fires on every Bash call and filters internally — closing the gap
where `if: "Bash(git commit *)"` never even dispatched on `git -c … commit`.

## gh detection

`fragment_gh_gated_surface` keys on the contiguous command path — `pr`
immediately followed by `create` / `edit`, or the `api` word. cobra lets a
subcommand's own flags be written ahead of the subcommand
(`gh -X POST api …`, `gh --repo owner/repo pr create …` both parse), so the
walk cannot assume the first bare word after `gh` is the subcommand. This
evasion form was surfaced by the code-review security pass and is closed
here, not left as a known gap.

## Tests

New evasion-resistance cases in `test_deny_private_project_refs.py`: deny
paths for every evasion form (global flag prefixes, env-var prefixes,
absolute paths, `$()` / backtick chains, gh flags-before-subcommand), allow
paths confirming the regex→word-walk swap is behavior-preserving, and a
fail-closed test for an unsourceable `_lib.sh`. Full suite: 993 passing,
lint clean.

## Review

`/plan-review` and `/code-review` complete; `claude-hook-review` plus
ciso-reviewer, staff-sdet, and staff-platform-engineer consulted on the
hook's security boundary and operational footprint.

## Known gaps (documented in the hook header, not closed here)

Shell-evaluation-boundary wrappers (`eval`, `xargs`, `sh -c`, `$()` body
content) and `\git` alias-escape remain documented known gaps — closing the
latter belongs in a separate `_lib.sh` PR, since that word test is shared
by several hooks.

### DIFF
--- a/claude/.claude/hooks/deny-private-project-refs.sh
+++ b/claude/.claude/hooks/deny-private-project-refs.sh
@@ -7,13 +7,15 @@
 # repo-root CLAUDE.md redaction rule ("Redact private-project-
 # identifying content").
 #
-# NOTE — `if`-dispatch is advisory; the real gate is the internal regex
-# at the top of this script. settings.json wires `if` entries
-# (`Bash(git commit *)`, `Bash(gh pr create *)`, `Bash(gh pr edit *)`,
-# `Bash(gh api *)`) for zero-cost early dispatch, but any drift between
-# those patterns and the IS_GIT_COMMIT / IS_GH_PR / IS_GH_API regexes
-# here creates silent coverage gaps. Update both surfaces when extending
-# coverage.
+# Dispatch: wired on the PreToolUse `Bash` matcher with NO `if`-condition,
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #268, opened 2026-05-19
check-runner: silence verdict on success, prevent invented counts

jcdendrite:
## Summary
- check-runner's per-command wrapper ran `tail -50` on the spool unconditionally — including on a passing run — so the haiku agent's own context held each sub-suite's summary lines and it volunteered (sometimes wrong) test counts in the verdict.
- Structural fix instead of another prose-prohibition layer: the wrapper's `tail` is now conditional on a non-zero exit. A passing command's verdict is its exit code; the agent never reads the sub-suite summary lines, so it cannot report a count.
- Trimmed the now-redundant count-prohibition prose from the umbrella-command paragraph.
- `CLAUDE.md`: new "Reporting test counts" rule — the parent reports counts only by grepping the full spool for the runner's own summary lines and quoting them, never from the verdict or from memory.
- `docs/design-decisions.md`: §14 records the recurrence and why this is not the "grounded count-extraction procedure" §12 rejected — extraction moves out of check-runner entirely to a parent-side grep.

## Test plan
- [ ] Dispatch check-runner on a passing suite — verdict reports exit code only, no test counts or sub-suite breakdown.
- [ ] Dispatch check-runner on a failing suite — the failure excerpt (`tail -50`) still appears.
- [ ] A parent reporting test counts greps the spool and quotes the runner's own summary lines, rather than quoting the verdict.


jcdendrite (commented):
lgtm

PR #269, opened 2026-05-19
fix(marker): resolve live Claude PID from ancestor walk

jcdendrite:
## Problem

`marker.sh activate` resolved the Claude PID stamped into the active-bypass
marker by content-scanning `~/.claude/sessions/` for the file whose contents
matched `SESSION_ID` (`_resolve_claude_pid_for_session`). A crashed or
restarted session can leave a stale per-session file behind; the directory
scan could match the stale file first and stamp a **dead PID** into a freshly
written marker. A dead PID defeats the `kill -0` liveness check that
`require-*.sh` hooks and `clear-stale` rely on.

## Fix

Resolve the PID from the process ancestor walk that already runs to resolve
the session id — the walk reaches the live Claude main process *as a process
ancestor* of the script, so a stale `sessions/` file can no longer be matched
by content.

- `_walk_session` now prints `"<session_id> <pid>"` and `return`s 2 on failure
  (rather than `exit`ing) so command-substitution call sites guard with `|| exit 2`.
- Two thin resolvers — `_resolve_session_id`, `_resolve_claude_pid` — extract
  each field. `_resolve_session_id` keeps its original signature for the
  non-`activate` subcommands that need only the session id.
- `_resolve_claude_pid_for_session` (the reverse directory scan) is deleted.

## Tests

- `TestMarkerScriptStalePidLookup` — `activate` ignores a stale `sessions/`
  entry that sorts lexically before the live PID (what the old scan would have
  picked first); `write` arm still resolves the session id correctly.
- `TestMarkerScriptSessionMissing` docstring updated for `return`-vs-`exit`
  wording.

`pytest claude/.claude/` → 978 passed; `ruff check claude/.claude/` → clean.

### DIFF
--- a/claude/.claude/hooks/tests/test_marker_script.py
+++ b/claude/.claude/hooks/tests/test_marker_script.py
@@ -24,9 +24,10 @@ class TestMarkerScriptSessionMissing:
     """When the session file ($HOME/.claude/sessions/$PPID) does not exist,
     every write/activate/deactivate subcommand must exit 2 and write nothing.
     This guards against the silent-empty-suffix regression: without explicit
-    || exit 2 on the command-substitution call sites, exit 2 inside
-    _resolve_session_id() only exits the subshell — the parent continues and
-    writes a malformed marker with an empty session-id suffix."""
+    || exit 2 on the command-substitution call sites, the return 2 inside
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #270, opened 2026-05-19
feat(hooks): guard skipAutoPermissionPrompt in settings commit gate

jcdendrite:
## Problem

`~/.claude/settings.json` is a GNU stow symlink to this repo's tracked, public `claude/.claude/settings.json`. Claude Code persists the user-scoped `skipAutoPermissionPrompt` preference into `~/.claude/settings.json` whenever the key is absent — so it lands straight in the stow-distributed repo file and shows up as an uncommitted diff. There is no Claude-Code-side way to redirect that write (no user-scoped `settings.local.json` exists; the `.local.json` variant is project-scope only). Committing the key would ship a permission-prompt-skipping default to every stow user.

An existing commit guard already gates `git commit` against `model`/`effortLevel` changes in staged `settings.json` for exactly this class of accidental commit — it just didn't cover `skipAutoPermissionPrompt`.

## Change

- **Guard a third key.** `skipAutoPermissionPrompt` joins `model`/`effortLevel` in the staged-vs-`main` comparison — two `jq` extractions and one comparison, consistent with the existing differential semantics. New test mirrors `test_model_change_denies_commit`.
- **Rename the hook.** `guard-settings-model-effort.sh` → `guard-settings-session-keys.sh`, since the gate now covers three keys and the old name under-described it. Updates the `settings.json` wiring and `statusMessage`, the test module + its constant + test class, and the `docs/hooks.md` entry (which had also still listed only two keys).
- **Harden the hook.** Adds `set -uo pipefail`, and routes all five git calls through a `git_capped` wrapper that caps each at 5s via `timeout`, so a pathologically slow git cannot hang the synchronous pre-commit gate. On timeout a call fails like any other git error and the hook falls through to its existing no-opinion exit.

The rename and hardening were raised in this PR's code review and deferred as out-of-scope at the time; they are brought in here at the maintainer's request.

## Verification

- `pytest claude/.claude/` and `ruff check claude/.claude/` — green.
- `/code-review` (cumulative, vs `main`) clean. `claude-hook-review` and `staff-platform-engineer` (operational-footprint) reviewed the final hook; one comment-precision note on worst-case latency, addressed.


jcdendrite (commented):
lgtm

PR #271, opened 2026-05-19
feat(agents): add code-writer subagent with in-agent self-review

jcdendrite:
## What

Adds a `code-writer` subagent: delegated code-writing now routes to it instead
of the built-in `general-purpose` agent, and it self-reviews its own diff
before returning.

## Why

Delegated code-writing on `general-purpose` commits review-finding-class
defects — N+1 query shapes, missing idempotency on retryable writes, unhandled
error/empty states — that the parent's `/code-review` then catches. Each catch
costs a parent → review → re-dispatch round-trip. `general-purpose` also has no
`model:` of its own and inherits the parent.

`fail-seq` analysis across recent code-heavy branches showed near-zero
test-failure thrash — so the corrections are review-finding-class, not test
failures. The fix targets that class.

## How

- **`claude/.claude/agents/code-writer.md`** (new) — `model: sonnet`, scoped
  tools. Four-section system prompt: charter, positive-framed implementation
  baseline, mandatory pre-return self-review, structured return.
- The self-review reads the `staff-*` reviewer agent files **live** and
  verifies the diff against their review angles. No copied checklist ⇒ nothing
  to drift; when a reviewer gains an angle, `code-writer` inherits it. Step 3
  resolves each reviewer file from the routing table by name, so the table is
  the single in-file source of the reviewer set.
- **`CLAUDE.md`** — Model Routing sends delegated code-writing to `code-writer`
  (substitution only — delegation frequency unchanged).
- **`README.md`, `docs/auto-mode.md`, `docs/design-decisions.md`** (new entry
  §15, with verified sources) updated.

## Review

- `/plan-review`: Approve with changes — all applied.
- `/code-review`: clean — config/docs diff, no runtime/security/schema surface;
  no specialists needed.

## Follow-up

A manual smoke test — dispatch `code-writer` and confirm it performs the
self-review, not just writes code — is pending; there is no automated
`claude -p` eval harness in this repo by design. To be run in a fresh session
post-merge.


claude/.claude/agents/code-writer.md
```diff
@@ -0,0 +1,103 @@
+---
+name: code-writer
+description: Implements a specified code change, then self-reviews its own diff before returning. TRIGGER when delegating code-writing or implementation work to a subagent — feature code, bug fixes, refactors, migrations, schema, config, scripts. Returns a structured summary of what was implemented, what the self-review verified, and anything still uncertain. DO NOT TRIGGER for read-only exploration, codebase search, or research — use `Explore` or `general-purpose` for those.
```
jcdendrite, line 3: Do TRIGGER and DO NOT TRIGGER clauses work with agents or do they only work with skills? I realize that the reviewer agents already include these fields, but the check-runner does not. I want you to reference official documentation.

claude/.claude/agents/code-writer.md
```diff
@@ -0,0 +1,103 @@
...
+Once the change is written and the baseline holds, review your own diff as if it
+were someone else's pull request. Reviewing a finished diff is a sharper,
+more focused task than writing it — use that asymmetry deliberately.
+
+1. Re-read your complete diff, file by file. For each hunk, ask what a reviewer
+   would flag.
+2. Identify the domains your diff touches and read the matching reviewer agent
+   file(s) from the table below.
+3. Locate those files by absolute path — the `Read` tool does not expand `~`.
+   Run `ls ~/.claude/agents/staff-*.md ~/.claude/agents/ciso-reviewer.md` in
```
jcdendrite: This could become stale if we add more agents. Might be better to identify all expert/reviewing agents collectively so that this doesn't get stale. It's buried in this agent's file. This might be buried elsewhere in the repo in some skill as well. Best to standardize how we refer to these reviewer agents.

claude/.claude/agents/code-writer.md
```diff
@@ -0,0 +1,103 @@
...
+| Your diff touches | Reviewer file(s) to read |
+|---|---|
+| Server/API handlers, RPCs, background jobs, queue consumers, app schema | `staff-backend-engineer`, `ciso-reviewer`, `staff-sdet` |
+| Client components, hooks, client state, routing, forms | `staff-frontend-engineer`, `staff-sdet` |
+| Database migrations, pipelines, CDC / ETL | `staff-data-engineer` |
+| Warehouse models, transformations | `staff-analytics-engineer` |
+| CI/CD config, IaC, shell scripts | `staff-platform-engineer` |
+| User-visible behavior, copy, flows | `staff-product-engineer` |
+| Auth, authorization, secrets, trust boundaries | `ciso-reviewer` |
+| Test code | `staff-sdet` |
```
jcdendrite, line 87: Why did you deviate from referring to the routing table in ROUTING.md? Is it intentional to have a smaller copy of the table here for some reason? It's just another place where things might get stale and deviates from "single-source-of-truth" principle.

claude/.claude/agents/code-writer.md
jcdendrite: **[Claude Code]** Verified against official docs. A subagent `description` is free-form natural language — the orchestrator reads it to decide automatic delegation ("Claude uses each subagent's description to decide when to delegate tasks", https://code.claude.com/docs/en/subagents). "TRIGGER" / "DO NOT TRIGGER" are **not** parsed keywords for either agents or skills; a Skill's `description` is the same kind of field — free-form natural language used for selection. So the clauses work for `code-writer` by the same mechanism they work for the reviewer agents: the whole description feeds the delegation decision. The docs' own suggested convention for nudging delegation is phrasing like "use proactively." `check-runner` omits trigger clauses because the parent only ever dispatches it explicitly (it is never auto-delegated), so it needs no delegation-trigger guidance — `code-writer`, like the reviewer agents, is meant to be auto-delegatable, so the clauses earn their place. `disposition: AGREE-NO-CHANGE` | `rationale: clauses are effective natural-language delegation guidance for an auto-delegatable agent; no change warranted.`

claude/.claude/agents/code-writer.md
jcdendrite: **[Claude Code]** Fixed in the latest commit on this branch. Step 3 carried a parallel hardcoded set — a `staff-*.md` glob plus an explicit `ciso-reviewer.md` — duplicating the reviewer set the domain table right below it already names, so a new reviewer meant updating two places. Step 3 now resolves each reviewer file from the table by name, leaving the table as the single in-file source of which reviewer agents exist. `disposition: fixed-with-modification` | `rationale: scoped the fix to the in-file duplication. The broader point — standardizing how reviewer agents are referred to repo-wide (they are also enumerated in plan-review/ROUTING.md and in each agent's own frontmatter) — is a real cross-file concern but a separate effort from adding this one agent; expanding this PR to a repo-wide standardization would overreach its scope. Flagging it for a follow-up.`

claude/.claude/agents/code-writer.md
jcdendrite: **[Claude Code]** Intentional, not an oversight. Agent files have no import/include mechanism — the `@path` import syntax works in `CLAUDE.md` only, not in agent or skill files — so a "reference" to ROUTING.md could only ever be a prose pointer. And the repo's root `CLAUDE.md` ("No shared partials across skills") explicitly prescribes duplicating shared rule text into each file rather than extracting it, precisely to keep each file independently readable and avoid brittle cross-coupling. `plan-review/ROUTING.md` is also not a shared resource — it is plan-review's *runtime auxiliary*, which `CLAUDE.md` classifies as belonging to that one skill; an agent pointing into it would couple `code-writer` to plan-review's internals. The two tables are not the same table either: `ROUTING.md` routes a *plan* against plan-review's checklist items (one domain → one agent); `code-writer`'s table routes a *diff's domains* to reviewer-file *combinations* to mine for self-review angles — different consumer, different shape. Your staleness point is fair — a new `staff-*` reviewer does mean updating both tables — but that is the tradeoff `CLAUDE.md`'s "duplicate intentionally" rule already accepts repo-wide. `decision: code-writer keeps a self-contained domain→reviewer table` | `where-documented: root CLAUDE.md "No shared partials" section; docs/design-decisions.md §14.`

jcdendrite (commented):
lgtm

jcdendrite (commented):
changes from review lgtm

PR #273, opened 2026-05-19
docs: add property-based delegation test to global CLAUDE.md

jcdendrite:
## What

Adds a `### Delegation: hold conclusions, not transcripts` subsection to the global `claude/.claude/CLAUDE.md`, placed under `## Working Style` immediately before `### Heavy command output`.

## Why

The global `CLAUDE.md` routes delegation by *task type*: `### Heavy command output` (check suites → `check-runner`) and `### Codebase discovery` (code search → `Explore`/`general-purpose`). Type-indexing names individual slices and leaves the unnamed majority of read-heavy `Bash` work running inline on an auto-mode Opus parent — verbose output the parent only needs a conclusion from, re-billed on every turn. Adding one more task-type subsection would name one more slice and leave a fresh un-named "other".

## How

The new subsection states a generative principle and a two-part test, applied to *objectives* not individual commands:

- **Output test** — does the parent's reasoning consume the command's output, or only a conclusion drawn from it?
- **Judgment test** — does choosing the command (and the next) need parent-grade judgment?

Both pointing to delegate ⇒ dispatch the objective to a subagent. The two existing subsections survive unchanged as named instances of the principle (their titles are untouched — `README.md` references "Heavy command output" by title). `Edit`/`Write` is explicitly excluded as a delegation target — those runs are short and judgment-dense.

The subsection cross-references `## Model Routing` and `docs/auto-mode.md` for the dispatch-cost mechanics rather than restating them.

## Verification

- `pytest claude/.claude/` — 995 passed
- `ruff check claude/.claude/` — clean
- `/code-review` (clean) and `/ai-instruction-and-memory-files` audit (behavior test, no PR/ticket refs, correct CLAUDE.md-vs-skill routing) both pass

### DIFF
--- a/claude/.claude/CLAUDE.md
+++ b/claude/.claude/CLAUDE.md
@@ -21,6 +21,21 @@
 - Use descriptive variable and function names. No generic names.
 - When a session crosses ~60% context usage (visible via statusline `.context_window.used_percentage`) AND the current task is incomplete, proactively suggest the user run `/handoff`. Do not run it yourself — it's a user-invoked slash command. If the user agrees, the command writes `/tmp/<slug>-handoff.md` and provides the resume incantation. Run `analyze-context` if unsure about the threshold. (See README's "Threshold reference" section for the why-60%.)
 
+### Delegation: hold conclusions, not transcripts
+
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #274, opened 2026-05-19
feat(hooks): garbage-collect stale ~/.claude/sessions/ files

jcdendrite:
## Summary

`capture-session-id.sh` writes a bare-PID `session_id` mapping file to `~/.claude/sessions/` on every SessionStart/SubagentStart and never removed superseded ones, so the directory grew without bound (~895 files observed).

This adds a throttled garbage-collection sweep to the hook:

- Runs at most once per 24h, gated by a `~/.claude/.sessions-gc-stamp` timestamp sentinel — kept outside `sessions/` so that directory stays purely bare-PID files plus Claude Code's own `<pid>.json` sidecars.
- Deletes superseded prior incarnations of the current session, files whose PID is dead (`kill -0`), and files older than 30 days (the PID-reuse ceiling — `kill -0` can pass for an unrelated recycled PID).
- A 5-minute mtime floor protects a just-written sibling session's file; a `^[0-9]+$` basename filter excludes Claude Code's `<pid>.json` sidecars and is the single gate every deletion passes.
- Fails open — a SessionStart hook that fails closed would block session startup.

`marker.sh` is intentionally untouched. Retiring the bare-PID mechanism in favor of Claude Code's `<pid>.json` sidecars was considered and deferred — see the commit message.

## Test plan

- [ ] With a populated `~/.claude/sessions/`, start a fresh session; confirm stale bare-PID files are pruned while every `<pid>.json` sidecar survives.
- [ ] Start a second session within 24h; confirm the sweep is throttled (file count unchanged).

### DIFF
--- a/claude/.claude/hooks/capture-session-id.sh
+++ b/claude/.claude/hooks/capture-session-id.sh
@@ -68,4 +68,83 @@ for _active_dir in "$HOME/.claude"/.*-active.d; do
   fi
 done
 
+# Garbage-collect stale bare-PID files from ~/.claude/sessions/.
+#
+# Throttled to once per 24 hours via ~/.claude/.sessions-gc-stamp so the
+# sweep does not run on every SubagentStart in an agentic session. The
+# stamp lives outside sessions/ to keep that directory purely bare-PID
+# files plus Claude Code's <pid>.json sidecars.
+#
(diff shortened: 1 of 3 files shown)


jcdendrite:
Superseded by #275, which replaces the throttled sweep + three deletion heuristics with a SessionEnd destructor + content-match guard. See #275 for the design discussion.

PR #275, opened 2026-05-19
feat(hooks): add SessionEnd destructor for sessions/<pid> files

jcdendrite:
## Summary

`capture-session-id.sh` (SessionStart, SubagentStart) writes a `session_id` ↔ claude-PID lookup file at `~/.claude/sessions/<pid>`. Nothing removed those files; the directory grew monotonically with session count, reaching ~895 stale entries in ~2 weeks.

This PR adds `cleanup-session-id.sh` as the destructor `capture-session-id.sh` never had. Registered matcher-less on `SessionEnd`, it resolves the claude-PID via the same `ps -o ppid= -p $PPID` walk `capture-session-id.sh` uses, then deletes `~/.claude/sessions/<pid>` only when the file's content still equals the ending session's `session_id`.

## Why the content-match guard, not a blind `rm`

`/clear` ends one session and starts another in the *same* claude process — same claude-PID, new `session_id`. Both this `SessionEnd` hook (old id) and the `SessionStart` capture (new id) touch `sessions/<pid>`, and Claude Code does not order them. The guard makes both orderings correct:

- `SessionEnd` first → file holds old id → match → deleted, then `SessionStart` rewrites for the successor session.
- `SessionStart` first → file already holds new id → no match → kept (it now belongs to the live successor session).

The guard reduces deletion to one condition — "does this file still map to the session that is ending?" — replacing PR #274's three stacked heuristics with a single content equality check.

## Supersedes PR #274

PR #274 ([branch `sessions-dir-gc`](https://github.com/jcdendrite/claude-config/pull/274)) addressed the same observation with a throttled, once-per-24h directory *sweep*: three deletion heuristics (superseded-content, dead-PID, 30-day-age), a 5-minute mtime floor, and a `~/.claude/.sessions-gc-stamp` throttle file. That is a garbage collector — cron-like scheduling reimplemented inside a session-start hook — and its stacked, gap-closing heuristics are the wrong-foundation tell: each layer existed to paper over the previous layer's imperfection.

This PR replaces ~250 lines of hook + tests in #274 with a ~50-line destructor + 7 short tests. The directory self-bounds by *lifecycle pairing* rather than by a sweep — structurally, to roughly one file per currently live session.

After this lands, PR #274 will be closed unmerged.

## Residual leak — explicitly accepted

A session killed before its hooks can run (`SIGKILL`, OOM, power loss, abrupt terminal close) leaves its file behind. This residue is:

- *Bounded by crash rate, not session rate* — grows at a tiny fraction of the pre-fix rate.
- *Cosmetic, not a correctness defect* — `marker.sh`'s forward `_resolve_session_id` walk only reads `sessions/<pid>` for a live ancestor PID; the reverse `_resolve_claude_pid_for_session` matches on unique session-id UUIDs.
- *Self-healing on PID reuse* — `capture-session-id.sh` overwrites unconditionally; when the OS recycles a leaked file's PID to a new claude process, that session's `SessionStart` corrects the content.

Adding any sweep to collect this residue would reintroduce exactly the cron-like mechanism this PR removes, to reclaim kilobytes. It is left uncollected by design.

## Files

| File | Change |
|---|---|
| `claude/.claude/hooks/cleanup-session-id.sh` | New `SessionEnd` hook (mode 755) — content-match-guarded destructor |
| `claude/.claude/settings.json` | New matcher-less `SessionEnd` registration |
| `claude/.claude/hooks/tests/test_cleanup_session_id.py` | 7 cases via the capture→cleanup pairing idiom — both hooks resolve the same PID from the same `ps` walk, so the test never has to predict the PID |
| `docs/hooks.md` | New `cleanup-session-id.sh` bullet; extended `capture-session-id.sh` bullet to note the destructor pair |
| `README.md` | New hook-table row |

## Test plan

Automated: `pytest claude/.claude/` (1003 passed) + `ruff check claude/.claude/` (clean).

Manual smoke test (load-bearing, with a populated `~/.claude/sessions/`):

- [ ] Start an interactive session; confirm `sessions/<pid>` is written. Exit via Ctrl-D; confirm that file is removed and `<pid>.json` sidecars survive.
- [ ] Run a non-interactive `claude -p '...'` one-shot; confirm `SessionEnd` fires for it (`prompt_input_exit`/`other`) and its file is removed. **Contingency if it does not fire:** the change still ships as a partial win for interactive and resumed sessions; the "Residual leak" framing above would widen to include one-shots at one-shot rate. Not grounds to reintroduce a sweep.
- [ ] `/clear` mid-session; confirm the live session's `sessions/<pid>` file still exists afterward (content-match guard kept it).
- [ ] Resume a session; confirm the pre-resume file is gone and exactly one fresh file exists.
- [ ] Spawn a subagent, then end the session; confirm the file is still deleted (validates the `SubagentStart`-session-id equality assumption).


jcdendrite (commented):
lgtm

jcdendrite:
**[Claude Code]** Smoke test results (run 2026-05-19):

- [x] Test 1 — interactive exit removes `sessions/<pid>`: **PASS**, lookup file cleanly removed at Ctrl-D.
- [ ] Test 2 — `claude -p` one-shot SessionEnd fires: **FAIL** — `claude -p 'echo hi'` left one bare-PID file. `SessionEnd` does not fire for `claude -p` one-shots. This is the design's anticipated residual; triggers §5 contingency (follow-up PR widening the residual-leak framing in the hook header, `docs/hooks.md`, and `README.md`).
- [x] Test 3 — `/clear` content-match guard preserves successor: **PASS**, end-state check verified.
- [x] Test 4 — resume produces exactly one fresh file: **PASS** — initial session cleaned up at exit (4a); resumed session got exactly one fresh PID file (4b).
- [x] Test 5 — subagent + session end leaves no orphan: **PASS**, file removed at session exit.

Baseline `sessions/` count before: 917. After: 917 (net delta zero). Test 2's leaked PID file was directly observed in-run but was no longer present at the post-test snapshot — likely via PID reuse overwriting it with a new session's id, consistent with the hook's "self-heals on PID reuse" framing. No live-PID concurrent-session noise in any of the snapshots.

**Side observation (out of scope for this PR):** `~/.claude/sessions/<pid>.json` sidecars (264 bytes, JSON with `{pid, sessionId, cwd, startedAt, procStart, version, peerProtocol, kind}`) are written by Claude Code itself, not by our hooks (`capture-session-id.sh` only writes the bare-PID file). 11 currently exist in `~/.claude/sessions/`, all dated today, suggesting a recent platform-side feature. Whether they have their own destructor is unverified; if not, they accumulate at the same rate as the bare-PID files this hook now manages. Worth a follow-up if the sidecars also lack a lifecycle pair.

PR #276, opened 2026-05-19
Clarify: a lock you hold never blocks a subagent you dispatch

jcdendrite:
## Summary

Adds one paragraph to the "Heavy command output" subsection of global
`CLAUDE.md` stating that a lock held by a parent session never blocks a
dispatched subagent, because a subagent (Task/Agent tool) inherits the
parent's `session_id`. To any hook or single-instance-resource guard keyed
on session identity, a dispatched `check-runner` is the *same* session.

## Motivating incident

An Opus orchestrator holding a project-local Supabase test lock reasoned
that a `check-runner` subagent would be "a separate concurrent session"
and therefore blocked, so it ran `npm run verify` inline and burned
expensive parent context. The hook was correct — subagents inherit the
parent's `session_id` and pass the same-session claim check — but nothing
in the delegation doctrine stated the session-identity fact, so the
faulty inference formed. This paragraph closes that gap at the root, in
the section that already mandates dispatching `check-runner` for heavy
output.

## Companion PR

The project-specific consequence is carried in the consuming private
project's own CLAUDE.md and test-lock hook header.

## Test plan

- [ ] Confirm the new paragraph reads as intended in rendered GitHub markdown.
- [ ] Confirm cross-reference "HOOK_BLOCK verdict handled just below" still resolves after merge (next paragraph is unchanged).


jcdendrite (commented):
lgtm

PR #277, opened 2026-05-19
fix(skills): correct uncited length-target claims in ai-instruction-and-memory-files and skill-review

jcdendrite:
## Summary

Citation correctness fix to two skills' length-target sections.

**`ai-instruction-and-memory-files` §2** rewritten:
- Drops "1000+ words correlates negatively with compliance" — research traced this only to an unverifiable practitioner blog citing an uncited GitHub study. No Anthropic source uses a word-count threshold.
- Drops "~70% prose compliance / 100% hooks compliance" — neither figure appears in any cited source.
- The official **"under 200 lines per CLAUDE.md file"** rule is verbatim Anthropic from [code.claude.com/docs/en/memory](https://code.claude.com/docs/en/memory) — kept and now cited inline with a verbatim quote fragment.
- Adds explicit "line count alone undercounts density" caveat with the directive to "place critical rules near start or end" (preserves the "lost in the middle" arming from the original).
- The advisory/deterministic distinction is sourced verbatim to [Best Practices](https://code.claude.com/docs/en/best-practices): hooks "guarantee the action happens."

**Sibling fixes** in the same skill (§3 example, §3 don't-duplicate bullet, §5 routing cell, and two stale "§3 compliance asymmetry" back-pointer labels) for consistency.

**Cross-skill sibling fix** in `plugins/skill-review/skills/skill-review/SKILL.md` §4 — same uncited "~70%/100%" pair removed with parallel framing.

**REFERENCES.md** tier-grouped (Official Anthropic vs Practitioner) with per-source annotation of what each backs. Chroma URL canonicalized.

## Why

A prior session looked at a 106-line / 2,519-word stowed CLAUDE.md and called it "fine, under 200 lines," invoking the now-removed "1000+ words" claim to half-wave-off the density concern. Two problems:
- The word-count claim has no primary source (research traced it to a practitioner blog citing an uncited GitHub study).
- Line count alone undercounts density when each line is a long paragraph — the official docs themselves prefer "headers and bullets" over "dense paragraphs."

§2 now states "line count alone undercounts density" explicitly and points to the per-line behavior test as the binding density control.

## Plugin install

After merge, refresh skill-review in consuming repos with:

```
claude plugin install skill-review@claude-config --scope project
```

## Pre-existing observation (informational, out of scope)

[...]

PR #278, opened 2026-05-19
refactor(claude-md): extract delegation triad to subagent-delegation skill

jcdendrite:
## Summary

Reduces the always-loaded `claude/.claude/CLAUDE.md` (stowed to `~/.claude/CLAUDE.md`, applies every session for all stow users) from **109 lines / 2,728 words** to **73 lines / 1,748 words** — a **36% word reduction** — by extracting one coherent body of rules to a new auto-triggering skill.

## What moved (and what didn't)

**Extracted to new skill** at `claude/.claude/skills/subagent-delegation/SKILL.md` (164 lines, ~1,225 words):
- Three Working Style subsections — *Delegation: hold conclusions, not transcripts* / *Heavy command output* / *Codebase discovery* — combined into one skill with a coherent Step 1 (two-test gate) + Step 2 (pick the right subagent: `check-runner`, `Explore`/`general-purpose`, `code-writer`, or everything-else `general-purpose`).
- Every imperative directive, escape-hatch exception, forbidden-action callout, and verification command from the originals is preserved (the behavioral-equivalence audit in `/skill-review` walked them sentence-by-sentence). The `lock-doesn't-block-subagent` paragraph (session_id inheritance), all three `BLOCKED` block_type branches, and the test-count-via-grep recipe all survived intact.
- The on-demand body inlines its own rationale rather than back-referencing always-loaded files: `docs/auto-mode.md` and `~/.claude/CLAUDE.md` pointers were removed and replaced with the relevant single-sentence rationale at the site that needs it.

**Kept in CLAUDE.md:**
- "Use descriptive variable and function names. No generic names." — short and load-bearing; catches name issues at write time, not via post-hoc review.
- **New** "Default-consider delegation" bullet — prompts Claude before each Bash command / broad search / check suite / Read-heavy probe to ask whether the *objective* belongs in a subagent. Always-loaded; cross-references the skill body for criteria. Without it, the skill could silent-fail to trigger when Claude doesn't perceive the dispatch decision in the first place.
- 60%-context bullet compressed: kept the directive (threshold + suggest `/handoff` + don't run it yourself); dropped `/handoff`'s own operational detail.

## Why now

The previous PR (#277) sharpened the rule that **line count alone undercounts density** in a CLAUDE.md written as long-paragraph bullets. 109 lines / 2,728 words was exactly that pattern — under the 200-line cap, but paragraph-dense. The delegation triad was a clean candidate because its rules fire in a *detectable* shape (about to dispatch a check suite, about to do broad search, about to run a 2nd/3rd Bash command toward one question), making it skill-eligible per `ai-instruction-and-memory-files` §4 — "The rule applies only when a specific skill is running ⇒ edit that skill's SKILL.md, not CLAUDE.md."

## Trade-off worth knowing

Always-loaded prose fires every session. On-demand skills fire when the trigger detects its condition. The new CLAUDE.md "Default-consider delegation" bullet is an always-loaded bridge — it primes Claude to ask the dispatch question before running each command, and the on-demand skill body (loaded once the question is asked) carries the two-test gate, per-subagent routing, and inline-vs-delegate rules. This is the minimum always-loaded surface needed to ensure the on-demand layer actually triggers; the full criteria still live in the skill body. If the auto-trigger turns out to under-fire in practice, the fix is to tighten the TRIGGER text or expand the always-loaded bridge — not to restore the full triad to CLAUDE.md.

## Test plan

- [x] `/skill-review` — clean (behavioral-equivalence audit walked all compression diffs row by row, every row Y)
- [x] `/code-review` — clean (no findings on cumulative diff; no ripple specialists needed for documentation-only changes)
- [x] `/ai-instruction-and-memory-files` — clean on CLAUDE.md edits
- [x] `pytest claude/.claude/` — **1,005 passed**
- [x] Mechanical skill-length gate — passes (164 lines for the new skill; 73 lines for CLAUDE.md)


claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -0,0 +1,159 @@
+---
+name: subagent-delegation
+description: >
+  When and how to dispatch work to a subagent — preserves parent session
```
jcdendrite: Description text is quite long. Keep in mind that description text adds to skill context which can lead to skill descriptions be cut off. 

Trim parts of the description that aren't load-bearing and put them in REFERENCES.md instead for the skill. For example, "- preserves parent session context" - that seems important for REFERENCES.md as to why we have the skill (or maybe just the description of the skill in docs/skills.md, which needs to exist btw you didn't add it.)

claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+  (pytest/lint/typecheck/build) or any full-project verification command;
+  about to do broad codebase search (multi-file grep, glob sweeps,
+  exploratory reads spanning more than ~3 queries); about to run the 2nd
+  or 3rd Bash command toward the same question; reporting test counts
+  from a check-runner verdict. DO NOT TRIGGER when: single targeted
+  lookup (one grep for a known symbol, one Read of a known path);
+  comprehension reads where file content feeds your own writing, review,
+  or design; Edit/Write sequences; output you must reason over line by
+  line (a failure you're debugging, a diff you're designing against).
+---
```
jcdendrite: I feel like these trigger conditions are at once overly extensive and overly specific. The trigger and do not trigger conditions need to be reliable and project-agnostic. Hard-coding pytest here for example, gives the wrong weight to python-based projects. Separate the heuristics (which should live in the body) from the trigger points. Really carefully review the skill-review skill again.

And actuall, even though claude's marketplace plug-in skill skill-creator is disabled, you may just want to read the prose of that skill and apply it here.

claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+  design.
+- `Edit`/`Write` sequences (judgment-dense, not scratch).
+- Output you must reason over line by line (a failure you are
+  debugging, a diff you must design against).
+
+**No permission cost.** A subagent runs under the parent's permission
+mode. Under auto mode, its read-only diagnostics are evaluated by the
+same classifier that clears the parent's — delegating read-heavy work
+adds no permission prompts and needs no `permissions.allow` entries.
+See `docs/auto-mode.md` in the claude-config repo for the classifier's
```
jcdendrite: Is it a good idea to have a skill or CLAUDE.md reference docs? I thought docs were only for human eyes. I thought referencing docs was an anti-pattern but maybe I misremembered.

claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+behavior, and Model Routing in `~/.claude/CLAUDE.md` for when to pass
```
jcdendrite: Why would you need to back-reference CLAUDE.md here when that's pre-loaded for every conversation?

claude/.claude/CLAUDE.md
```diff
@@ -18,45 +18,7 @@
 - Always prefer minimal, targeted changes. Do not refactor entire files or expand scope beyond what was asked. If you see an opportunity for a broader improvement, mention it separately — do not bundle it in.
 - **Compounding defensive layers are a wrong-foundation tell.** When a design accumulates stacked defenses on a single mechanism — each new layer closing a gap that the prior layer's existence created — or starts citing its own prior findings, step back and ask whether a foundational change would dissolve them. Do not keep adding hardening. The right primitive usually has a simple shape; compounding complexity is a signal to question the foundation, not to defend it more carefully.
 - Before assuming anything about the environment, stack, or project conventions, check first. Read the actual config files rather than guessing defaults.
-- Use descriptive variable and function names. No generic names.
```
jcdendrite, line 21: I think the `Use descriptive variable and function names` is actually quite useful. Remember that code-review only catches things after the fact and it's best to just get things right the first time. This is a short sentence that carries a lot of weight.

claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+`model: opus` vs `model: sonnet` to `general-purpose`.
+
+## Step 2 — Pick the right subagent
```
jcdendrite, line 58: Also the code-writer agent merged since you were working on this, so you should include that agent as a reference as well.

claude/.claude/CLAUDE.md
```diff
@@ -18,45 +18,7 @@
...
-
-This does not apply to *comprehension* reads: when you need a file's content in your own reasoning — to write or modify it, review it, or design against it — read it directly. The split is locate-and-report (delegable) vs. read-and-reason (not).
+- When a session crosses ~60% context usage (per statusline `.context_window.used_percentage`) AND the current task is incomplete, suggest the user run `/handoff` — it's user-invoked; don't run it yourself.
```
jcdendrite, line 23: There's a flaw in removing reference to subagent delegation in CLAUDE.md - it now has no guidance on WHEN to delegate. The new subagent-delegation skill only triggers when the parent agent thinks to delegate. What if it thinks it's not necessary? I think there need to be more explicit instructions in CLAUDE.md regarding delegation. It needs to be always thinking if it should delegate a command/Read/Write/Bash/etc to a subagent or not.

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Trimmed the description from ~110 words to ~75 — dropped the "preserves parent session context" rationale (already in body opening paragraph), the pytest/lint/typecheck parentheticals, and the adjacent-condition examples in DO NOT TRIGGER (still in body's "Stays inline" bullets). Added a `subagent-delegation` entry to `docs/skills.md` so the skill's purpose is documented for the skill catalog. The why-the-skill-exists rationale lives in the body opening + docs/skills.md entry — no REFERENCES.md created (no edit-time source quotes or framework notes to capture for this skill yet).

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Dropped `pytest/lint/typecheck/build` from the description — those were stack-leaking parentheticals weighting toward Python projects. The trigger now says "running a full check suite or other full-project verification command" (action verb, stack-agnostic). Heuristics ("≥3 queries", "2nd/3rd command toward one question") are now elaboration in the body, not embedded in the trigger string. The body keeps a `pytest claude/.claude/`, `ruff check claude/.claude/` example for the "enumerate exact command strings" rule consistent with the existing `ready-for-review/SKILL.md` pattern (line 48) — left alone to avoid drifting from the established skill voice across the repo.

Read skill-creator while drafting; the relevant guidance ("description is the primary triggering mechanism", "make descriptions a little pushy") was applied alongside skill-review §2 (specificity sources: globs > action verbs > context cues).

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Removed the `docs/auto-mode.md` pointer from the skill entirely. Your recall is right — `docs/` is for human readers (REFERENCES.md is the edit-time co-located reference for a skill, not docs/). The relevant rule ("no permission cost under auto mode for read-only diagnostics") is self-contained in the surviving paragraph; Claude doesn't need to fetch the doc to act on it.

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Removed both `~/.claude/CLAUDE.md` back-references from the skill body (the Safety citation on the exact-match rule, and the Model Routing citation in the Codebase-discovery section). Inlined the relevant single-sentence rationale at the site that needs it:

- Codebase discovery: "Always pass an explicit `model` on `general-purpose` — it has no model of its own and inherits the parent's, which under auto mode may be Opus."
- Everything-else: "with an explicit `model: sonnet` (a no-op when the parent is already Sonnet; keeps the dispatched work off Opus when the parent is not)."

The skill now stands on its own; CLAUDE.md's Model Routing section is always-loaded and applies regardless.

claude/.claude/CLAUDE.md
jcdendrite: **[Claude Code]** Fixed. Restored "Use descriptive variable and function names. No generic names." to CLAUDE.md Working Style. Agreed — short, load-bearing, and catches name issues at write time rather than via post-hoc `/code-review`. That's exactly the case for keeping an always-loaded rule even if a downstream review could also flag it.

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Added an "Implementation work → `code-writer`" section to Step 2 of the skill (between Codebase discovery and Everything-else): "When delegating code-writing — feature code, fixes, refactors, migrations, schema, scripts — dispatch the `code-writer` subagent, not `general-purpose`. It carries `model: sonnet` and self-reviews its own diff against staff-engineer reviewer angles before returning, catching review-finding-class defects in its own context instead of as a parent round-trip."

Also added "delegating implementation work" to the TRIGGER list in the skill description, and included `code-writer` in the docs/skills.md entry for the skill.

claude/.claude/CLAUDE.md
jcdendrite: **[Claude Code]** Fixed with modification. Added a new "Default-consider delegation" bullet to CLAUDE.md Working Style — *not* the full triad, but a single always-loaded prompt:

> **Default-consider delegation.** Before running a Bash command, starting a broad search, initiating a check suite, or beginning a Read-heavy probe, ask whether the *objective* (not the individual command) belongs in a subagent. The parent's context is re-read every turn, so verbose tool output left in it is paid for repeatedly. See the `subagent-delegation` skill for the two-test gate, which subagent fits which case, and what stays inline.

This is the always-loaded bridge that ensures Claude *asks* the dispatch question on every Bash/search/check/Read-heavy probe — the trigger that makes the on-demand skill actually fire. The skill body then carries the two-test gate and the per-subagent routing. Refreshed the PR description's "Trade-off worth knowing" section to document the always-loaded-bridge / on-demand-body split.

`disposition: fixed-with-modification` | `rationale: added a single always-loaded prompt bullet instead of restoring the full delegation triad — keeps the heavy detail on-demand in the skill while ensuring the trigger actually fires`

jcdendrite (commented):
lgtm

PR #283, opened 2026-05-20
feat(lovable-cloud-migration-sync): add edge-runtime mount precondition

jcdendrite:
## Summary

- Adds an explicit precondition to Step 5 of `lovable-cloud-migration-sync`: confirm the local Supabase edge runtime container is bind-mounted to the current worktree's `supabase/functions` (and that the path exists on disk) before dispatching `check-runner`.
- Provides the `docker inspect` check command and `supabase stop && supabase start` as the remediation when the mount points elsewhere or the path is gone.
- Documents why this belongs in the parent and not in `check-runner`: the subagent is stateless and cannot detect a stale bind mount; the failure mode (a wave of 503 `BOOT_ERROR` responses) looks like a regression and is easy to misattribute to pre-existing failures on `main`.

## Scope

Limited to Step 5 of this one skill. The mount-discipline rule is specific to Lovable Cloud's per-worktree Supabase wrapper; the plugin's shared docs deliberately do not absorb it.

## Test plan

- [ ] Read the diff and confirm Step 5's precondition reads cleanly without prior-PR context.
- [ ] Confirm no project-identifying content survived from the briefing.

### DIFF
--- a/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
+++ b/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
@@ -92,6 +92,30 @@ with the user.
 
 ### 5. Verify locally before deleting
 
+**Precondition: confirm the edge runtime is mounted on the current worktree.**
+The local Supabase edge runtime container is bind-mounted to the
+`supabase/functions` directory of whichever worktree last ran `supabase start`.
+If the mount points at a different worktree — or the mounted path was removed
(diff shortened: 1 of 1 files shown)


jcdendrite:
Closing — wrong fix at wrong layer. Edge-runtime mount discipline belongs at the check-runner invocation layer (project repo), not in this skill body. Replacement PR will remove the hard-coded `supabase db reset` from Step 5 and delegate to the project's verify entry point instead.

PR #286, opened 2026-05-20
test(skills): enforce description cap and strict YAML frontmatter

jcdendrite:
## Summary

- Adds `TestModelInvokableDescriptionLength` — parametrized pytest catching skills whose rendered `description` + `when_to_use` would exceed the harness's 1,536-char per-skill listing cap. The harness silently truncates above that, losing the tail of the description (TRIGGER / DO NOT TRIGGER blocks if they sit at the end). Cap value cited to https://code.claude.com/docs/en/skills.md.
- Adds `TestModelInvokableSkillFrontmatterIsStrictYaml` — parametrized pytest rejecting inline-unquoted frontmatter values containing `: `. The Claude Code harness's parser currently accepts these (e.g. `description: TRIGGER when: ...`), but strict YAML parsers don't — so any future parser-strictness change would silently truncate descriptions at the first `: `, dropping TRIGGER blocks.
- Adds a `_skill_frontmatter()` helper that parses (rather than raw-scans) YAML so block-folded scalars (`description: >`) and quoted strings are measured at their rendered length, not their source-text length.
- Switches `code-review` and `respond-pr` SKILL.md `description:` values to double-quoted form to satisfy the new strict-YAML check (those two were the only skills relying on lenient parsing; verified the harness was reading them in full before the change, so this is fragility removal, not bug fix).
- Adds `pyyaml==6.*` to the CI install line so the new tests' `import yaml` resolves under the workflow's pip environment.

## Test plan

- [x] `pytest claude/.claude/` passes
- [x] `ruff check claude/.claude/` passes

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,7 +88,7 @@ jobs:
 
       - name: Install dependencies
         if: steps.detect.outputs.changed == 'true'
-        run: pip install 'pytest==8.*' 'ruff==0.6.*'
+        run: pip install 'pytest==8.*' 'ruff==0.6.*' 'pyyaml==6.*'
 
       - name: Run tests
         if: steps.detect.outputs.changed == 'true'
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

PR #287, opened 2026-05-20
feat(hooks): wire check-runner mutation guard via settings.json

jcdendrite:
## Summary

Wires the previously-unwired `check-runner-bash-guard.sh` via `settings.json` as a `PreToolUse(Bash)` hook, with a payload-discriminator that scopes enforcement to the `check-runner` subagent only. Layers a vendor-name-free global deny set with a per-project extension file for stack-specific vendor-named patterns.

## What shipped

**Hook (`claude/.claude/hooks/check-runner-bash-guard.sh`)**
- Defense-in-depth `tool_name == "Bash"` filter — a misconfigured matcher in `settings.json` can't make the hook read fields off the wrong payload shape.
- `.agent_type == "check-runner"` discriminator — parent and non-check-runner subagent calls pass through unchanged.
- Global generic-shape deny patterns (no vendor names): database-CLI verb pairs (`<word> db (reset|push|migrate|seed)`), force-pushes (`git push --force|-f`), hard resets (`git reset --hard`), categorical `rm -rf` of `/`, `$HOME`, `~`.
- Pre-existing `git` read-only allowlist (from `_lib.sh`) preserved as a third layer.
- Project-layer extension: `<cwd>/.claude/check-runner-deny-patterns.txt` (regex per line, comments allowed). Read at decision time; cited by file + line number on match.

**Settings (`claude/.claude/settings.json`)**
- New `PreToolUse(Bash)` entry pointing at the hook, ordered first under the matcher.

**Template (`claude/.claude/hooks/check-runner-deny-patterns.template.txt`)**
- Reference template with commented exemplars for package installs, container lifecycle, cloud-provider CLIs, and seed scripts. Copy to `<project>/.claude/check-runner-deny-patterns.txt` and uncomment what applies.

**Tests (`claude/.claude/hooks/tests/test_check_runner_bash_guard.py`, `claude/.claude/tests/helpers.py`)**
- Discriminator allow paths (parent + four other subagent types).
- Global deny shapes (9 database-CLI fixtures, force-push, hard-reset, categorical `rm -rf`).
- Positive `rm -rf` subpath allow (`/tmp/...`, `./node_modules`, `$TMPDIR/...`).
- Project-layer behaviors: match + citation, file present but unmatched, file absent (global-only), unreadable file (open-fail), malformed regex on one line (skip and continue), blank/comment lines skipped.
- Fail-closed on malformed JSON; empty/missing command field allowed.
- `bash_input` helper gains optional `agent_type` parameter.

[...]

PR #280, opened 2026-05-19
feat: add proactive branch-divergence detection

jcdendrite:
## Summary

Add proactive branch-divergence detection across the contributor workflow, so merge conflicts surface — and are resolved — at the earliest sensible point instead of only after `git push` or at PR-review time.

Three checkpoints, one detection primitive (`git merge-tree --write-tree`), single resolution owner (existing `/git-feature-branch-sync`):

- **SessionStart advisory hook** (`check-branch-divergence.sh`) — warns when the current branch has fallen behind `origin/<default>`. Skip-silent on not-in-repo / default branch / detached HEAD / no-origin / unresolved `origin/HEAD`; portable `timeout` → `gtimeout` → skip-fetch wrapper; prompt-suppressed fetch (`GIT_TERMINAL_PROMPT=0`, empty `SSH_ASKPASS` / `GIT_ASKPASS`); stale-ref fallback when fetch fails. Advisory-only — never blocks.
- **`/ready-for-review` Step 1** — divergence precondition; routes to `/git-feature-branch-sync` when behind > 0 and re-runs verification on the synced tree. Step 8's completion marker records the post-resync HEAD SHA so it matches what the push-gate hook checks.
- **`/respond-pr` new Step 3** — pre-reply divergence check; aborts before applying any `FIXED` change onto a stale tree (replies acknowledging fixes would point at a commit the reviewer cannot cleanly resolve).
- **Canonical detection recipe** in `git-feature-branch-sync/SKILL.md` so all three sites share one primitive.

Settings change: new `SessionStart` hook entry with matcher `startup` only (intentionally narrower than `session-marker-dashboard.sh`'s `startup|clear|compact` — divergence is on-disk state and doesn't need re-emission on `/compact` or `/clear`).

## Test plan

- [x] `pytest claude/.claude/` — 1014 passed, including 11 new hook cases (a–j): not-in-repo, default branch, no-origin, behind=0, behind>0 clean trial-merge, behind>0 conflict, fetch-timeout with cached behind=0 / behind>0, `origin/HEAD` unset, detached HEAD, neither `timeout` nor `gtimeout` on `$PATH`.
- [x] `ruff check claude/.claude/` — clean.
- [x] Hook smoke-tested in-tree: emits nothing when branch is in sync.
- [ ] Reviewer: confirm hook fires advisory on a deliberately-diverged worktree at the next SessionStart.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #288, opened 2026-05-20
refactor(claude-md): add prose-placement bullet under Code Comments umbrella

jcdendrite:
## Summary

Restructures the `Code Comments and Durable Documentation` section of
`claude/.claude/CLAUDE.md` into `Code Comments, Documentation, and Prose`
with two subsections:

- **Where to put it** (new) — adds a single bullet on prose placement:
  *Place prose where its reader and altitude match*. Includes the
  failure-mode examples (README altitude mismatch, agent-spec FYI
  dumps, doc back-references from skills) and the write-time test
  *"does it belong here?"*
- **When to write it and what to include** (existing three bullets,
  unchanged: no PR-defined terminology, no `used to be X` framing,
  self-test for survivability).

The existing section was the natural umbrella because durability and
placement are sibling write-time disciplines. Placement comes first:
if you can't decide where, durability is moot.

## Why

Recurring class of PR feedback where prose lands in the wrong file —
e.g. README walkthrough altitude issues (#280), agent-spec
implementation FYI (#287), doc back-references inside skill bodies
(#278). No existing CLAUDE.md rule covered prose **placement**; this
adds the missing write-time discipline so Claude self-corrects before
review-time, rather than relying on the reviewer to catch it.

The new bullet is ~80 words, comparable to other CLAUDE.md bullets
(`Compounding defensive layers` ~75 words, `Default-suspect
over-powered primitives` ~95 words). CLAUDE.md grows 74 → 80 lines —
well under the 200-line cap.

## Test plan

- [x] `pytest claude/.claude/` passes (verified locally — green)
- [x] `ruff check claude/.claude/` passes (verified locally — green)
- [ ] Reviewer agrees `Where to put it` belongs before
  `When to write it and what to include` (placement is foundational
  to durability)
- [ ] Reviewer agrees the umbrella heading
  `Code Comments, Documentation, and Prose` accurately covers both
  subsections (the new bullet applies to all prose, not just durable
  docs)


jcdendrite (commented):
lgtm

PR #289, opened 2026-05-20
feat(brief): add /brief skill for cold-start task briefings

jcdendrite:
## Summary

Adds `/brief`, a user-invoked skill that writes a cold-start task briefing at `/tmp/<slug>-task.md`. Sibling to `/handoff` — different shape, different reader.

- **`/handoff`** captures mid-flight session state for a future continuation of the same task.
- **`/brief`** captures a well-scoped piece of work the current session is *not* going to do (abandoned PR, settled-scope ticket, follow-up surfaced mid-session) so a brand-new session can pick it up cold.

Skill body uses a §1–§7 template (Goal, What it does, Anchors, Current state, Decisions to make, Steps to ship, Out of scope), includes two "When to use" disambiguation sections (vs `/handoff`, vs `/plan-it`), cold-start readability rules (no PR-defined terminology, no "prior session did…" framing, no implicit context), a pre-write checklist, and a generic worked example.

The `/plan-it` disambiguation was added after PR review surfaced that the §1 Goal / §2 What it does sections read like plan sections. The distinction: `/plan-it` is the design phase (committed plan, reviewed by `/plan-review`); `/brief` is the operational handoff for already-scoped work (ephemeral `/tmp` file, no design step). Closes with a §5-shape diagnostic.

Frontmatter sets `disable-model-invocation: true` — user-invoked only, no cost against the auto-loaded skill-listing budget. Matches the pattern used by `/handoff`, `/read-docx-comments`, and `/transcript-analysis`.

`docs/skills.md` gets a one-line entry alongside `/handoff`.

## CLAUDE.md reminder — considered, rejected

Considered adding a parallel reminder beside the existing `/handoff` line at `claude/.claude/CLAUDE.md:23` (which fires at ~60% context + incomplete task). Rejected: `/handoff`'s reminder is tied to a precise, externally-detectable signal (context %); `/brief` has no equivalent. "Scope surfaces that doesn't fit current plan" is judgment-based, and the existing behavioral memory already covers that trigger. A vague CLAUDE.md line would duplicate the memory's role without sharpening the trigger.

## Test plan

- [ ] `/brief` appears in the slash menu after `git pull` (stowed via `claude/.claude/`)
- [ ] Invoking `/brief` writes a usable `/tmp/<slug>-task.md` skeleton on a real abandoned-PR scenario
- [ ] `/handoff`, `/brief`, and `/plan-it` read as clearly differentiated in both the skill bodies and `docs/skills.md`


claude/.claude/skills/brief/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+---
+
+Write a cold-start task briefing at `/tmp/<descriptive-slug>-task.md` using the structure below. A fresh session — with no memory of the originating conversation — must be able to read the file and execute the work.
+
+## When to use this vs `/handoff`
+
+- **`/handoff`** captures mid-flight session state: "I'm partway through, here's exactly where, here's how to resume." The reader is a future continuation of the same task.
+- **`/brief`** captures a well-scoped piece of work the current session is *not* going to do: an abandoned PR nobody is driving, a settled-scope ticket, a follow-up surfaced mid-session that doesn't belong in the current plan. The reader is a brand-new session picking it up cold.
+
+If you are still mid-task and running low on context, use `/handoff`. If the work is known, abandoned, and someone else (or a future you) should pick it up cold, use `/brief`.
```
jcdendrite, line 14: Think about your audience. Are these instructions helpful for AI agents? For humans? Or for both? Explain.

claude/.claude/skills/brief/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+Before writing the file, verify:
+- Every section §1–§7 above is populated
+- No placeholder text ("TBD", "TODO", "fill in later") in any section
+- §4 Current state is objective — facts a fresh session could verify, not editorial
+- §5 Decisions to make lists open questions, not foregone conclusions framed as questions
+- §6 Steps to ship name concrete commands or file edits, not vague verbs
+- §7 Out of scope names at least one specific temptation to leave alone
+- File contains no references to the originating conversation, prior sessions, or unstated context
+
+## Worked example
```
jcdendrite, line 72: Might want to distinguish when to use /brief and when to use /plan-it 

The example you've described sort of sounds like a plan. Unless the difference is that a plan is complicated enough that it warrants a review? It's not mechanical?

claude/.claude/skills/brief/SKILL.md
jcdendrite: **[Claude Code]** The skill body is loaded into AI-agent context when `/brief` is invoked, so the immediate reader is the AI executing the skill. The output (the `/tmp/<slug>-task.md` file) is also read by a future AI session (cold-start). Humans request `/brief` and read the eventual PR/result, but never the skill body or the briefing file itself.

That makes most of the body (sections §1–§7, Cold-start readability rules, Slug naming, Pre-write checklist, Worked example) pure AI-execution instructions. The two "When to use this vs ..." sections also serve the AI primarily — they let it judge when to recommend `/brief` (vs `/handoff` or `/plan-it`) to the user — though the framing reads cleanly enough that a human deciding which slash command to invoke benefits too.

No code change for this comment; the audience model matches every other user-invoked skill in `claude/.claude/skills/` (`/handoff`, `/read-docx-comments`, `/transcript-analysis`), which all carry `disable-model-invocation: true` and execute as AI-directed instructions.

claude/.claude/skills/brief/SKILL.md
jcdendrite: **[Claude Code]** Good catch — the §1 Goal / §2 What it does sections do read like plan sections. Your hypothesis (review = the line) is close but not quite the boundary; the actual distinction is lifecycle position.

Added a parallel "When to use this vs `/plan-it`" section to the skill body in the latest commit on this branch, mirroring the existing "vs `/handoff`" section:

- **`/plan-it`** is the design phase: approach is unsettled, exploration and clarifying questions are needed, output (`.claude/plans/<slug>.md`) is reviewed by `/plan-review`, committed.
- **`/brief`** is the operational handoff: scope is settled (or mechanical), no design step remains, output (`/tmp/<slug>-task.md`) is ephemeral.

The reviewability you noted is a downstream effect of the lifecycle split — `/plan-it` exists to produce something `/plan-review` will scrutinize before code is written; `/brief` exists to hand off work whose design is already done.

Closed the section with a §5-shape diagnostic: if `§5 Decisions to make` is filling with open design questions rather than specific items awaiting an answer, write a plan instead. That gives the writer (AI executing `/brief`) a concrete signal to route to `/plan-it` mid-draft.

`disposition: fixed-as-requested` (the reviewer asked for a distinction; one is now in the skill body).

jcdendrite (commented):
lgtm

PR #290, opened 2026-05-20
feat(skill-management): rename plugin and bundle structural SKILL.md validator

jcdendrite:
## Summary

Renames the `skill-review` plugin to `skill-management` and ships PR #286's structural SKILL.md checks as a portable validator that runs on every staged SKILL.md before the behavioral-equivalence marker check. The skill *inside* the plugin keeps its `skill-review` name to preserve byte-stable references in `marker.sh`, `settings.json`, hook-alignment fixtures, and `code-review/SKILL.md`.

### What ships

- **New:** `plugins/skill-management/scripts/validate_skill_structure.py` — structural validator that catches frontmatter which would silently truncate from the harness's skill listing or fail strict-YAML parsing. Callable as library (`validate(skill_path: Path) -> list[str]`) and CLI (`python3 validate_skill_structure.py <paths...>`). The harness cap is encoded as a named constant with primary-source citation in the script; user-facing surfaces describe the concept rather than hardcoding the value. Path-based interface — does not depend on claude-config's own skill discovery, so downstream installers can use it directly.
- **Hook integration:** `require-skill-review.sh` now runs the validator on staged SKILL.md paths via `git show :<path>` into a tmp tree (not the working-tree file), so a stage-broken-then-fix-locally flow cannot bypass the gate.
- **Plugin packaging:** `plugins/skill-review/` → `plugins/skill-management/`, `plugin.json` `name` and `version` updated (1.0.3 → 2.0.0, breaking).
- **Test suite refactor:** `claude/.claude/skills/tests/test_skills.py` imports `validate()` from the validator via `pyproject.toml`'s `[tool.pytest.ini_options] pythonpath`. Single source of truth — the commit-gate hook and the test suite share the same module, no rule duplication and no `sys.path` manipulation.
- **Regression tests:** Two new tests in `claude/.claude/hooks/tests/test_require_skill_review.py` guard the staged-blob isolation in both directions (broken-staged-clean-working denies; clean-staged-broken-working allows).
- **Bundled skill trim:** `plugins/skill-management/skills/skill-review/SKILL.md` 201 → 198 lines, applying its own §4 behavior-test rules. Leaves headroom for future one-line adds before re-tripping the 200-line gate.
- **Contributor setup:** `./install.sh` now provisions `pyyaml` automatically — fresh clones can run `pytest claude/.claude/` without a separate manual step.

### Breaking change — downstream migration

```
claude plugin uninstall skill-review --scope project
```

Between `uninstall` and `install` there is a brief window with no SKILL.md gate active — do not commit SKILL.md changes during that window.

The hook requires `python3` on `$PATH` with `pyyaml` available to that interpreter. For consumers installing the plugin into another repo, run `python3 -m pip install pyyaml` once; the hook fails closed if `pyyaml` is unavailable to the system `python3`. (For contributors to claude-config itself, `./install.sh` provisions this automatically.)

## Test plan

- [x] `pytest claude/.claude/` — full skill + hook test suite passes
- [x] Migration command verified by user on the known downstream installer (out-of-band)
- [x] `/skill-review` slash command loads under new plugin packaging


docs/skills.md
```diff
@@ -82,7 +82,7 @@ Three skills that primarily apply to this repo's own workflow — editing `SKILL
 
 | Plugin | What it provides | When to install |
 |---|---|---|
-| `skill-review@claude-config` | Behavioral-equivalence audit for `SKILL.md` changes; gates `git commit` when staged changes include a `SKILL.md` | Repos that author their own `SKILL.md` files |
+| `skill-management@claude-config` | Structural validation (description + when_to_use ≤ 1536 chars, strict-YAML frontmatter) at `git commit` on every staged SKILL.md, plus behavioral-equivalence audit via `/skill-review` | Repos that author their own `SKILL.md` files |
```
jcdendrite: Don't hard-code 1536 everywhere - what if Claude Code changes that at some point? Just say this ensures the threshold for skill context isn't hit or something but more tersely and clearly.

.claude-plugin/marketplace.json
```diff
@@ -16,12 +16,12 @@
       "category": "productivity"
     },
     {
-      "name": "skill-review",
-      "description": "Behavioral-equivalence audit for SKILL.md changes. Gates git commit when staged changes include a SKILL.md. Install at project scope in repos that author SKILL.md files.",
+      "name": "skill-management",
+      "description": "SKILL.md quality toolkit: structural validation (description + when_to_use ≤ 1536 chars, strict-YAML frontmatter) run automatically at git commit on every staged SKILL.md, plus behavioral-equivalence audit via /skill-review. Install at project scope in repos that author SKILL.md files.",
```
jcdendrite: This description is messy. I think it reads more like a cryptic literal feature than a holistic purpose for the plugin.

claude/.claude/skills/tests/test_skills.py
```diff
@@ -29,21 +29,24 @@
...
+import sys
 from pathlib import Path
 
 import pytest
-import yaml
+
+# Import structural validation from the skill-management plugin so the test
+# suite and the commit-gate hook share a single implementation of the rules.
+# parents[4] is the repo root: test_skills.py → tests/ → skills/ → .claude/ → claude/ → root
+sys.path.insert(  # noqa: E402
```
jcdendrite: Is there really no reasonable alternative to skipping linting here?

claude/.claude/skills/tests/test_skills.py
```diff
@@ -29,21 +29,24 @@
...
+    0,
+    str(Path(__file__).resolve().parents[4] / "plugins/skill-management/scripts"),
+)
+from validate_skill_structure import validate  # noqa: E402
```
jcdendrite: Is there really no reasonable alternative to skipping linting here?

plugins/skill-management/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "skill-management",
+  "description": "SKILL.md quality toolkit: structural validation (description + when_to_use ≤ 1536 chars, strict-YAML frontmatter) run automatically at git commit on every staged SKILL.md, plus behavioral-equivalence audit via /skill-review. Install at project scope in repos that author SKILL.md files.",
```
jcdendrite: See comment earlier

plugins/skill-management/scripts/validate_skill_structure.py
```diff
@@ -0,0 +1,93 @@
...
+import sys
+from pathlib import Path
+
+import yaml
+
+# Per https://code.claude.com/docs/en/skills.md: "the combined `description`
+# and `when_to_use` text is truncated at 1,536 characters in the skill
+# listing to reduce context usage." Configurable via maxSkillDescriptionChars
+# setting; 1,536 is the default the harness applies when no override is set.
+MAX_SKILL_DESCRIPTION_CHARS = 1536
```
jcdendrite, line 24: Addendum to my comment earlier about hard-coding this everywhere. I think this along with the research and link belong in the script itself (since it's directly relevant for changing logic here) and in the human-focused README or docs - whereever it belongs (although you still probably don't need that 1536 cap, just reference the direct Anthropic source and talk about the concept of the limit and what happens if it's exceeded).

plugins/skill-management/skills/skill-review/SKILL.md
```diff
@@ -34,9 +34,7 @@ Optional frontmatter:
   invoked by a parent via Glob+Skill-tool — `test-conventions`, `plan-review`,
   and `code-review` all use this pattern; auto-trigger is not needed.
 
-The harness loads only the description at startup; the body is fetched
-on demand. Description text is always-loaded context budget; body
-text is conditional. Frontmatter overspend hurts more than body overspend.
+Description text is always-loaded context budget; body text is conditional.
```
jcdendrite: Cleaner, but you've taken out the part that explains why this matters. "Frontmatter overspend hurts more than body overspend." Perhaps you can figure out how to more elegantly convey that with your terse update above.

README.md
```diff
@@ -173,7 +173,12 @@ This repo exposes a marketplace via `.claude-plugin/marketplace.json`. Each plug
...
 
 - **`lovable-cloud`** — Skills for Lovable Cloud projects: Project/Workspace Knowledge fields, edge-function auth model (ES256 gateway constraint, two-tier auth), and migration-sync workflow. `claude plugin install lovable-cloud@claude-config --scope project`
-- **`skill-review`** — Behavioral-equivalence audit for `SKILL.md` changes; gates `git commit` when staged changes include a `SKILL.md`. `claude plugin install skill-review@claude-config --scope project`
+- **`skill-management`** — SKILL.md quality toolkit: structural validation (description + when_to_use ≤ 1536 chars, strict-YAML frontmatter) run automatically at `git commit` on every staged SKILL.md, plus behavioral-equivalence audit via `/skill-review`. `claude plugin install skill-management@claude-config --scope project`
+
+  Requires `python3` and `pyyaml` on PATH — the commit-time hook fails closed if either is missing. Install via `pip install pyyaml` (or your environment's equivalent).
```
jcdendrite: Ooh, this is scope creep to the setup and python adds latency to hooks. Is this really the approach we want to do? Or is the bash equivalent workable? You may want to get staff platform engineer's and staff backend engineer's opinions here.

README.md
```diff
@@ -173,7 +173,12 @@ This repo exposes a marketplace via `.claude-plugin/marketplace.json`. Each plug
...
+
+  > **Migration from `skill-review`:** renamed from `skill-review` in v2.0.0. Downstream installers run `claude plugin uninstall skill-review && claude plugin install skill-management@claude-config --scope project` to migrate. Between uninstall and install there is a brief window with no SKILL.md gate active; do not commit SKILL.md changes during that window.
```
jcdendrite: This change does not belong in the README. It belongs in the CHANGELOG as you have already edited.

docs/skills.md
jcdendrite: **[Claude Code]** Fixed. Reworded the `docs/skills.md` row to describe the concept ("frontmatter that would silently truncate from the harness's skill listing or fail strict-YAML parsing") without the literal `1536`. The value stays as a named constant with Anthropic-doc citation only in `validate_skill_structure.py`, where it's load-bearing for the validator logic.

.claude-plugin/marketplace.json
jcdendrite: **[Claude Code]** Fixed. The new description leads with the holistic purpose ("Authoring guardrails for SKILL.md files") and frames the two capabilities by what they catch rather than as a literal feature list. The same text is mirrored in `plugin.json` (intentional per plan §53 — same plugin, two registry surfaces).

claude/.claude/skills/tests/test_skills.py
jcdendrite: **[Claude Code]** Fixed. Moved the import-path wiring out of `test_skills.py` entirely by adding `plugins/skill-management/scripts` to `pyproject.toml`'s `[tool.pytest.ini_options] pythonpath` array (which already had two entries). The top-level `from validate_skill_structure import validate` now works with no `sys.path` manipulation and no lint suppression — pytest auto-resolves the import via its standard pythonpath mechanism.

claude/.claude/skills/tests/test_skills.py
jcdendrite: **[Claude Code]** Fixed in the same change as the sibling comment — the `# noqa: E402` on the import disappears once `sys.path.insert` is gone. See the `pyproject.toml` change for the mechanism.

plugins/skill-management/.claude-plugin/plugin.json
jcdendrite: **[Claude Code]** Fixed. `plugin.json` now mirrors the reworded `marketplace.json` description (the two are intentionally byte-identical per plan §53 — same plugin, two registry surfaces).

plugins/skill-management/scripts/validate_skill_structure.py
jcdendrite: **[Claude Code]** Fixed. The citation + URL stay in `validate_skill_structure.py` (lines 20–23) as the named-constant comment where they're load-bearing for the logic. User-facing surfaces (`README.md`, `docs/skills.md`, `.claude-plugin/marketplace.json`, `plugins/skill-management/.claude-plugin/plugin.json`) now describe the concept ("frontmatter that would silently truncate from the harness's skill listing") without hardcoding `1536`. The historical `CHANGELOG.md` entry retains the literal — historical records describe what shipped at v2.0.0 and are not maintained as the world changes, unlike active docs.

plugins/skill-management/skills/skill-review/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Added the implication as a tail clause on the surviving sentence: `Description text is always-loaded context budget; body text is conditional — so overspend in the description costs more than overspend in the body.` Single line, preserves the WHY, file stays at 198/200.

README.md
jcdendrite: **[Claude Code]** Consulted both `staff-platform-engineer` and `staff-backend-engineer` on the foundation question. Both verdicts: **keep Python + pyyaml**, do not replace with bash.

The reasoning that converged across both reviews:

1. **Validation correctness.** The rule is "this must not silently truncate in the harness's strict-YAML parser." Only a strict YAML parser is correct-by-construction for that rule — a bash regex covers the trivial 90% but fails on legal block scalars, multi-line quoted strings, and inline-unquoted `: ` patterns (which is exactly the failure mode the validator exists to catch). Approximating turns a correctness gate into a coverage gate.
2. **Test-share contract.** `claude/.claude/skills/tests/test_skills.py` imports `validate()` directly from the same module the hook shells out to. A bash hook would split the implementation in two — drift is then guaranteed at the next rule change.
3. **Latency is bounded.** Python startup ~50–120ms only fires when a SKILL.md is staged (line 52–55 of the hook's early-exit guard) — not on every commit. Acceptable cost for a check that catches a class of bug otherwise invisible at author time.

Real concern surfaced by the platform review: the `pip install pyyaml` instruction was ambiguous about which python env. Fixed in `README.md` — now specifies `python3 -m pip install pyyaml` against the system interpreter the hook spawns (not a project virtualenv).

`disposition: fixed-with-modification` | `rationale: kept Python+pyyaml after explicit staff-platform + staff-backend review on the bash-vs-python question; fixed the README pyyaml install instruction surfaced by the platform review`

README.md
jcdendrite: **[Claude Code]** Fixed. Removed the migration paragraph from `README.md` — same content lives in `CHANGELOG.md` line 23 (already there), which is the right altitude for transitional ops detail.

README.md
```diff
@@ -173,7 +173,10 @@ This repo exposes a marketplace via `.claude-plugin/marketplace.json`. Each plug
...
 
 - **`lovable-cloud`** — Skills for Lovable Cloud projects: Project/Workspace Knowledge fields, edge-function auth model (ES256 gateway constraint, two-tier auth), and migration-sync workflow. `claude plugin install lovable-cloud@claude-config --scope project`
-- **`skill-review`** — Behavioral-equivalence audit for `SKILL.md` changes; gates `git commit` when staged changes include a `SKILL.md`. `claude plugin install skill-review@claude-config --scope project`
+- **`skill-management`** — Authoring guardrails for SKILL.md files: a commit-time structural validator (catches frontmatter that would silently truncate from the harness's skill listing or fail strict-YAML parsing) plus a behavioral-equivalence audit via `/skill-review`. `claude plugin install skill-management@claude-config --scope project`
+
+  Requires `python3` on `$PATH` with `pyyaml` available — install via `python3 -m pip install pyyaml` against the system interpreter the hook spawns (not a project virtualenv). The commit-time hook fails closed if either is missing.
```
jcdendrite: This critical instruction is hidden in the README. There's two problems with this:
(1) contributors to claude-code won't have a working setup to contribute after running ./install.sh because the tests will fail because the validator is run in the test harness and the validator requires this dependency
(2) this is a very non-standard way to highlight dependencies. I would imagine claude code docs have prescribed a way to do this for plugins.

jcdendrite (commented):
verified review comments have been fixed but found another blocker

README.md
jcdendrite: **[Claude Code]** Fixed (1); declining (2).

`disposition: fixed-with-modification`

(1) `install.sh` now `pip install --user pyyaml` after the existing tool-availability check, guarded by `import yaml` so re-runs are no-ops. `python3` joined the required-deps list. On failure (PEP 668, etc.) the script bails with three explicit alternatives (`--break-system-packages`, `pipx`, repo-local venv) rather than continuing to a half-installed state. README per-plugin paragraph rephrased: contributors get install.sh automatically; consumers installing the plugin elsewhere still need one `python3 -m pip install pyyaml`. PR body migration paragraph updated to match.

(2) I verified the claim against the [Claude Code plugins reference](https://code.claude.com/docs/en/plugins-reference). There is no manifest field for external runtime dependencies — the `dependencies` field in `plugin.json` is **plugin-to-plugin only**, not Python packages. The docs describe a *pattern* (not a prescription) using a `SessionStart` hook + `${CLAUDE_PLUGIN_DATA}` persistent venv with diff-based reinstall (`diff -q $ROOT/requirements.txt $DATA/requirements.txt || (cp + python3 -m venv + pip install)`). Adopting it requires: new `requirements.txt`, `SessionStart` hook entry, and rewriting `require-skill-review.sh` to prefer the venv's `python` with a fallback to system `python3` (so `pytest` — which loads the validator outside any plugin context — keeps working). That's ~25 LoC across three files and a dual-mode python selector in the hook. Given the scope-creep concerns already on this PR, I'm deferring it to a follow-up tracked as: *"Adopt the `SessionStart` + `${CLAUDE_PLUGIN_DATA}` venv pattern for skill-management's pyyaml dependency."* Happy to do it inside this PR if you'd rather close the loop now — say the word.

`rationale: install.sh covers the immediate contributor breakage in scope; the documented SessionStart pattern is a follow-up because adopting it requires hook-command rewrite + venv-vs-system python selector and conflicts with the PR's stated scope-tightness.`

jcdendrite (commented):
brittle patch, but will do follow-up PR to handle the python dependency in the right (plugin-preferred) way

PR #294, opened 2026-05-20
feat(code-review): require explicit disposition for every reviewer finding

jcdendrite:
## Summary

- Add **Finding disposition** section to `/code-review` SKILL.md: every reviewer-spawned finding must be tagged ADDRESS (default) or DEFER (closed list of 5 criteria). Names the dismissal anti-patterns that previously surfaced as silent skips — *"informational only," "no correctness impact," "general skill guidance, not PR scope," "pre-existing gap"* — and rules them out as DEFER rationales.
- Grounded in Fowler's *Opportunistic Refactoring* (default ADDRESS, tests-already-running condition) + Google eng-practices' file-a-ticket rule for pre-existing structural debt. Citations and key quotes live in a new co-located `REFERENCES.md`.
- Carve out a 500-line override in `check-skill-length.sh` for `/code-review` and `/plan-review` — structural-dispatcher skills whose item-ownership and routing tables are load-bearing and resist trimming. Default stays 200 for all other skills. Policy documented in `docs/skills.md`.

## Motivation

Orchestrating agent was filtering reviewer-subagent findings through invented dismissal categories (*"informational only," "no correctness impact"*) and dropping legitimate findings without user-visible disposition. The new section forces structured enumeration with named DEFER criteria so silent dismissal is no longer the default failure mode.

Walked against a recent multi-reviewer review: 5 of 7 findings flip from skip/defer to ADDRESS under the new rule; the remaining 2 can still legitimately DEFER under criterion #2 (coordinated multi-PR effort) with proper rationale + ticket filed + human surfaced.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/test_check_skill_length.py` — 16/16 pass, including 3 new tests covering the per-skill override (`code-review` allows up to 500, denies past 500, `plan-review` also uses the override)
- [x] `/skill-review` on the SKILL.md addition — clean (length flagged but within new 500-line cap; behavior test passes on every line)
- [x] `/claude-hook-review` + `staff-platform-engineer` on the hook diff — clean (one informational finding about plugin-subtree override gap, addressed inline as documentation in the hook header)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #296, opened 2026-05-20
respond-pr: require holistic triage before per-comment work

jcdendrite:
## Summary

- Add new step 4 **Holistic triage** to `/respond-pr`: before classifying or fixing anything, read the complete unresolved-comment set as one body and name shared root causes, contradictions, and batched-design choices across them.
- Require a four-column disposition table (Comment | Theme | Disposition | Plan) in step 5 when there are 3+ unresolved comments, forcing all comments to be seen at once before per-comment work begins.
- Renumber existing steps 4–8 → 5–9 and update cross-references in marker-lifecycle, classify lead-in, FIXED-field template, per-comment apply step, and PATCH-safety guideline.

## Why

The existing skill body already named batching as a requirement ("Classify, then batch"), but agents reached the right *commit shape* (single commit) without doing the *cross-comment analysis* — so adjacent comments that should have changed the approach (or revealed the same root cause) were patched independently.

Concrete examples from recent merged PRs in this repo:

- **PR #290** (`skill-management` rename + validator) — four threads shared one root cause: a hardcoded numeric value surfacing across README, docs, marketplace.json, plugin.json, CHANGELOG. Per-comment patches stripped the documented rationale; thread #7 then reported a regression caused by thread #6's fix. A holistic read would have framed all four as one decision about what belongs in the validator script vs. surface docs.
- **PR #278** (extract delegation triad to subagent-delegation skill) — two threads were constraints on the same design question (boundary between always-loaded CLAUDE.md signal and on-demand skill body), but were addressed as independent issues.
- **PR #280** (proactive branch-divergence detection) — two threads, same root cause (detail at wrong altitude in two README sections); two near-identical "Fixed — reverted…" replies where one would suffice.

## Mechanism

Mirrors PR #294's per-finding disposition pattern for `/code-review` subagent findings, adapted to reviewer comments. The disposition table from `/code-review` becomes the per-comment table here; the new Holistic-triage step is the missing layer that makes "Classify, then batch" actionable rather than rule-only.

## Test plan

- [x] `pytest claude/.claude/` — passes (recipes extract by `HOOK_TEST_FIXTURE` markers, not step numbers, so renumbering doesn't break them)
- [x] `ruff check claude/.claude/` — passes
- [x] `/skill-review` walked: frontmatter, trigger, voice, length (111 lines, under 200), behavior test on new lines, behavioral-equivalence audit on renumbering/compressions — all clean
- [x] `/code-review` walked on staged diff — one matched Change-type row (skill edit) dispatched to `/skill-review` already; no `staff-*` co-spawn warranted (edit changes Claude's PR-reply process, not any `staff-*` lane's output)
- [ ] Smoke test on the next `/respond-pr` invocation with 3+ comments — verify the table format renders and the triage step lands

### DIFF
--- a/claude/.claude/skills/respond-pr/SKILL.md
+++ b/claude/.claude/skills/respond-pr/SKILL.md
@@ -15,17 +15,24 @@ Fetch all review comments on the current branch's open pull request and address 
    ```
    The `require-respond-pr.sh` PreToolUse hook bypasses while THIS session's marker exists and its stored PID is alive (PID liveness check via `kill -0`). Per-session keying prevents parallel respond-pr sessions from thrashing on cleanup or leaking bypass to unrelated sessions. If the chain fails (empty `SESSION_ID`, etc.), the `capture-session-id.sh` SessionStart hook didn't run — abort and report; do not proceed without the marker, since every gated `gh` call below will be blocked.
 
-   **Marker lifecycle:** this marker must stay active for the entire skill session — step 8 is the only step that removes it. If you run other skills as intermediate steps (e.g., `/ready-for-review` before pushing in step 7), their cleanup must not touch this marker. If the marker is accidentally removed mid-session, restore it in a standalone Bash call before any subsequent `gh` command — the hook fires before the shell executes, so you cannot create the marker and use it atomically in the same call.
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #297, opened 2026-05-20
fix(skill-management): graceful SessionStart when python3 -m venv fails

jcdendrite:
## Summary

The `skill-management` plugin's SessionStart hook ran `python3 -m venv` inline. On Debian/Ubuntu systems missing the `python3-venv` apt package, Python emits a 7-line `ensurepip is not available...` banner that Claude Code surfaces as a session-start error at the top of every new session. The validator the venv supports is best-effort (`require-skill-review.sh` already falls back to system `python3`), so the venv bootstrap should degrade silently with a single graceful warning.

## What shipped

**`plugins/skill-management/hooks/provision-validator-venv.sh` (new).** SessionStart bootstrap extracted from the inline `hooks.json` command. Stderr from the `python3 -m venv` + `pip install` block is suppressed; on failure the script prints a single one-line warning that names the distro fix.

**Robustness against partial provisioning.** Added a `venv/bin/python -c 'import yaml'` import probe as the final step of the success chain — `pip install` can exit 0 while pyyaml is absent from site-packages (cached resolution, yanked release), and `require-skill-review.sh` line 104 elects `venv/bin/python` as `VALIDATOR_PYTHON` whenever the file is executable. The failure-cleanup branch now `rm -rf venv/` in addition to removing the cached `requirements.txt`, so a half-created venv left by a failed `pip install` cannot be elected as `VALIDATOR_PYTHON` between sessions.

**`plugins/skill-management/hooks/hooks.json`.** SessionStart command now invokes the script with `CLAUDE_PLUGIN_ROOT` / `CLAUDE_PLUGIN_DATA` passed through.

**`plugins/skill-management/.claude-plugin/plugin.json`.** Patch bump `2.1.1` → `2.1.2` per the `plugin-semver` skill (hook bug fix; no behavior change on systems where the venv was already provisioning cleanly).

**`claude/.claude/hooks/tests/test_provision_validator_venv.py` (new, 7 tests).** Covers: venv-creation failure path (no stderr leak, exit 0, graceful warning), pip-failure cleanup (`rm -rf venv/`), import-probe catches the pyyaml-absent corner case, python3-absent path, cache-hit short-circuit. Includes a JSON-parsed drift-detector that asserts the SessionStart command in `hooks.json` still references the script (catches re-inlining the logic).

## Test plan

- [x] `pytest claude/.claude/` — all tests pass (7 new + existing suite)
- [x] Live smoke test on the affected Debian/Ubuntu machine: ensurepip banner suppressed, single graceful warning emitted, no stale `venv/` or `requirements.txt` left in plugin data dir
- [ ] Open a new Claude Code session on the affected machine post-merge and confirm the SessionStart banner is gone

### DIFF
--- a/claude/.claude/hooks/tests/test_provision_validator_venv.py
+++ b/claude/.claude/hooks/tests/test_provision_validator_venv.py
@@ -1,0 +1,241 @@
+"""Tests for plugins/skill-management/hooks/provision-validator-venv.sh.
+
+The script bootstraps a Python venv with pyyaml for the SKILL.md structural
+validator. The motivating regression: on Debian/Ubuntu systems without
+`python3-venv`, `python3 -m venv` emits a multi-line ensurepip banner that
+SessionStart surfaces at the top of every new session. The script must
+degrade gracefully — single one-line warning, exit 0, no banner leakage.
+"""
+from __future__ import annotations
+
+import json
+import subprocess
+from pathlib import Path
+
+import pytest
+from helpers import HOOKS_DIR
+
+_PLUGINS_DIR = HOOKS_DIR.parent.parent.parent / "plugins"
+PROVISION_SCRIPT = (
+    _PLUGINS_DIR / "skill-management" / "hooks" / "provision-validator-venv.sh"
+)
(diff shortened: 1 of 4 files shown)


jcdendrite (commented):
lgtm

PR #300, opened 2026-05-20
fix: allow chained marker.sh write && git commit in pre-commit gates

jcdendrite:
## Summary

PreToolUse hooks fire **once per Bash tool call before the chain runs**, so the naturally-typed `~/.claude/scripts/marker.sh write code-review && git commit` was denied — the on-disk marker check fails because the chain hasn't yet created the marker. The two-call workaround was non-obvious; agents hit this repeatedly.

The branch lands in two commits: the bug fix, and a security tightening that closes two text-match bypasses surfaced by post-review questioning.

## What shipped

**Commit 1 — fix: allow chained marker.sh write && git commit**

- `claude/.claude/hooks/_lib.sh` — new `_lib_chains_marker_write_before_commit` helper that detects an in-chain `marker.sh write <skill>` preceding `git commit`. Mirrored byte-identical into `plugins/skill-management/hooks/_lib.sh` (the existing `test_plugin_lib_sh_matches_stowed_lib_sh` pins these in sync).
- `require-code-review.sh` and `require-skill-review.sh` exit 0 when the helper matches — `marker.sh` is the only sanctioned writer in either case, so the in-chain write is equivalent evidence the on-disk marker would later provide. The structural validator in `require-skill-review.sh` still fires before the bypass.
- `enforce-marker-script-shape.sh` — new `VALID_CHAINED_COMMIT_PATTERN` accepts a chain of `marker.sh write <skill>` shapes followed by `git commit`. Trailing content after `git commit` cannot contain `& | ; < >`, so `marker.sh write X && git commit && curl evil.com` stays denied.
- `plugins/skill-management/.claude-plugin/plugin.json` bumped 2.1.1 → 2.2.0 (minor: gate trigger broadened to fire in strictly more cases).

**Commit 2 — fix(security): anchor marker.sh chain bypass at command start**

Investigation prompted by post-CISO-review questioning ("are there gaps with globs or heredoc bodies?") found two real bypass paths the fragment-walking helper allowed:

- `echo ~/.claude/scripts/marker.sh write code-review && git commit -m foo` — `echo` doesn't actually invoke marker.sh, but the marker.sh path text after whitespace text-matched the fragment-walking helper and bypassed the gate. Same vector via `FOO=bar marker.sh ...`, `bash -c '...marker.sh...'`, and `sudo marker.sh ...`. `enforce-marker-script-shape`'s Stage-2 anchor doesn't catch these either — it only fires when the WHOLE command starts with marker.sh.
- `cat <<EOF | bash\n~/.claude/scripts/marker.sh write code-review\nEOF && git commit` — heredoc body lines were treated as fragments by the splitter, and the marker.sh text inside the body wedged the gate. The piped `bash` would execute the entire body (including any additional commands like `curl evil.com`) and the commit would then proceed.

Fix: replace fragment-walking with a strict command-start anchored regex that mirrors `VALID_CHAINED_COMMIT_PATTERN`. Only the literal shape `<marker.sh write skill> (&& <marker.sh write skill2>)* && git commit ...` is honored; wrapper commands, env-var prefixes, `bash -c`, heredoc bodies, and pipes all fail the anchor.

**Tests** — 13 new negative-path tests in total across `test_require_code_review.py`, `test_require_skill_review.py`, and `test_enforce_marker_script_shape.py`:
- Chained marker.sh write && git commit allowed (with/without existing on-disk marker).
- Wrong-skill chain does not authorize.
- Marker write after commit does not authorize (order matters).
- Bare-name `marker.sh` and non-canonical paths do not authorize.
- `echo`, env-var prefix, `bash -c`, heredoc-pipe wrappers do not authorize.
- Activate-skill chain does not authorize (only `write` permitted in chain prefix).
- Quoted marker-write text inside a commit message does not authorize.
- Post-commit chain operators (`&&`, `;`, `>`) denied at the shape gate.
- Structural validator still fires when chained (parallel to the wrong-state path).

## Security review

CISO-reviewer pass identified two findings, both addressed in commit 1:

1. **Trailing chain operators after `git commit`** — `(\s.*)?$` allowed `marker.sh write X && git commit && curl evil.com` through the shape gate. Tightened to `(\s+[^&|;<>]*)?$`.
2. **Non-canonical marker.sh path** — fragment-walker matched any path ending in `/marker.sh`. Pinned to canonical paths.

Follow-up review surfaced two more (commit 2):

3. **Wrapper-command text-match** — `echo`/`FOO=bar`/`bash -c`/`sudo` could appear before marker.sh in a fragment and text-match the helper without actually invoking marker.sh.
4. **Heredoc-pipe body text** — heredoc body lines containing marker.sh text were treated as fragments, wedging the gate via body text rather than execution.

Both closed by anchoring the helper at command start (mirrors `VALID_CHAINED_COMMIT_PATTERN`).

## Test plan

- [x] `pytest claude/.claude/` passes (107 tests in the three touched suites).
- [x] `ruff check claude/.claude/` clean.
- [ ] Spot-check the chained form works end-to-end after install: stage any diff, run `/code-review`, then `~/.claude/scripts/marker.sh write code-review && git commit -m "..."` should succeed; `echo ~/.claude/scripts/marker.sh write code-review && git commit -m "..."` should be denied.

## Post-merge install (skill-management plugin)

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to `claude-config` are documented here. Format follows [Keep
 
 ### Fixed
 
+- **Chained `marker.sh write && git commit` no longer denied** — three coordinated changes: `enforce-marker-script-shape.sh` accepts a chain of one or more `marker.sh write <skill>` invocations followed by `git commit`; `require-code-review.sh` and `require-skill-review.sh` (plugin bumped to 2.2.0) honor an in-chain marker write that precedes the commit. PreToolUse fires once per Bash tool call before the chain runs, so an on-disk marker check would otherwise deny these naturally-typed forms even though the chain would create the marker. Chains to non-commit tails (curl, rm, semicolon-arbitrary) stay denied, and the structural validator still fires when chained, so malformed SKILL.md files can't slip through.
 - Two false-positives in `enforce-marker-script-shape.sh` that blocked legitimate `marker.sh` invocations with certain argument orderings (#187)
 - `cleanup-merged-branches.sh` now skips locked worktrees rather than erroring; a follow-up fix adds unlock-and-remove for worktrees that can be released (#174, #177)
 - Memory-write hook debounce replaced with active-marker bypass pattern, eliminating per-turn UUID thrash (#170)
--- a/claude/.claude/hooks/_lib.sh
+++ b/claude/.claude/hooks/_lib.sh
@@ -77,40 +77,43 @@ _lib_split_fragments() {
 }
 
 # Decide whether a command chains `marker.sh write <skill>` before its first
-# `git commit` fragment. PreToolUse hooks fire once per Bash tool call before
-# the chain runs, so an on-disk marker check denies naturally-typed forms like
+# `git commit`. PreToolUse hooks fire once per Bash tool call before the chain
+# runs, so an on-disk marker check denies naturally-typed forms like
 # `marker.sh write code-review && git commit`. When the same Bash call will
 # write the marker before invoking commit, the in-chain marker.sh invocation
-# is the same evidence the on-disk marker would later provide -- marker.sh is
+# is the same evidence the on-disk marker would later provide — marker.sh is
 # the only sanctioned writer in either case.
 #
+# **Anchored at command start, not fragment start.** A fragment-walking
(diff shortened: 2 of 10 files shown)


jcdendrite (commented):
lgtm

PR #302, opened 2026-05-21
fix(skill-management): probe yaml import in cache-hit branch (2.2.1)

jcdendrite:
## Summary

The cache-hit short-circuit in `provision-validator-venv.sh` gated `exit 0`
on `diff -q requirements.txt` alone — the cache key is *what was requested*,
not *what currently works*. A previously-good venv can break without
`requirements.txt` changing (system Python upgrade evicts the interpreter
target, `rm` against `site-packages`, partial snapshot restore, `pip
uninstall` against the wrong interpreter, ABI swap on distro upgrade). In
any of those cases the short-circuit returned `exit 0` without
re-provisioning, and `require-skill-review.sh` then elected the broken
`venv/bin/python` as `VALIDATOR_PYTHON` (it gates on `-x` alone). The
structural validator failed at commit time with `ModuleNotFoundError: No
module named 'yaml'` or silently ran against a broken interpreter.

Fix: gate the cache-hit `exit 0` on the same yaml import probe the
provisioning path already uses (`venv/bin/python -c 'import yaml'`), so
drift is caught on both sides of the cache decision. On probe failure the
script falls through to the existing reprovisioning path, whose cleanup
removes the broken venv.

- `plugins/skill-management/hooks/provision-validator-venv.sh` — add
  `[ -x venv/bin/python ]` + `python -c 'import yaml'` to the cache-hit
  conditional; expand the WHY comment.
- `claude/.claude/hooks/tests/test_provision_validator_venv.py` — add
  `_stub_venv_python` helper; update the existing cache-hit test to stand
  up a working venv stub; add
  `test_broken_venv_falls_through_to_reprovisioning` covering the drift
  case.
- `plugins/skill-management/.claude-plugin/plugin.json` — bump
  `2.2.0` → `2.2.1` (patch, backward-compatible bug fix).

## Background

Follow-up to #297 (graceful SessionStart) and #301 (the 2.1.3 re-bump that
delivered #297 to caches). #297 added the import probe to the
*initial-provisioning* path but not to the *cache-hit* path; this PR closes
that gap symmetrically.

Rebased onto main after #300 landed (main bumped to 2.2.0 by the
marker.sh-chain fix); the patch ships as 2.2.1.

## Test plan

- [x] `pytest claude/.claude/` (full suite passes, including the new drift
      test in `test_provision_validator_venv.py`)
- [ ] After merge: confirm consumer plugin caches refresh to 2.2.1 and the
      new probe fires on a SessionStart whose cached venv has had its yaml
      package removed.

## Post-merge install (per plugin-semver)

```
```

### DIFF
--- a/claude/.claude/hooks/tests/test_provision_validator_venv.py
+++ b/claude/.claude/hooks/tests/test_provision_validator_venv.py
@@ -193,14 +193,36 @@ class TestPython3Absent:
         assert "python3 not found" in result.stderr
 
 
+def _stub_venv_python(data: Path, body: str) -> Path:
+    """Drop an executable stub at data/venv/bin/python with `body` as its
+    script body. Used to simulate a previously-provisioned (or
+    previously-provisioned-then-broken) venv interpreter."""
+    venv_bin = data / "venv" / "bin"
+    venv_bin.mkdir(parents=True, exist_ok=True)
+    stub = venv_bin / "python"
+    stub.write_text("#!/bin/bash\n" + body)
+    stub.chmod(0o755)
+    return stub
+
+
 class TestCacheHit:
-    """When the cached requirements.txt matches the plugin's requirements.txt,
-    the script must short-circuit — no python3 invocation, no stderr."""
+    """When the cached requirements.txt matches the plugin's requirements.txt
+    AND the cached venv's python can still `import yaml`, the script must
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #303, opened 2026-05-21
docs: worktree-enforcement case study + captured deny artifact

jcdendrite:
## Summary

Adds `docs/case-studies.md` — a longer-form case study examining whether the worktree-enforcement hook is over-engineered vs. CLAUDE.md prose alone — plus one-line cross-references in `README.md` and `docs/design-decisions.md`. The case study cites primary sources (hook script line numbers and the PRs that introduced each behavior) and embeds a captured live transcript of the chained-`cd` hard-deny gate firing.

## What's in the case study

- **Question + short answer.** Both can be true. Prose-only is viable for a single-session, single-developer, private-repo workflow; the hook earns its keep when the workflow deviates from any of those three premises.
- **Failure mode 1: cross-session race.** Strictly out of prose's reach — the failing state isn't in either session's context.
- **Failure mode 2: approximate within-session compliance.** Prose can carry most of the load with a sufficiently mechanical instruction; the hook closes the residual escape rate at the tool-call boundary. Whether the mechanical guarantee is worth the hook is a cost-asymmetry call (favored on a public repo).
- **Hook-internal refinement: self-correcting deny messages.** Not a prose-vs-hook comparison (the gotcha exists only because the hook exists) — documented as a hook-design principle. Splits PR #59's `cwd_anchor_note_if_chained` helper (suffixes the malformed/non-allowlist deny paths) from PR #131's chained-`cd` hard-deny gate at lines 146–155 (its line-153 string is what the Live demo captures).
- **Live demo.** A captured transcript of the hard-deny gate firing, embedded as a Markdown code block. Captured once, manually — see "How the artifact was captured" below.
- **The general philosophy / When prose-only is sufficient.** Honest scoping of the prose-only claim and the cost-asymmetry framing for when the hook earns its complexity.

## How the artifact was captured

The embedded transcript is a **captured-once manual smoke test**, not a CI-run check. The repo intentionally has no `claude -p` automated harness for hook smoke tests because:

- The repo is public, and `--dangerously-skip-permissions` on a public repo is a security non-starter.
- The harness would charge against a personal API key, not a Max plan.
- Single-shot Claude runs are probabilistic; CI failures would be flaky in ways that don't map to real bugs.

Capture model: a single Claude Code session, working in this branch's worktree, deliberately invoked the chained `cd && git log` pattern from inside the worktree to elicit the line-153 hard-deny gate. The session paused, the human captured the conversation, the session resumed to embed the verbatim `permissionDecisionReason` text. The embedded text was diffed byte-for-byte against the on-disk hook script before commit.

## Out of scope

- Demonstrating failure modes 1 (cross-session race) or 2 (within-session compliance) live — Mode 1 needs two concurrent sessions; Mode 2 needs a longer edit-then-merge sequence; neither fits a single-session capture.
- Adding pytest cases — coverage of the hook's decision logic is already 40+ tests.
- Refactoring `require-worktree-for-git-writes.sh` or its file-write sibling.
- Migrating `docs/design-decisions.md` §10–§14 into `case-studies.md` (possible future cleanup, not authorized here).

## Test plan

- [ ] Visit the new `docs/case-studies.md` "Live demo" subsection in the GitHub diff view and confirm the embedded deny block renders cleanly.
- [ ] Spot-check the embedded deny string against `claude/.claude/hooks/require-worktree-for-git-writes.sh` line 153 (the hard-deny `emit_deny` call inside the chained-`cd` gate at lines 146–155, introduced by PR #131).
- [ ] Spot-check the helper-text quote in the "Hook-internal refinement" section against `cwd_anchor_note_if_chained` (lines 65–69, introduced by PR #59).
- [ ] Confirm the README and design-decisions cross-references render and link correctly.


jcdendrite (commented):
lgtm

PR #310, opened 2026-05-22
feat(skill-management): elevate aggregate corpus budget check to plugin

jcdendrite:
## Summary

- Moves the 8000-char aggregate skill-listing budget check from `claude-config`'s pytest suite into the `skill-management` plugin so downstream plugin consumers get it automatically at commit time
- Adds `corpus_budget_violations()` pure function and `--corpus` CLI mode to `validate_skill_structure.py` (single source of truth — hook and test suite both import from here)
- Adds a non-blocking corpus warning block to `require-skill-review.sh` that fires on every SKILL.md commit; warns to stderr, never denies (hard enforcement stays in pytest/CI)
- Widens `.github/workflows/tests.yml` change-detection REGEX to include `plugins/skill-management/` so CI runs on plugin file changes
- Bumps plugin to `2.3.0`

## Design decisions

- **Hard enforcement stays in pytest/CI, not as a commit deny.** A budget violation doesn't block the commit — the warning nudges the author, but denial would create an unrecoverable state if the corpus is already over budget when the plugin is first installed. Hard gate via `test_total_within_listing_budget`.
- **Two separate validator invocations in the hook.** Per-file structural validation (hard deny) and corpus check (soft warn) use separate subprocess calls so their exit codes remain separable.
- **`corpus_budget_violations()` is a pure function.** No filesystem discovery — caller supplies the file list, so project-scope and user-scope listings are never inadvertently conflated.
- **`timeout 10s` on corpus subprocess.** Bounds the hook on slow Python startup (NFS venv, throttled container) without blocking commits — the `|| true` inside the command substitution preserves the non-blocking contract on timeout.

## Test plan

- [x] `pytest claude/.claude/` — 1202 tests pass (includes `TestCorpusBudgetWarning` (4 hook tests) and `TestCorpusBudgetFunction` (8 unit tests))
- [ ] Manual smoke: stage a SKILL.md change in a repo with the plugin installed and confirm corpus warning appears on stderr when total exceeds 8000 chars

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -54,8 +54,8 @@ jobs:
 
           # Include any directory that contains test files or whose contents are
           # read by tests. Update this list when a new test or test-input
-          # directory appears under claude/.claude/.
-          REGEX='^(claude/\.claude/(hooks|skills|scripts|tests)/|\.github/workflows/tests\.yml|pyproject\.toml)'
+          # directory appears under claude/.claude/ or plugins/skill-management/.
+          REGEX='^(claude/\.claude/(hooks|skills|scripts|tests)/|plugins/skill-management/|\.github/workflows/tests\.yml|pyproject\.toml)'
 
           MATCHED=$(echo "$CHANGED" | grep -E "$REGEX" || true)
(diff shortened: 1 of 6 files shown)


jcdendrite (commented):
lgtm

PR #312, opened 2026-05-22
docs: note contributor .venv is main-tree-only

jcdendrite:
## Summary

- Linked git worktrees never inherit the contributor `.venv` (it is gitignored at the repo root). Documents the correct invocation (`../../../.venv/bin/pytest` / `../../../.venv/bin/ruff`) in all three contributor-facing docs: `CLAUDE.md`, `README.md`, and `CONTRIBUTING.md`.
- README previously carried the venv note in two places ("Working inside a worktree" and "Tests"). Trims the first occurrence to a forward reference to keep a single canonical location.
- Adds "exactly three levels deep" at each mention so the relative prefix `../../../` is traceable to the documented worktree path (`.claude/worktrees/<branch>/`) rather than reading as a magic number.

## Test plan

- [ ] Verify `../../../.venv/bin/pytest --version` resolves correctly from `.claude/worktrees/<any-branch>/` once `pip install -r requirements-dev.txt` has been run in the main tree.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,8 +25,9 @@ non-read-only git operations must run inside a linked worktree
 (`git worktree add .claude/worktrees/<branch> -b <branch>`) or an agent with
 `isolation: worktree`. See README "Worktree enforcement" for why. The
 contributor `.venv` lives at the main worktree root only — it is gitignored,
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #313, opened 2026-05-22
docs: design-decisions §16 — check-runner is a verdict-quality tool

jcdendrite:
## Summary

Adds **§16** to `docs/design-decisions.md`, recording a measured finding about `check-runner`.

A transcript-corpus measurement found the Claude Code harness already truncates Bash tool output at 30 KB — persisting the overflow to a file and returning a ~2 KB preview. A heavy test-suite run inline therefore does not flood the parent context, and no large inline test result in the corpus was followed by a compaction. So `check-runner` is **not a context-cost optimization**: the platform already prevents the flood. §16 records the corrected rationale — `check-runner` is kept, justified on verdict quality (a curated pass/fail + failure tail, vs. the harness's blind first-2 KB preview), not token economy.

Net diff is §16 only. The branch's first commit drafted three "cost-optimization case studies"; review reduced that to this single design-decision plus two opportunity issues — #314 (`/handoff` adoption) and #315 (Opus vs. Sonnet routing audit). The draft commit is reverted within the branch.

## Test plan

Docs-only change — no executable code.

- [ ] §16's figures were re-derived during `/code-review`: 30 KB truncation cap, ~2 KB preview, 0/134 large-inline-run-to-compaction correlation, ~100-token median inline output.
- [ ] No private-project identifiers.


jcdendrite (commented):
lgtm

PR #316, opened 2026-05-22
fix: surface python3-venv prerequisite for contributor venv setup

jcdendrite:
## Problem

On Debian/Ubuntu, `python3 -m venv` exits 0 even when the `python3-venv` apt package is absent — it writes the venv skeleton and `pyvenv.cfg`, then fails silently at the ensurepip pip-bootstrap step, leaving a venv with **no `pip`**. The documented next step `.venv/bin/pip install -r requirements-dev.txt` then fails with a "no such file" error that reads as a path bug rather than a setup bug.

This is a confirmed incident: a session hit a bare main-tree `.venv` and could not run the pinned test suite. Verified on the affected machine — `python3.12-venv` is `Installed: (none)`, `/usr/lib/python3.12/ensurepip` does not exist, and `python3 -c "import ensurepip"` fails.

PR #312 diagnosed an adjacent symptom (worktree relative-path resolution) and explicitly dismissed "python3-venv not installed" as a misdiagnosis — but that *was* the root cause. #312 was docs-only, so it never populated a venv and the real blocker remained.

## What shipped

- **README.md** — the canonical home for the fix: a Debian/Ubuntu-scoped prerequisite note in the Tests section naming the symptom (`.venv/bin/pip` absent after venv creation) and the fix (`sudo apt install python3.12-venv`). macOS (python.org / Homebrew) bundles `ensurepip` and is marked unaffected.
- **CONTRIBUTING.md** — the "Tests and lint" section duplicated README's "Tests" section near-verbatim (same setup commands, same `.venv` worktree note, same CI note). Collapsed it to a one-line deferral to [README#tests](./README.md#tests) so there is a single canonical home; the contributor-process fact (both checks must pass before PR review) is kept.
- **install.sh** — probes `import ensurepip` and prints a non-fatal warning before the venv setup hint, so a broken `python3` is flagged at the point a contributor would otherwise copy-paste into a silent failure. Non-fatal because venv tooling is optional — `install.sh`'s core job (stow) does not need it.

## Notes

Approach reviewed by the staff-platform-engineer reviewer, which prescribed this docs-note + setup-probe shape and confirmed the failure mode is Debian/Ubuntu-specific.

## Test plan

- [x] `ruff check claude/.claude/` — clean (ruff 0.6.9, matches pinned `0.6.*`).
- [x] `pytest claude/.claude/` — passes; the diff is docs + a non-Python script, so the suite is unaffected.
- [ ] Reviewer: confirm the `README#tests` anchor resolves and the CONTRIBUTING deferral reads correctly on GitHub.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,20 +30,7 @@ This repo is public. Do not commit anything that identifies a private project, o
 
 ## Tests and lint
 
-Pins live in [`requirements-dev.txt`](./requirements-dev.txt) — the same file CI's install step reads, so versions stay in sync.
-
-```bash
-python3 -m venv .venv
-.venv/bin/pip install -r requirements-dev.txt
-.venv/bin/pytest claude/.claude/    # hook and skill tests
-.venv/bin/ruff check claude/.claude/  # lint
-```
-
-**Debian/Ubuntu:** `python3 -m venv` needs the `python3-venv` apt package (or the version-specific `python3.12-venv`). Without it, venv creation exits 0 but produces no `pip`, so the second command fails with a confusing "no such file" error. If `.venv/bin/pip` is missing after the first command, run `sudo apt install python3.12-venv`, delete `.venv`, and re-run. macOS (python.org or Homebrew) bundles `ensurepip` and is unaffected.
-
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #318, opened 2026-05-22
docs: add single-source-of-truth rule to global CLAUDE.md

jcdendrite:
## Summary

Duplicated content across files — the same rule, value, or explanation copied into multiple places — recurred as a PR-review finding (most recently a README ↔ CONTRIBUTING duplication surfaced by #316). The global `CLAUDE.md` Engineering Judgment section carried only a partial form of the rule (`"avoid duplicating managed values across files where they can drift out of sync"`) — scoped to values and easily read as code-only.

## What shipped

- **`claude/.claude/CLAUDE.md`** — replaced the partial bullet with a general **single source of truth** rule. It names DRY, states explicitly that the principle governs *knowledge* (so it covers prose and docs, not just code text), gives the action (one canonical home; other sites defer), and lists the deliberate, named exceptions: test code is DAMP; load-bearing instructional prose that must stand alone; a small duplicated value over a bad abstraction. The replaced bullet's content (drift, hardcoded-value-vs-over-abstraction) is preserved inside the new rule. One physical line in, one out — no change to the always-loaded line budget.
- **`docs/design-decisions.md`** — new `## 17` entry recording *why* the rule was elevated, why a `/code-review` checklist item was rejected (it would duplicate an always-loaded rule on a drift-prone surface), and a `### Sources` block. The primary-source citation lives here, not in `CLAUDE.md` — the always-loaded file states the rule; the sourcing is load-on-demand reference, matching the §15/§16 pattern.

## Primary source

DRY's canonical definition is the single-source-of-truth statement: *"Every piece of knowledge must have a single, unambiguous, authoritative representation within a system"* — Hunt & Thomas, *The Pragmatic Programmer* (20th Anniversary Edition, 2020), "The Evils of Duplication." Verified directly against the publisher's chapter extract; the chapter frames duplication as duplicated *knowledge* across specs, code, and tests — which is why the rule extends to prose and docs.

## Test plan

- [x] `/ai-instruction-and-memory-files` — placement, line budget, behavior test, no-PR-refs, no harmful duplication: all pass.
- [x] `/code-review` — no findings.
- [ ] Reviewer: confirm the rule's wording reads well as an always-loaded instruction and the §17 entry is at the right altitude.


jcdendrite (commented):
lgtm

PR #327, opened 2026-05-26
feat(claude-md): add function-decomposition heuristic to Engineering Judgment

jcdendrite:
## Summary

- Adds a new **Engineering Judgment** rule: _Extract functions when you need to explain what a fragment does_
- Grounded in Fowler's comprehension-effort signal — the test is whether a fragment requires effort to understand its *what*, not its line count
- Includes a false-positive guard: a large function that expresses one nameable thing without inner confusion is fine

## Primary source

> "If you have to spend effort into looking at a fragment of code to figure out *what* it's doing, then you should extract it into a function and name the function after that 'what'."
> — Martin Fowler, "Function Length," martinfowler.com/bliki/FunctionLength.html (30 Nov 2016)

Triangulated by Beck (_Smalltalk Best Practice Patterns_, Composed Method pattern) and Martin (_Clean Code_, ch. 3) — all three converge on the comprehension/task signal over line count.

## Test plan

- [x] 1261 tests pass (`pytest claude/.claude/`)
- [x] `ai-instruction-and-memory-files` review: placement correct, behavior test passes, no duplication, under 200-line ceiling


jcdendrite (commented):
lgtm

Issue #315, opened 2026-05-22
Audit Opus vs. Sonnet routing — Opus is ~92% of model spend


jcdendrite:
## Problem

Over a 35-day transcript corpus, estimated API-priced spend by model:

| Model | Sessions | Est. cost | Share |
|---|---|---|---|
| Opus | 530 | ~\$31,100 | ~92% |
| Sonnet | 848 | ~\$2,780 | ~8% |
| Haiku | 42 | ~\$10 | <1% |
| **Total** | | **~\$33,900** | |

Sonnet is exactly 1/5 of Opus per token on every line of the rate card; Haiku is 1/15. Opus cache-reads alone — pure context replay — are ~\$15,300.

## What is already realized

Structural routing is in place: reviewer agents and `code-writer` carry `model: sonnet`; `Explore` is pinned to Haiku; `claude/.claude/CLAUDE.md` Model Routing assigns Opus to judgment/orchestration and Sonnet to code reading/writing.

## What is unmeasured — the audit

The open question is *parent-session* routing: when an Opus orchestrator session does a long stretch of code reading or writing itself instead of delegating, that is Sonnet-tier work running at Opus rates. `token-analyzer.py`'s `Opus -> Sonnet candidates` heuristic found 0 clean candidates — but only because it excludes any session with edits, and edits are exactly Sonnet's lane. The heuristic under-counts the headroom.

## Sensitivity

Rerouting a slice of Opus work to Sonnet replaces its cost with 1/5:

| Opus spend rerouted | Saved (per 35 days) |
|---|---|
| 10% | ~\$2,500 |
| 30% | ~\$7,500 |
| 50% | ~\$12,500 |

## Proposed work

Sample the high-cost Opus sessions and classify each stretch as judgment/orchestration (correctly Opus) vs. code-tier work (should be Sonnet or delegated). That converts the sensitivity range into a real number and indicates whether the fix is behavioral (delegate more) or structural (routing rules).

## Measurement notes

Per-model token volumes from `claude/.claude/scripts/token-analyzer.py`; costs are volumes x published Anthropic API rates. Per-model session counts double-count multi-model sessions. Subagent sessions are not separate transcript files, so only aggregate per-model spend is available.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jcdendrite:
Audit complete. Findings and the new `audit-routing` subcommand are in PR #329.

[...]

PR #332, opened 2026-05-26
GH-315: add intent-based nudge to subagent-delegation skill

jcdendrite:
## Summary

- Replaces threshold-based stay-inline language with an intent-based rule in `subagent-delegation/SKILL.md`
- Fires before the first exploratory read (target unknown), not after the 2nd/3rd command
- Adds "first exploratory read (target unknown)" to the frontmatter TRIGGER list

## Context

GH-315 phase 3. The `audit-routing-shape` histogram (35-day Opus transcript window) showed 86% of code-read turns end without an inline edit or dispatch (D3-neither). ~97% of code-read streaks are length-1 — the existing "2nd/3rd Bash command" threshold catches almost nothing.

## What changed

Four targeted edits to `subagent-delegation/SKILL.md`:

1. **Frontmatter** — added `first exploratory read (target unknown);` to the TRIGGER list
2. **Step 1 "Stays inline" bullet** — replaced "A single targeted lookup" with explicit intent language: if you already know the target, the lookup stays inline; if you are exploring to find it, dispatch even for a single command
3. **Step 2 codebase-discovery paragraph** — same intent-based rule replacing "dispatch when the search is broad or spans more than ~3 queries"; the ~3-query threshold is retained as a trailing-indicator note
4. **Step 2 deduplication** — removed the verbatim inline-carve-out sentence from the discovery paragraph (it restated Step 1 exactly); Step 1 is the canonical source for that rule

## Test plan

- [x] `/skill-review` clean — behavioral-equivalence audit passed, no N rows
- [x] `/code-review` clean — no findings
- [ ] Manual spot-check: verify the three borderline scenarios (inline-eligible, dispatch-required, ambiguous-now-clear) give clear answers with the new prose

### DIFF
--- a/claude/.claude/skills/subagent-delegation/SKILL.md
+++ b/claude/.claude/skills/subagent-delegation/SKILL.md
@@ -137,11 +137,9 @@ than running it inline:
 
 Discovery output inhales into the parent context exactly like a check
 suite does, and an auto-mode parent on Opus pays that in the most
-expensive tokens — for output it only needed an *answer* from. A lookup
-stays inline when you already know the target — one `grep` for a
-specific symbol, one `Read` of a path you have already identified.
-Dispatch any read whose purpose is to explore, map, or locate — even a
-single command. The ~3-query threshold is a trailing indicator; by the
+expensive tokens — for output it only needed an *answer* from. Dispatch
+any read whose purpose is to explore, map, or locate — even a single
+command. The ~3-query threshold is a trailing indicator; by the
 time a third command fires, multiple turns of exploratory output have
 already landed in context.
 
--- a/claude/.claude/skills/subagent-delegation/SKILL.md
+++ b/claude/.claude/skills/subagent-delegation/SKILL.md
@@ -3,8 +3,9 @@ name: subagent-delegation
 description: >
   When to dispatch to a subagent vs inline. TRIGGER when: full
   check suite or full-project verification; broad codebase search;
-  2nd/3rd Bash command toward same question; delegating
-  implementation; reporting check-runner test counts.
+  first exploratory read (target unknown); 2nd/3rd Bash command
+  toward same question; delegating implementation; reporting
+  check-runner test counts.
   DO NOT TRIGGER when: single targeted lookup; comprehension read
   feeding your own writing/review/design; Edit/Write sequences;
   output requiring line-by-line reasoning.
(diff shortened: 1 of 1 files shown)


claude/.claude/skills/subagent-delegation/SKILL.md
```diff
@@ -134,10 +137,13 @@ than running it inline:
 
 Discovery output inhales into the parent context exactly like a check
 suite does, and an auto-mode parent on Opus pays that in the most
-expensive tokens — for output it only needed an *answer* from. A single
-targeted lookup — one `grep` for a known symbol, one `Read` of a known
-path — stays inline; dispatch when the search is broad or spans more
-than ~3 queries.
+expensive tokens — for output it only needed an *answer* from. A lookup
+stays inline when you already know the target — one `grep` for a
```
jcdendrite: There's some prose here that's repeated from above. Where is it load-bearing? Do you need to phrase it differently here? Why or why not?

jcdendrite (commented):
I'd also like you to run this prose against some of the non-delegation cases you discovered in transcript-analysis and see how it would do.

In fact, this may be a very good use case to build out more of the skill tests that use claude -p.

claude/.claude/skills/subagent-delegation/SKILL.md
jcdendrite: **[Claude Code]** Fixed. The sentence "A lookup stays inline when you already know the target — one `grep` for a specific symbol, one `Read` of a path you have already identified." was restating Step 1 verbatim. Step 1 is the canonical source for the inline carve-out; Step 2's unique content is the "even a single exploratory command" sharpening and the trailing-indicator note. Removed in the latest commit. `disposition: fixed-as-requested`

jcdendrite:
**[Claude Code]** Ran `audit-routing-shape` against the last 60 days of transcripts to check the new prose against non-delegation cases.

**D3 cross-tab findings (key non-delegation cases):**

- **2,038 inline-edit turns** (Opus read then edited) — the "comprehension read feeding your own writing/review/design" carve-out correctly classifies these as stays-inline. No over-triggering.
- **2,484 neither/D1=1 turns** (single reads not followed by edit or dispatch) — these are the ambiguous territory. The intent-based distinction ("already know the target" vs. "exploring to find it") is the right discriminator, and both carve-outs (targeted lookup, comprehension read) protect the legitimate inline cases.

The prose handles the common non-delegation cases correctly through the carve-outs.

**`claude -p` skill tests:** Good call — filed as issue #333 with five test-case shapes (comprehension read feeding edit, targeted lookup with known path, targeted grep, exploratory read, exploratory one-command). `disposition: DEFERRED` | `follow-up-ticket: #333`

jcdendrite (commented):
changes lgtm

Issue #333, opened 2026-05-26
Build claude -p skill tests for subagent-delegation non-delegation cases


jcdendrite:
## Background

Raised in PR #332 review. The `subagent-delegation` skill was updated with intent-based language for the codebase-discovery dispatch rule. The reviewer asked to validate the new prose against non-delegation cases from transcript-analysis and noted this is a good use case for `claude -p` skill tests.

## What to build

Adversarial test cases that verify the skill **does not** over-trigger on stays-inline scenarios:

1. **Comprehension read feeding edit** — Opus reads a file, then edits it. Expected: skill fires DO NOT TRIGGER ("comprehension read feeding your own writing/review/design").
2. **Targeted lookup with known path** — single `Read` of a specific known file. Expected: DO NOT TRIGGER ("lookup where you already know the target").
3. **Targeted grep for known symbol** — single `grep` for a symbol already named in the task. Expected: DO NOT TRIGGER.
4. **Exploratory read (target unknown)** — reading to find where something is defined. Expected: TRIGGER ("first exploratory read (target unknown)").
5. **Exploratory even for one command** — "let me check what's in this directory" without a specific file in mind. Expected: TRIGGER.

## Data source

`transcript-analysis.py audit-routing-shape` gives the turn-shape distributions:
- D3-inline-edit turns (2,038, 35d window): Opus read then edited — cases 1–3 above
- D3-neither/D1=1 turns (2,484): single reads not followed by edit or dispatch — ambiguous territory (cases 4–5)

## Implementation approach

`claude -p` adversarial harness similar to the existing skill eval runner (`evals/run_skill_evals.py`). See also the reviewer's note in PR #332 about this being a good `claude -p` use case.

## Related

- PR #332 (subagent-delegation intent-based nudge)
- GH-315 (parent issue)

PR #334, opened 2026-05-26
GH-333: add adversarial trigger-cases for subagent-delegation skill

jcdendrite:
## Summary

- Add `claude/.claude/skills/subagent-delegation/evals/trigger-cases.json` with 7 `description-fidelity` cases for the existing `evals/run_skill_evals.py` harness

## Context

PR #332 added intent-based carve-outs to `subagent-delegation` ("comprehension read feeding your own writing/review/design", "lookup where you already know the target"). The PR reviewer asked to validate that prose against real non-delegation turn shapes from transcript analysis. This PR wires up the measurement.

**Method:** `description-fidelity` (correct for an advisory skill — `runtime` false-zeros advisory skills per `evals/README.md`). Same choice as `test-conventions` and `test-evaluation`.

## Case design

Four adversarial DO-NOT-TRIGGER cases (the load-bearing measurement):

| id | query | carve-out exercised |
|----|-------|---------------------|
| `comprehension-read-feeding-edit` | "Add a docstring to the `parse_config` function in `src/utils.py`." | comprehension read feeding writing/review/design |
| `targeted-lookup-known-path` | "Show me the implementation of `render_widget` in `src/ui/widgets.py`." | lookup where you already know the target |
| `targeted-grep-known-symbol` | "Where is `render_widget` called in this codebase?" | same carve-out, grep flavor |
| `edit-sequence` | "Bump the version field in `package.json` from 1.0.0 to 1.1.0." | Edit/Write sequences stay inline |

Three confirming TRIGGER cases (anchors so a collapsed skill also fails):

| id | query | trigger condition |
|----|-------|------------------|
| `exploratory-search-unknown-target` | "Find every place we render widgets across the codebase." | first exploratory read, target unknown |
| `exploratory-one-command` | "What's in the `utils/` directory?" | exploratory, even single command |
| `full-check-suite` | "Run the full test suite and tell me what's failing." | full check suite |

Cases 3 and 5 use the same `render_widget` family and codebase-grep shape, varying **only** on whether the target is named — isolating target-namedness as the discriminating axis.

**Uncovered TRIGGER condition:** "2nd/3rd Bash command toward same question" — the operational trigger. Worth a follow-up case after this baseline lands and K=30 rates are observed.

## Test plan

- [ ] Run baseline eval locally and paste per-case triggered N/K rates as a comment on this PR:
  ```bash
  python evals/run_skill_evals.py --skill subagent-delegation --samples 30 --verbose
  ```
  Expected: all 7 cases PASS. Any DO-NOT-TRIGGER case triggering ≥50% of samples, or TRIGGER case firing <50%, signals a prose discrimination problem.

### DIFF
--- a/claude/.claude/skills/subagent-delegation/evals/trigger-cases.json
+++ b/claude/.claude/skills/subagent-delegation/evals/trigger-cases.json
@@ -1,0 +1,41 @@
+{
+  "skill_name": "subagent-delegation",
+  "method": "description-fidelity",
+  "cases": [
+    {
+      "id": "comprehension-read-feeding-edit",
+      "query": "Add a docstring to the `parse_config` function in `src/utils.py`.",
+      "should_trigger": false
+    },
+    {
+      "id": "targeted-lookup-known-path",
+      "query": "Show me the implementation of `render_widget` in `src/ui/widgets.py`.",
+      "should_trigger": false
+    },
+    {
+      "id": "targeted-grep-known-symbol",
+      "query": "Where is `render_widget` called in this codebase?",
+      "should_trigger": false
+    },
+    {
+      "id": "edit-sequence",
+      "query": "Bump the version field in `package.json` from 1.0.0 to 1.1.0.",
+      "should_trigger": false
+    },
+    {
+      "id": "exploratory-search-unknown-target",
+      "query": "Find every place we render widgets across the codebase.",
+      "should_trigger": true
+    },
+    {
+      "id": "exploratory-one-command",
+      "query": "What's in the `utils/` directory?",
+      "should_trigger": true
+    },
+    {
+      "id": "full-check-suite",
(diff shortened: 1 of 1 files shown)


jcdendrite:
Closing in favor of a replacement approach. Critical eval surfaced two problems with this case set: (1) `description-fidelity` strips intent context that the rule's discrimination depends on (e.g. 'show me X' can be edit-prep or terminal info-request — same query, opposite delegation call); (2) the prose has a contradiction between 'broad codebase search' TRIGGER and 'single targeted lookup' DO-NOT-TRIGGER for codebase-wide named-symbol greps. Replacement work on GH-333/audit-routing-samples-subcommand adds a per-turn sample-dumper for manual curation of real transcript turns, and folds a SKILL.md prose fix into the same PR.

PR #335, opened 2026-05-26
docs(migration-sync): add ordering-artifact failure guidance to step 5

jcdendrite:
## Summary

- Adds a paragraph to step 5 of the `lovable-cloud-migration-sync` skill explaining what to do when `supabase db reset` fails with an "already exists" error during migration sync verification.
- Names the root cause (local timestamp-ordering artifact where the original runs before the duplicate, not a logic error in the duplicate).
- Specifies the correct fix: delete the original first, then re-run the reset with only the Lovable duplicate present.
- Explicitly prohibits modifying the Lovable duplicate to add `DROP IF EXISTS` guards — the duplicate must exactly match what Lovable Cloud executed; editing it creates repo/production divergence even for a harmless no-op DROP.
- Scopes the existing "do not proceed on FAIL" rule: it applies to logic errors in the duplicate, not to ordering artifacts that disappear once the original is deleted.

## Test plan

- [ ] Skill review checklist applied (all 12 items clean).
- [ ] 1311 existing tests pass (`pytest claude/.claude/`).
- [ ] No private project identifiers, tracker IDs, or structural fingerprints in the added prose.
- [ ] Step numbering, acceptable-differences list, and all other existing content unchanged.

### DIFF
--- a/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
+++ b/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
@@ -111,6 +111,23 @@ reset and harmless — Lovable duplicates run later in timestamp order
 than the originals did, so the reset replays the originals' creation
 against the duplicates' drops.
 
+**Ordering-artifact failures: delete first, then reset.** If `supabase db reset`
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #350, opened 2026-05-28
fix(gitignore): restore .claude/plans/ as tracked so the plan-review gate works as designed

jcdendrite:
## Summary

- Removes `.claude/plans/` from the root `.gitignore` so plan files can be committed alongside their implementation PR.
- Updates the comment above the remaining gitignored entries to explain what stays gitignored and why `.claude/plans/` is now tracked.
- Keeps `claude/.claude/plans/` gitignored — stow-package guard added in GH-257 to prevent `stow --adopt` from re-tracking harness scratch plans into the stow package.

**Root cause:** PR GH-257 (May 17) redesigned the `require-plan-review.sh` gate so committed-clean plan files are treated as historical and don't re-arm the gate in new sessions. The prescribed workflow is `/plan-it → /plan-review → commit-with-implementation`. But the root `.gitignore` line added in GH-24 (April) made all plan files forever untracked — so the gate's "tracked + clean = historical" branch never engaged. Any stale plan file from a prior session permanently armed the gate, particularly acute because sessions are cleared after `/handoff`.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/test_require_plan_review.py` — 29 passed (existing tests cover the committed-clean allow / untracked deny / modified-committed deny branches)
- [x] `git check-ignore .claude/plans/fix-plans-gitignore.md` returns exit 1 (not ignored) ✅
- [x] `git check-ignore claude/.claude/plans/anything.md` matches `.gitignore:13` (stow guard intact) ✅
- [ ] Manual: fresh session in a post-merge clone — write a plan, commit it, open new session, confirm gate does not arm

### DIFF
--- a/.claude/plans/fix-plans-gitignore.md
+++ b/.claude/plans/fix-plans-gitignore.md
@@ -1,0 +1,133 @@
+# Restore commit-with-implementation flow for plan files
+
+## Context
+
+**Goal:** Remove `.claude/plans/` from this repo's root `.gitignore` so plan
+files get committed alongside their implementation, restoring the lifecycle
+the `require-plan-review.sh` gate was designed for in PR #257.
+
+**Problem.** A recurring gate false-arm — a fresh session opens, finds a
+stale plan file from a prior session in `.claude/plans/`, and refuses any
+`Write`/`Edit` until `/plan-review` runs against an unrelated plan. Recent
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #351, opened 2026-05-28
feat(check-runner): dispatch-id contract

jcdendrite:
## Summary

- **Dispatch-id contract.** Parent generates a UUID per check-runner invocation
  and passes it as `dispatch-id: <uuid>` in the dispatch prompt. The agent uses
  it as the spool filename prefix and echoes it back verbatim in the verdict. The
  parent can verify the verdict's origin and locate the spool deterministically
  — a UUID prefix is globally unique, so prior sessions' spools cannot share it.
  Agent refuses dispatches without a dispatch-id (FAIL verdict + explicit message).
- **`subagent-delegation/SKILL.md`** gains the dispatch-id requirement bullet and
  a "Doubting the verdict" recovery paragraph.
- **`ready-for-review/SKILL.md`** dispatch recipe updated with dispatch-id.
- **Incident 6** recorded in `docs/case-studies/check-runner.md` — prose-drift
  verdict + parent re-execution; dispatch-id is the structural fix.
- **`check-runner-bash-guard.sh`** gains stderr logging on emit_deny (one-line
  addition; hook deny categories unchanged from main).
- **Incident 1 stale text in `docs/case-studies/check-runner.md`** updated: hook
  wiring described as "pending" — now reflects live `settings.json` + `agent_type`
  discrimination.
- **Tests:** 1386 passing, ruff clean.

## Design decision

The original branch also added a hook file-read guard covering cat, head, grep, etc.
After review against design-decisions.md §10 ("stacking command-string regexes is
the compounding-defensive-layers tell"), the guard was withdrawn — acknowledged
bypasses in its own header comment confirmed the wrong-foundation signal. Dispatch-id
ships alone.

## Out-of-scope notes

- **Agent-side prose-drift.** Failure-tail-in-context problem documented in
  Incident 6; foundational fix (extend Incident 5 silent-on-success to fail-path)
  is a separate plan.
- **check-runner redesign.** PR review surfaced a broader design question: Sonnet
  vs. Haiku, script-vs-agent evaluation. Separate discussion item.

## Test plan
- [x] 1386 tests pass: pytest claude/.claude/
- [ ] Dispatch check-runner without dispatch-id in prompt → FAIL verdict
- [ ] Normal dispatch with dispatch-id echoes id back in verdict

## Deferred review findings
|---------|---------|-----------------|-----------|
| Dispatch-id verification reasoning-layer only | ciso-reviewer | 3 – Gold-plating | Unenforceable third leg; documented |
| agent_type self-reported | ciso-reviewer | 1 – Orthogonal scope | Harness concern |
| grep -E fork in project-layer reader O(N) | staff-platform-engineer | 1 – Orthogonal scope | Pre-existing |
| No jq timeout guards | staff-platform-engineer | 1 – Orthogonal scope | Pre-existing |
| "often turns out" hedged prose in subagent-delegation/SKILL.md | skill-review | 3 – Gold-plating | Sentence provides WHY context motivating spool inspection; imperative rewrite loses explanatory value |
| check-runner.md missing TRIGGER/DO NOT TRIGGER blocks | agent-review | 1 – Orthogonal scope | Pre-existing gap; executor-style agents use summary descriptions |
| check-runner.md description omits dispatch-id/working-directory pre-conditions | agent-review | 3 – Gold-plating | Intentionally summary-level; body is the authoritative spec for callers |


claude/.claude/hooks/check-runner-bash-guard.sh
```diff
@@ -22,15 +22,28 @@
...
 # The agent's only legitimate write is its spool file under
 # ${TMPDIR:-/tmp}/; everything else is reportable as a verdict, not
 # resolvable by mutation.
+# Known gaps not closed by this hook:
+#   - Shell quoting variants (e.g. c''at file) pass the \b-anchored deny
+#     patterns. The threat model is prose-drift prevention by a good-faith
+#     model; adversarial quoting bypasses are out-of-scope for this hook.
+#   - Scripting-language one-liners (python3 -c, node -e, ruby -e) are not
+#     denied. The check-runner's enumerated-commands-only posture is the
+#     primary control for this surface.
```
jcdendrite: This is over-engineered. Remember that line in CLAUDE.md about overcomplication? 

Let's think of what we need check-runner to do, just run a simple command according to what the parent specified. Is there a way that we can encode what we need it to do in a script that the parent can run? Or do we just abandon check-runner because of how Claude works when output generates more than 2 KB (if I recall correctly - should be in REFERENCES or design-docs or something). 

We may want to redesign this. The UUID makes sense, but I almost wonder if it would be more effective to use Sonnet here and have it focus on debugging checks and tests. Definitely want to rethink the design.

jcdendrite (commented):
Why not just fix this? It's a simple change:
> Incident 1 stale text. docs/case-studies/check-runner.md Incident 1 still describes check-runner-bash-guard.sh as a "reference implementation pending agent-scoped hook wiring" — the wiring now exists via settings.json agent_type discrimination. Cleanup is a separate PR.

claude/.claude/hooks/check-runner-bash-guard.sh
jcdendrite: **[Claude Code]** The hook read-guard (the over-engineered part) was withdrawn in the latest commit on this branch — same call you are making here. What remains in the diff is the dispatch-id contract: a UUID the parent generates per invocation, passed in the prompt, echoed back in the verdict. One field addition; no regex stack.

The 2 KB concern: the Re-grounding section in `docs/case-studies/check-runner.md` has the measurement — the actual harness truncation is 30 KB, already in place, so the original context-flood premise does not hold. The current justification is signal quality and verdict ergonomics (curated per-command pass/fail + failure tail, handled on Haiku).

The broader design questions you raise — Sonnet vs. Haiku, script-vs-agent approach, whether check-runner is worth keeping — are tracked in #352.

`disposition: deferred` | `deferral-reason: over-engineering concern already addressed by hook removal; broader redesign tracked in #352`

jcdendrite:
**[Claude Code]** Fixed. Updated both stale passages in Incident 1 (line 15) and the Sources section (line 84) of `docs/case-studies/check-runner.md` to reflect that the hook is now wired via `settings.json` `PreToolUse(Bash)` with internal `agent_type: check-runner` discrimination. Addressed in the latest commit on this branch.

`disposition: fixed-as-requested` | `commit: latest on this branch`

jcdendrite (commented):
changes lgtm

PR #353, opened 2026-05-28
perf(check-runner-bash-guard): replace grep-fork loops with bash builtin regex

jcdendrite:
## Summary

- Replaces two `grep -E` subprocess-fork loops in `check-runner-bash-guard.sh` with bash builtin `[[ =~ ]]` — zero forks, same POSIX ERE semantics
- Validation loop (lines 117–136): was forking `printf + grep` per project-layer pattern on every Bash hook invocation to test regex compile; now uses `{ [[ "" =~ $line ]]; } 2>/dev/null`
- Matching loop (lines 159–167): was O(N×M) forks per invocation; now uses `[[ "$fragment" =~ $pattern ]]`, matching the shape of the existing global-pattern arm at line 152
- Comments at both sites document the unquoted-RHS requirement (`[[` context makes word-splitting safe; literal-match warning for `[ ]` / `test` contexts) and the bash >= 3.2 exit-code assumption

## Context

Deferred from PR #351 (dispatch-id contract for check-runner) as orthogonal scope. This is PR A of two; PR B (guard-hook fail-closed alignment + jq timeout, ~11 files) is planned separately.

## Test plan

- [x] `pytest claude/.claude/` — 1386 passed, including `test_malformed_regex_logs_and_skips_line` which asserts the stderr message shape is preserved
- [ ] Manual smoke: create `.claude/check-runner-deny-patterns.txt` with one valid ERE and one malformed pattern; confirm hook logs "skipping malformed regex" for the bad line and denies matching commands from the valid line

### DIFF
--- a/claude/.claude/hooks/check-runner-bash-guard.sh
+++ b/claude/.claude/hooks/check-runner-bash-guard.sh
@@ -122,10 +122,16 @@ if [ -e "$PROJECT_PATTERNS_FILE" ]; then
       line="${line%"${line##*[![:space:]]}"}"
       [ -z "$line" ] && continue
       case "$line" in '#'*) continue ;; esac
-      # Validate the regex compiles before adding it. grep exits 0 on
-      # match, 1 on no-match (valid regex), 2 on regex-parse error —
-      # use exit==2 to mean malformed, not exit!=0.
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #355, opened 2026-05-29
feat(hooks): fail-closed alignment + jq timeout across all guard hooks

jcdendrite:
## Summary

- **Security gap closed:** 10 of the gate-class hooks were fail-open on jq parse failure — malformed or absent JSON produced an empty COMMAND, matched no denylist pattern, and silently allowed the blocked operation.
- Adds `_lib_jq` (5s timeout backstop) + `_lib_parse_tool_input_or_deny` to `_lib.sh` (main + plugin mirror); all gate hooks adopt the helper.
- Uses ASCII Unit Separator (0x1f) as field delimiter instead of newlines, preventing a crafted `tool_name` with embedded newline from corrupting COMMAND (newline injection fix from CISO review).
- Adds suite-wide alignment enforcement: `test_hook_alignment.py` (Layer 1 static + Layer 2 behavior, including static emit_deny ordering check), `test_lib.py` (unit tests including null literal, newline injection, hung jq), `test_install_sh_timeout_warning.py`.
- Updates `plugins/claude-hook-review/skills/claude-hook-review/SKILL.md` to document the canonical pattern, three deny paths, `# hook-class:` headers, and post-validation jq guidance.
- Follow-up commits address reviewer comments: removed stale numeric counts from SKILL.md, clarified `# hook-class: informational` = hardening posture (not functional importance), closed test gaps identified in code review.

## Security model

Three deny paths protect against silent-allow:
- **(a) jq non-zero exit** — parse failure, `timeout` exit=124, or missing jq binary; also fires on structural-type errors (`.tool_input` not an object).
- **(b) empty INPUT** — stdin EOF, closed pipe, or harness misbehavior.
- **(c) empty TOOL_NAME** — valid JSON but `.tool_name` absent (e.g. `{}`).

`emit_deny` now pre-assembles the JSON payload before `printf` (atomic write). `# hook-class: gate|informational` header added to every hook.

## Design decisions

- D1: Hybrid — helper in `_lib.sh`, `emit_deny` inline per hook.
- D2: 5s timeout, matching `git_capped` precedent in `guard-settings-session-keys.sh`.
- D3: Uniform across all gate hooks.
- D4: `# hook-class:` header + filename-prefix enforcement in `test_hook_alignment.py`.

## Incidental edits

`# hook-class: informational` added to 7 informational hooks, `check-branch-divergence.sh`, and `provision-validator-venv.sh` — required by the alignment test's exhaustive hook enumeration.

## Test plan

- [ ] Review `_lib.sh` helper logic (three deny paths, unit separator delimiter, newline injection guard)
- [ ] Review `test_hook_alignment.py` Layer 1 (static, including emit_deny ordering) + Layer 2 (behavior) coverage
- [ ] Review `test_lib.py` boundary cases (null literal, newline injection, hung jq, absent timeout)
- [ ] Spot-check 2-3 gate hooks for correct `_lib_parse_tool_input_or_deny` adoption
- [ ] Verify `# hook-class:` headers present on all changed hooks

## Deferred review findings

| 0x1f byte in `tool_input.command` truncates COMMAND at delimiter split (field-confusion, not silent-allow) | ciso-reviewer + staff-sdet | Orthogonal scope | Requires prompt injection through Claude's Bash tool generation — distinct attack surface from the fail-closed parse-failure scope of this PR; no silent-allow path exists |
| Lone-newline stdin takes path (b) rather than path (a) — untested as a distinct execution path | staff-sdet | Edge case below current scale | Different code path, same deny result; no bypass risk at operating scale |
| BSD/macOS bare-jq fallback (no timeout bound) when coreutils absent | ciso-reviewer + staff-backend-engineer | Gold-plating beyond declared user surface | Post-exploitation scenario; documented in code comments + install.sh warning; accepted tradeoff for developer-workflow tool |

### DIFF
--- a/plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
+++ b/plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
@@ -32,7 +32,7 @@ Rejected forms: bare `./<rest>`, `../<rest>`, or unprefixed script names. Concre
 
 ## 3. Script skeleton
 
-The idiomatic pattern from `deny-private-project-refs.sh` and `require-stow-reminder.sh`:
+New gate hooks must follow the canonical pattern in Section 4 (`_lib_parse_tool_input_or_deny`) rather than reimplementing this inline form. The structural conventions — `set -uo pipefail`, `emit_deny` defined before sourcing `_lib.sh`, exit-0 contract — still apply:
 
 ```bash
 #!/bin/bash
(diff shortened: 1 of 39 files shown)


plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
```diff
@@ -42,38 +42,66 @@ set -uo pipefail   # explicit failure modes: unbound vars fail loudly,
...
+- After `_lib_parse_tool_input_or_deny`, `COMMAND` may still be empty: `jq -r '... // empty'` exits **0** on missing fields (empty string, not non-zero). For allow-list-style gates (deny anything outside a known set), guard on `[ -z "$COMMAND" ]`. For deny-list-style gates (deny specific operations, allow all else), empty `COMMAND` correctly falls through.
 
 ## 4. Fail-open vs fail-closed posture
 
-**Fail-closed** when the gate prevents a leak: parse failure → deny (a hook that can't read its input can't verify the operation is safe). Example: `deny-private-project-refs.sh:101–104`.
+**Fail-closed** when the gate prevents a leak: parse failure → deny (a hook that can't read its input can't verify the operation is safe). All 21 gate-class hooks use `_lib.sh::_lib_parse_tool_input_or_deny` as the canonical fail-closed pattern — source that function rather than reimplementing the inline JQ_EXIT check.
```
jcdendrite: `All 21 gate-class hooks` -- Numbers get stale. Delete the number. "All gate-class hooks" is perfectly fine.

plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
```diff
@@ -42,38 +42,66 @@ set -uo pipefail   # explicit failure modes: unbound vars fail loudly,
...
 
 **Fail-open** when the gate is advisory and absence of input is normal: if `~/.claude/private-projects.md` is missing, allow normally (`deny-private-project-refs.sh:259–260`).
 
 State the chosen posture in the script header. Reviewers shouldn't have to re-derive it from the code.
 
+**`# hook-class:` header** — every hook must declare its class on the second line:
+- `# hook-class: gate` — fires PreToolUse and may deny. Required on all 21 guard hooks.
+- `# hook-class: informational` — fires PostToolUse/SessionStart/etc. and never denies.
+
+Flipping a marker from `gate` to `informational` is a security-class change that removes a deny path; the commit message must state the rationale.
```
jcdendrite, line 103: Is there another category than "informational" - some of the hooks that you've marked as informational are essential to the workflow rather than just providing useful "information". Others are "advisory" to the agent. I think "informational" refers to a distinct and unified category of hook from the perspective of hook hardening implementation per this PR, but you may want to explain more what you mean by "informational". My hunch is that this is just a documentation gap.

jcdendrite (commented):
Meta-comment: On timeout, does the hook print a message to the caller? It should probably tell the caller what happened and to inform the human or retry, depending on the type of hook. For example, PreToolUse like capture-session-id if it fails will result in an irrecoverable situation so human might want to restart the session, whereas others like marker creation or deletion could just be retried.

plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Removed the hardcoded count from both occurrences in §4 ("All gate-class hooks use...") and the `# hook-class:` block ("Required on all guard hooks."). Addressed in the latest commit on this branch.

plugins/claude-hook-review/skills/claude-hook-review/SKILL.md
jcdendrite: **[Claude Code]** Good catch — this is a documentation gap. The label describes the hardening posture (whether the hook can issue a PreToolUse denial), not the hook's functional importance. I added a clarifying sentence to the `# hook-class: informational` bullet:

> "The label describes the hardening posture (cannot deny), not functional importance — an essential workflow hook (e.g., `capture-session-id.sh`) and a purely advisory one both carry this label if neither issues a PreToolUse denial."

Both categories map to `informational` because the hardening work here (fail-closed parsing, `_lib_parse_tool_input_or_deny`, `# hook-class:` enforcement) applies only to PreToolUse gate hooks. Non-gate hooks that run at SessionStart/PostToolUse cannot deny by contract, so they are in a separate hardening class regardless of how important they are to the workflow. Addressed in the latest commit on this branch.

jcdendrite:
**[Claude Code]** Responding to the review meta-comment on timeout behavior:

**For gate hooks (PreToolUse):** On jq timeout or parse failure, `_lib_parse_tool_input_or_deny` calls `emit_deny` with the caller-supplied message (e.g., "Blocked by gate: could not parse tool-input JSON."). This is written to stdout as `hookSpecificOutput.permissionDecisionReason`, which Claude reads and surfaces to the user — so the caller does receive a message, though it does not distinguish "timeout" from "parse failure" (both mean "cannot verify safety → deny").

**For `capture-session-id.sh` specifically:** This hook does not use `_lib_jq` — it calls bare `jq` without the 5s timeout wrapper. Its failure modes are handled differently: each failure path emits a `[capture-session-id] ...` diagnostic to stderr (visible in the terminal), and the `/respond-pr` skill's Step 0 fails loudly with a clear message when the lookup file was not written.

**On "retry vs restart" guidance:** Gate hooks deny with a reason Claude can relay to the user. Claude can decide whether to retry (for what looks transient) or report to the user (for persistent failures). For `capture-session-id.sh`, each stderr message already ends with "respond-pr skill will fail at Step 0" — explicit guidance that the session state is missing and the user should check the hook's output.

The distinction between timeout and parse failure is intentionally collapsed in gate hooks: both mean "can't verify safety → deny," and the operational response is the same. This is documented in the PR description's deferred findings section (BSD/macOS timeout fallback is noted as an accepted tradeoff for dev-workflow tools).

`disposition: explain-design`

jcdendrite (commented):
changes lgtm

PR #358, opened 2026-05-29
feat(transcript-analysis): add skill-invocation subcommand

jcdendrite:
## Summary

- Adds `skill-invocation` subcommand to `transcript-analysis.py` — tallies per-skill invocation counts across three buckets: top-level (description-driven auto-trigger), routed (fired while another skill's body was active via `attributionSkill`), and user-slash (`<command-name>` tag in user records)
- Reports routed pairs and a classification summary (load-bearing, routed-only, slash-only candidates) useful for skill-description budget analysis
- 20 tests covering all three buckets, sidechain exclusion (assistant and user records), edge cases (empty skill field, multi-block records), project filtering, cross-project aggregation, and mixed-session invariants
- Adds `skill-invocation` reference section to `docs/transcript-analysis.md` with purpose, flags, sample output, and when-to-reach-for-it guidance

## Test plan

- [ ] Run `transcript-analysis.py skill-invocation` against real corpus and verify the three columns and classification summary look plausible
- [ ] Confirm routed-only candidates list matches known dispatcher-only skill invocations

## Deferred review findings

| Output-parsing via `split()[N]` in test assertions is format-coupled — column-width or column-order changes break assertions simultaneously rather than the counting logic they intend to cover | staff-sdet | Orthogonal scope | Format coupling is a test-maintenance concern; counting correctness is covered by the tests; restructuring assertions to read module-level dicts instead of formatted strings is disproportionate to the regression risk |

### DIFF
--- a/claude/.claude/scripts/tests/test_transcript_analysis.py
+++ b/claude/.claude/scripts/tests/test_transcript_analysis.py
@@ -2943,8 +2943,10 @@ class TestSkillInvocation:
         asst_rec["attributionSkill"] = "ready-for-review"
         _write_jsonl(fake_projects / "s2.jsonl", [asst_rec])
         out = self._run(capsys)
-        # Routed pair must appear in the ROUTED PAIRS section
-        assert "ready-for-review -> code-review" in out
+        # Routed pair must appear in the ROUTED PAIRS section with count 1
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #359, opened 2026-05-29
feat(skills): convert 4 utility skills to skillOverrides: name-only

jcdendrite:
## Summary

- Remove `disable-model-invocation: true` from `brief`, `handoff`, `read-docx-comments`, and `transcript-analysis` — the four former slash-only command skills now use `skillOverrides: name-only` in `settings.json` instead
- Add `name-only` entries to `settings.json` `skillOverrides` block for all four skills
- Expand `test_skills.py` to model the new invocation category: `_settings_skill_overrides()` + `_name_only_skills()` helpers, `TestNameOnlySkillContracts` class (positive SKILL.md existence + no-disable-flag tests), and `_model_invokable_skills()` updated to exclude name-only from TRIGGER-block enforcement
- Add "Skills available by name" section and four-value override reference table to `docs/skills.md`

**Why:** `skillOverrides: name-only` (v2.1.129+) lets the model invoke these skills by exact name without loading their descriptions into the always-on listing budget. The previous `disable-model-invocation: true` flag prevented model invocation entirely; `name-only` enables name-based invocation while keeping descriptions out of budget — a strictly better tradeoff for skills that are useful in mid-session but not worth auto-trigger description cost.

## Test plan

- [x] `../../../.venv/bin/pytest claude/.claude/` passes (verified in this PR: 1628+ passed)
- [x] `../../../.venv/bin/ruff check claude/.claude/` passes (verified in this PR: clean)
- [ ] In a Claude Code session (v2.1.129+): invoke `/brief`, `/handoff`, `/read-docx-comments`, `/transcript-analysis` by name — all should fire without description in the always-loaded listing
- [ ] Confirm `/skills` menu still shows all four skills (they remain slash-invocable via `name-only`)

## Deferred review findings

|---|---|---|---|
| `COMMAND_SKILLS = []` produces skipped parametrize tests; a new skill with `disable-model-invocation: true` not added to `COMMAND_SKILLS` would fail silently | staff-sdet Q1 (cumulative review) | Orthogonal scope | No command skills exist today; a forward-looking guard for a category with zero current members is outside this PR's scope |

### DIFF
--- a/claude/.claude/settings.json
+++ b/claude/.claude/settings.json
@@ -39,6 +39,10 @@
   },
   "model": "opusplan",
   "skillOverrides": {
+    "brief": "name-only",
+    "handoff": "name-only",
+    "read-docx-comments": "name-only",
+    "transcript-analysis": "name-only",
     "claude-api": "off",
     "fewer-permission-prompts": "off",
     "init": "off",
--- a/claude/.claude/skills/brief/SKILL.md
+++ b/claude/.claude/skills/brief/SKILL.md
@@ -1,7 +1,6 @@
 ---
 name: brief
 description: Produce a cold-start task briefing at /tmp/<slug>-task.md for a fresh session to pick up known, well-scoped work (abandoned PR, surfaced follow-up, ticket whose scope is settled). Distinct from /handoff, which captures mid-flight session state.
-disable-model-invocation: true
 ---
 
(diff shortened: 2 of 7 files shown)


jcdendrite (commented):
lgtm

jcdendrite (commented):
lgtm

jcdendrite (commented):
lgtm

Issue #363, opened 2026-05-29
test: replace positional column-index parsing in TestSkillPair with header-anchored assertions


jcdendrite:
## Problem

`TestSkillPair` (and the new `TestSkillPairSubagentFile`) use positional `cols[N]` splits on `cmd_skill_pair` output rows:

```python
cols = lines[0].split()
assert cols[1] == "1"   # Lead=1
assert cols[2] == "0"   # Main=0
assert cols[3] == "1"   # Side=1
```

This is brittle: a whitespace/formatting change to the output table shifts column indices and breaks tests without any logic change. Conversely, a logic bug that shifts a count to the wrong column produces a failure that looks like a formatting regression.

## Suggested fix

Parse column positions from the header row first, then index by name:

```python
header = next(ln for ln in out.splitlines() if "Lead" in ln)
cols_header = header.split()
lead_idx = cols_header.index("Lead")
...
data_cols = lines[0].split()
assert data_cols[lead_idx] == "1"
```

Or use a regex anchored to the column label pattern from the output format string.

## Scope

This is a test-only refactor. The pattern exists across ~6 existing `TestSkillPair` methods — a suite-wide update keeps the style consistent. Surfaced during code review of #fix-transcript-subagent-split-format.

PR #364, opened 2026-05-29
fix(transcript-analysis): read split subagents/ files in iter_sessions

jcdendrite:
## Summary

- `iter_sessions` never descended into `<session>/subagents/*.jsonl`, so `subagents` and `skill-pair` reported 0 sidechain turns — a trusted-but-wrong zero that caused a real post-hoc analysis to conclude no subagent check-runners were dispatched.
- Fix: add `include_subagents=False` opt-in to `iter_sessions`; the two sidechain consumers pass `True`. All 15 other callers keep the default (zero blast radius).
- Adds a corpus-wide format-drift canary in both consumers: warns to stderr when main-thread Agent/Task spawns are present but zero isSidechain turns were read, so the next format change is caught at point-of-use rather than producing another silent trusted zero.
- Two new module-scope constants (`SUBAGENT_SUBDIR`, `_SPAWN_TOOL_NAMES`) single-source the discriminators shared between the reader, the canary, and `cmd_subagent_mix`.

## Test plan

- [x] `../../../.venv/bin/pytest claude/.claude/` — 223 tests pass
- [x] `../../../.venv/bin/ruff check claude/.claude/` — clean
- [x] End-to-end: `day-285-sentry-edge-pr1` now shows 25 Sonnet + 156 Haiku sidechain turns (was 0). `fail-seq` and `struggle` output unchanged.
- [x] Canary smoke check: clean corpus emits no warning; corpus with Agent spawns but no subagent files emits WARNING to stderr, not stdout.

## Notes for reviewer

- Constants `SUBAGENT_SUBDIR` and `_SPAWN_TOOL_NAMES` are defined at ~line 667 (near `REVIEW_SKILLS`) but first referenced in functions earlier in the file. Python resolves module-level names at call time, so there is no runtime issue — but hoisting them before `iter_sessions` is a cosmetic follow-up (GH-363 already tracks the `TestSkillPair` column-index counterpart).
- A full sweep of the ~14 `isSidechain`/`gitBranch` string literals to named constants was explicitly deferred; introducing only the two discriminators the new code and canary share avoids a partial spelling where neither form is authoritative.

### DIFF
--- a/.claude/plans/fix-transcript-subagent-split-format.md
+++ b/.claude/plans/fix-transcript-subagent-split-format.md
@@ -1,0 +1,247 @@
+# Fix `transcript-analysis.py` subagent false-negative (split-transcript format)
+
+## Context
+
+**Goal:** Make `transcript-analysis.py subagents` (and the sibling `skill-followers`
+metric) correctly count turns that ran inside named subagents — check-runner,
+Explore, `staff-*` reviewers — instead of reporting zero.
+
+**Problem & why now:** Running
+`transcript-analysis.py subagents --branches day-285-sentry-edge-pr1` reports
+**0 subagent turns / 208 main-thread turns**, even though that branch dispatched
+many subagents. A real post-hoc analysis trusted this output and wrongly concluded
+no check-runner agents were ever dispatched. The script is stowed from this repo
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #368, opened 2026-06-02
feat: extend file-output canary to all 8 reviewer agents

jcdendrite:
## Summary

- Extend `Write` tool + `### File-based output` section from `staff-backend-engineer` to all seven remaining reviewer agents (`ciso-reviewer`, `staff-analytics-engineer`, `staff-data-engineer`, `staff-frontend-engineer`, `staff-platform-engineer`, `staff-product-engineer`, `staff-sdet`)
- Generalize `code-review/SKILL.md` dispatcher to pass `findings_path` to every reviewer (previously `staff-backend-engineer` only)
- Exempt `agent-reviews/` writes from `require-plan-review.sh` so reviewer finds never fall back to inline during a plan-in-flight session (15% fallback rate observed; caused by gate blocking `Write`)
- New `test_agent_roster.py`: 5 tests enforcing the canary convention — roster completeness, Write tool presence, block presence, uncategorized-agent inverse scan (`NON_REVIEWER_AGENTS`), and sorted-list invariant (anchors drift-guard canonical)
- `TestFileBasedOutputBlockConsistency`: byte-equivalence test asserting all 8 agents carry an identical `### File-based output` block (agent name normalized before comparison); drift becomes a CI failure. Agent `@path` import is CLAUDE.md-only (not available in agent bodies — Anthropic docs confirmed), so intentional duplication + test enforcement is the canonical SSOT approach for this repo.
- 6 new tests in `test_require_plan_review.py` for the `agent-reviews/` exemption boundary (Write/Edit/MultiEdit × allow/deny paths, including directory-itself deny); fixed `test_missing_file_path_denies` docstring to describe actual hook control-flow (empty TARGET_PATH skips scope filter → default deny)
- Document the mechanism in `docs/design-decisions.md §12`; update README reviewer roster note
- Fix `code-writer.md` comment in `NON_REVIEWER_AGENTS`: "does not review" → "self-reviews its own output, not a dispatcher-spawned reviewer"
- Move new-reviewer onboarding instruction from `code-review/SKILL.md` to `agent-review/SKILL.md` item 15 — code-review fires during review sessions, not when a new agent is added; the enforcement point belongs in agent-review
- `docs/design-decisions.md §12`: add ~86–92% inline context reduction figure and note that findings files remain readable by human reviewers in `agent-reviews/` without re-running the review

Estimated saving: 32–52 KB per `/ready-for-review` cycle in the Opus orchestrator window (6×–11× reduction vs inline findings per reviewer), based on empirical scan of 1,496 transcripts.

## Test plan

- [x] `pytest claude/.claude/` — all tests pass
- [ ] Smoke test: on a branch with a staged plan file, run `/code-review` so the constellation spawns; confirm each reviewer writes to `agent-reviews/` and returns a pointer line (no full-inline fallback), and that the plan-review gate does not block the writes
- [ ] Empirical re-check after some real dispatches: re-run the compliance scan (`iterate ~/.claude/projects/*/*.jsonl`, match `staff-*/ciso-reviewer` `findings_path` dispatches to their `tool_result` byte sizes) and confirm fallback rate drops

## Ordering rule (for reviewer)

When a new reviewer agent is added to the roster, the `agent-review` checklist (item 15 — reviewer-agent output contract) is the enforcement point: verify the agent carries `Write` + `### File-based output` before dispatching it with `findings_path`. The `test_no_uncategorized_agents` test enforces that any new agent is explicitly categorized in `REVIEWER_AGENTS` or `NON_REVIEWER_AGENTS`.

## Deferred review findings

| Deny-path tests in `test_require_plan_review.py` declare `plan_review_home` fixture but don't use it (harmless overhead) | staff-sdet | Orthogonal scope | Tests produce correct verdicts with or without the fixture; no enforcement gap. Only the deny-path tests are affected (not the new tests added in this follow-up). |


claude/.claude/agents/ciso-reviewer.md
```diff
@@ -74,3 +76,35 @@ For each finding:
...
+   parent reads them from the file. Including full findings inline when
+   `findings_path` is present is a defect.
+   If the dispatch prompt poses specific questions, answer them inside the
+   findings file (e.g. under an `## Answers` heading) — not in the inline
+   return. The inline summary stays one sentence regardless of how many
+   questions the prompt asks.
+   **If the Write call fails**, do not report success. Instead, state the failure
+   explicitly and fall back to the **Inline output** format.
+
+When `findings_path` is absent, ignore this section and use the **Inline output** format.
```
jcdendrite, line 110: This text is duplicated across all agents. It is indeed load bearing, but is it preferable to pull it from a source-of-truth file? What do Anthropic docs say? Do these agents files respect imported markdown? I wouldn't be worried so much about text duplicated across two files, but all seven is a recipe for drift, and I could see us iterating on the instructions here.

jcdendrite (commented):
These findings in the PR description should not be deferred:
- "No test for write to agent-reviews/ directory itself (no filename) — hook glob agent-reviews/* excludes this edge". That's an easy test.
- "test_missing_file_path_denies docstring describes deny as "can't be resolved" but actual mechanism is empty-TARGET_PATH bypasses scope filter" - That's an easy fix.

claude/.claude/agents/ciso-reviewer.md
jcdendrite: **[Claude Code]** Checked the [Anthropic docs on CLAUDE.md imports](https://code.claude.com/docs/en/memory#import-additional-files): the `@path` syntax is documented for CLAUDE.md files only. Agent files (`.claude/agents/*.md`) are loaded as the subagent's system prompt and have no equivalent import mechanism — there is no way to pull from a single source file.

The duplication is intentional and the only option given that constraint. The drift risk it creates is now addressed by `TestFileBasedOutputBlockConsistency` (added in `cdd7723`), which extracts the `### File-based output` block from all 8 agents, normalizes the per-agent H1 name line, and asserts byte-equivalence. Any future edit that diverges one agent's copy from the others will fail the test suite before it reaches CI.

jcdendrite:
**[Claude Code]** Both items are addressed in the latest commits on this branch:

- **Directory-itself deny test** — `test_agent_reviews_directory_itself_not_exempt` added in `cdd7723`. Pins that a bare `agent-reviews/` path with no filename fails the `agent-reviews/*` glob and is denied, guarding against future glob-widening refactors.
- **`test_missing_file_path_denies` docstring** — fixed in `cdd7723`. Replaced vague "fail-closed" language with the precise shell code path: empty `TARGET_PATH` causes the scope-filter guard (`if [ -n "$TARGET_PATH" ]`) to be false, falling through to `emit_deny`.

CI passed on these commits.

claude/.claude/hooks/tests/test_agent_roster.py
```diff
@@ -0,0 +1,214 @@
...
+    "staff-sdet.md",
+]
+
+# Agents that exist in the directory but are not code-review dispatched
+# reviewers — they do not receive findings_path and do not need the canary.
+# When a new non-reviewer agent is added, add it here; the
+# test_no_uncategorized_agents test will fail until it is categorized.
+NON_REVIEWER_AGENTS = [
+    "check-runner.md",  # executor-style; dispatches check suites, does not review
+    "code-writer.md",   # implementer; writes code, does not review
```
jcdendrite: Misleading comment. The code-write agent does review actually - it reviews its own code, but it is not spawned by the parent thread as an official reviewer.

claude/.claude/skills/code-review/SKILL.md
```diff
@@ -247,7 +247,7 @@ Report every matched row's verdict via the **Spawn decisions:** line in the *Out
...
...ne-sentence summary, finding count. After it returns: if the reviewer reported a successful write, `Read` the findings file — Recommendations section first, full file when count > 0; if the reviewer reported a write failure and fell back to inline findings, use those inline findings directly instead. Before the first spawn, add `agent-reviews/` to `$(git rev-parse --git-dir)/info/exclude` idempotently (grep-check before appending) so findings files can't be accidentally staged; cleanup is automatic with the worktree. The `agent-reviews/` directory does not need to be pre-created — the reviewer's `Write` call creates it. All reviewers now carry the `Write` tool and the file-based-output section; the file-based output section activates only when `findings_path` is present in the prompt, so reviewers that have not been passed a `findings_path` continue to return findings inline. If a new reviewer agent is added to the roster, add `Write` + the `### File-based output` section to the agent *before* updating this dispatcher paragraph — passing `findings_path` to an agent without the file-based-output section re-arms the heredoc-abort-on-large-findings bug that the canary exists to avoid.
```
jcdendrite: > If a new reviewer agent is added to the roster, add `Write` + the `### File-based output` section to the agent *before* updating this dispatcher paragraph — passing `findings_path` to an agent without the file-based-output section re-arms the heredoc-abort-on-large-findings bug that the canary exists to avoid.

This text needs to be adjusted. The code review skill isn't going to know if a new agent is added. The check for a reviewer agent to include File-based output should be added to the agent-review skill instead.

docs/design-decisions.md
```diff
@@ -117,6 +117,10 @@ and to the `code-review` skill's umbrella term — not a persona job title.
...
 
+## 12. Reviewer file-based output via `findings_path`
+
+All eight reviewer agents write structured Markdown findings to `agent-reviews/<agent-name>-<epoch>-<slug>.md` when the dispatch prompt includes `findings_path:`, and return only a pointer line inline (~650 B median vs ~4,500–7,800 B for full inline findings). The parent reads the file after the reviewer returns, starting with the `## Recommendations` section. Each reviewer carries the `Write` tool and a `### File-based output` section in its `## Output format`; the section activates only when `findings_path` is present, so reviewers dispatched without it continue to return findings inline. The `agent-reviews/` directory is gitignored and created on first write by the reviewer's `Write` call — no `mkdir` step is needed in the dispatcher. The plan-review gate (`require-plan-review.sh`) exempts writes to `<repo>/agent-reviews/*` with an exact prefix match, so reviewer writes are not blocked when a plan is in flight — without the exemption, a reviewer dispatched during code-review would fall back to full-inline output, defeating the context savings the mechanism exists to provide.
```
jcdendrite: Include some statistics on percentage context savings - the specific numbers / data points are helpful for explaining the rationale.

jcdendrite: Another benefit is discoverability for the human reviewer. If they really want to, then can read specialist reviews in the agent-reviews folder.

claude/.claude/hooks/tests/test_agent_roster.py
jcdendrite: **[Claude Code]** Fixed. Updated the comment to: `# implementer; self-reviews its own output, not a dispatcher-spawned reviewer`. Addressed in the latest commit on this branch.

claude/.claude/skills/code-review/SKILL.md
jcdendrite: **[Claude Code]** Fixed. Moved the instruction out of `code-review/SKILL.md` (which fires during review sessions, not when adding agents) and into `agent-review/SKILL.md` as item 15 — "Reviewer-agent output contract" — which fires exactly when someone edits a reviewer agent file. The `code-review/SKILL.md` paragraph now cross-references item 15 as the enforcement point. Addressed in the latest commit on this branch.

docs/design-decisions.md
jcdendrite: **[Claude Code]** Added: ~650 B / ~4,500–7,800 B = ~86–92% inline context reduction per reviewer. Added in the latest commit on this branch.

jcdendrite: **[Claude Code]** Added a sentence to §12: "The findings files also remain readable in `agent-reviews/` for the duration of the worktree, giving human reviewers direct access to each specialist's full analysis without re-running the review." Added in the latest commit on this branch.

jcdendrite (commented):
lgtm

Issue #367, opened 2026-06-02
macOS bash 3.2 incompatibility: declare -A / mapfile break two scripts and silently disable the handoff-nudge hook


jcdendrite:
## Summary

macOS ships **bash 3.2.57** at `/bin/bash` (Apple froze it at the last GPLv2 release in 2007). Three shell files in this repo use **bash 4.0+ features** (`declare -A` associative arrays, `mapfile`), causing hard errors for the scripts and a silent failure for the hook.

Reproduction (stock macOS, no Homebrew bash):
```
$ ~/.claude/scripts/update-claude-config-plugins.sh
Refreshing claude-config marketplace...
.../update-claude-config-plugins.sh: line 100: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

CI runs on Ubuntu (bash 5) and never caught this. Local `pytest` for the affected scripts also fails on macOS for the same reason.

## Affected files

| File | Bash-4 feature | Lines | Impact on macOS |
|------|----------------|-------|-----------------|
| `claude/.claude/scripts/update-claude-config-plugins.sh` | `declare -A` ×2 | 100–101 (decl), 121–122, 164, 227 (use) | Hard error, exits immediately |
| `claude/.claude/scripts/cleanup-merged-branches.sh` | `mapfile` ×2, `declare -A` ×3 | 197, 202–203, 401 (decl/mapfile); 249–250, 257–258, 274 (use) | Hard error |
| `claude/.claude/hooks/nudge-handoff-near-context-cap.sh` | `mapfile` ×1 | 33 | **Silent no-op** — fail-open guards swallow it, so the handoff nudge never fires on macOS |

The hook is the most insidious because it doesn't visibly fail — the `|| true` fail-open guard swallows the error, and the nudge simply never fires for macOS users.

## Fix approach

**1. Portable rewrite of all three files (bash 3.2 compatible):**

- `mapfile -t arr < <(cmd)` → `arr=(); while IFS= read -r line; do arr+=("$line"); done < <(cmd)`
- For the hook's 4-field case, prefer direct sequential reads over an intermediate array:
  ```bash
  IFS= read -r SESSION_ID
  IFS= read -r AGENT_TYPE
  IFS= read -r PERMISSION_MODE
  IFS= read -r TRANSCRIPT_PATH
  ```
[...]

PR #369, opened 2026-06-02
GH-367: bash 3.2 portability for scripts and nudge hook

jcdendrite:
## Summary

- Replaces `declare -A`, `mapfile`, and `sort -V` (all bash-4+/GNU-only) with bash-3.2-portable equivalents in `cleanup-merged-branches.sh`, `update-claude-config-plugins.sh`, and `nudge-handoff-near-context-cap.sh`
- The nudge hook's silent macOS failure was the most insidious: `|| true` swallowed the `mapfile` error, so the handoff nudge never fired without any visible signal
- Adds `test_no_bash4_constructs.py` — parametrized regression grep over all shipped `.sh` files with vacuous-pass guard and `# noqa-bash4` escape hatch
- Adds two fail-open behavioral tests to the nudge hook's test module
- Folds in sibling macOS fix: `sort -V → python3` integer-tuple version compare (same file, same portability theme)

## Incidental edits

- `sort -V` fix in `update-claude-config-plugins.sh` (line 175): same macOS-portability theme as the issue but a distinct mechanism — GNU-only flag replaced with portable `python3 -c` version compare using the existing `python3` dependency.

## Test plan

- [x] `pytest claude/.claude/` — 1725 passed, 19 skipped
- [x] `test_no_bash4_constructs.py` passes on the rewritten tree (and sanity: fails if one file is reverted)
- [ ] Manual macOS verification: run `cleanup-merged-branches.sh --dry-run` and `update-claude-config-plugins.sh` under `/bin/bash` (bash 3.2.57) to confirm correct behavior; confirm nudge hook fires near the context cap

## Deferred review findings

| Hook tests: `session_id`-absent-but-fields-present case not covered | staff-sdet | Contract pinned at another layer | Existing `test_empty_json_object_input_is_fail_open` covers the absent-session_id case; `{}` produces empty string for all four fields |
| Plugin version compare: `a <= b` treats `installed == latest` as outdated (same behavior as original `sort -V`) | staff-backend-engineer | Orthogonal scope | Pre-existing behavior faithfully preserved; the inline comment "strictly higher" predates this change; semantic fix is a separate concern |
| Plugin lookup: two identical 7-line linear search loops not extracted to a helper as the plan prescribed | staff-backend-engineer | Orthogonal scope | bash 3.2 lacks namerefs making a clean helper non-trivial; portability-only scope |
| Nudge hook: `mkdir`/`touch` on fired-marker path has no timeout; can block on NFS-mounted `$HOME` | staff-platform-engineer | Orthogonal scope | Pre-existing behavior, not introduced by this change |
| Nudge hook: INPUT parse `jq` in group redirect has no timeout | staff-platform-engineer | Orthogonal scope | Pre-existing behavior, not introduced by this change; payload is structurally small |

### DIFF
--- a/.claude/plans/https-github-com-jcdendrite-claude-confi-melodic-flurry.md
+++ b/.claude/plans/https-github-com-jcdendrite-claude-confi-melodic-flurry.md
@@ -1,0 +1,109 @@
+# Plan: macOS bash 3.2 compatibility for two scripts + one hook
+
+Closes #367.
+
+## Context
+
+Stock macOS ships **bash 3.2.57** at `/bin/bash` (frozen at the last GPLv2 release in 2007). Three shell files in this repo use bash-4.0+ features — `declare -A` associative arrays and `mapfile` — so they hard-error or silently no-op for macOS stow users. CI runs on Ubuntu (bash 5) and never caught it. Because `claude/` is stowed to every user who clones this repo, the macOS audience is real, not hypothetical.
+
+The most insidious case is the hook `nudge-handoff-near-context-cap.sh`: its `2>/dev/null || true` fail-open guard swallows the `mapfile` error, so the handoff nudge **silently never fires** on macOS rather than erroring visibly.
+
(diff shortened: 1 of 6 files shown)


jcdendrite (commented):
lgtm

PR #371, opened 2026-06-03
feat(subagent-delegation): add read-then-edit decision-made test for code-writer routing

jcdendrite:
## Summary

- Narrows the stay-inline `Edit`/`Write` bullet in `subagent-delegation/SKILL.md` from a blanket rule to "where you are still deciding approach, scope, or the substantive content," with a cross-reference to the new exception
- Adds a **Read-then-edit: decision-made test** paragraph to the `### Implementation work → code-writer` section: a read-then-edit routes to `code-writer` only when (1) the change is already decided before you read and (2) reaching the target costs non-trivial context
- Narrows the frontmatter `DO NOT TRIGGER` exclusion from blanket `Edit/Write sequences` to `Edit/Write sequences where scope or content is still forming`, so the description accurately reflects the new boundary
- Ships the implementation plan in `.claude/plans/` per repo convention

## Motivation

Monthly transcript analysis found 1,532 Opus turns where Opus read 0–1 files and then edited inline, vs. 506 turns that delegated first (75/25 split), with ~33% of Opus output estimated as Sonnet-tier work. The existing stay-inline bullet overfired: it kept clearly-mechanical read-locate-then-apply edits inline even when the change was fully decided before the read.

The discriminating principle: if the change was decided *before* the read (reading only to locate a known target), it routes to `code-writer`. If the parent is still deciding scope, approach, or content *as it reads*, it stays inline. Discovery reads (mapping unfamiliar territory before deciding) continue to route to `Explore`/`general-purpose` per the existing section.

Test-case grounding is in `~/.claude/skill-test-cases/subagent-delegation-read-then-edit.md` (personal durable log, not committed — 6 annotated cases with verdicts).

## Test plan

- [x] `/skill-review` run clean (skill length gate: 200 lines exactly at limit)
- [x] `/code-review` run clean (two passes: staged diff + cumulative PR diff)
- [x] `pytest claude/.claude/` — 1,737 passed, 19 skipped
- [x] `ruff check claude/.claude/` — no issues

### DIFF
--- a/.claude/plans/subagent-delegation-read-then-edit-criteria.md
+++ b/.claude/plans/subagent-delegation-read-then-edit-criteria.md
@@ -1,0 +1,60 @@
+# Read-then-edit delegation criteria for `subagent-delegation`
+
+## Context
+
+**Goal:** add a narrow, affirmative-signal-gated rule to `subagent-delegation/SKILL.md` that tells a parent session exactly when a read-then-edit sequence routes to `code-writer` — without softening the standing default that inline edit is the safe choice.
+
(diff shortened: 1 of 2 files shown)


jcdendrite (commented):
lgtm

PR #372, opened 2026-06-03
fix(hooks): allow trailing 2>/dev/null on valid marker.sh commands

jcdendrite:
## Summary

- Agents reflexively append `2>/dev/null` to suppress stderr, but `enforce-marker-script-shape.sh` denied any suffix after the skill name, costing a one-turn recovery retry on every such invocation.
- Extended `VALID_PATTERN` and `VALID_CHAINED_MARKER_PATTERN` to accept an optional trailing ` 2>/dev/null` as a fixed ERE literal. `VALID_CHAINED_COMMIT_PATTERN` is deliberately excluded — its `[^&|;<>]*` tail already treats `>` as a forbidden boundary character, and there is no transcript evidence of friction on the `write && git commit` form (documented via comment at the pattern site).
- `marker.sh` is exit-code-honest on all failure paths (every guard exits 2 before reaching the marker write), so suppressing stderr cannot mask a failed write; the non-zero exit still propagates and downstream gates still deny.

## What changed

**`claude/.claude/hooks/enforce-marker-script-shape.sh`**
- `VALID_PATTERN` (line 75): `([[:space:]]+2>/dev/null)?` appended before trailing `[[:space:]]*$`
- `VALID_CHAINED_MARKER_PATTERN` (line 113): same trailing extension
- `VALID_CHAINED_COMMIT_PATTERN` (line 96): comment added noting the intentional `2>/dev/null` exclusion and why
- Two comment blocks updated to note the `2>/dev/null` exception

**`claude/.claude/hooks/tests/test_enforce_marker_script_shape.py`**
- Additive only — no existing tests flip
- `TestDevNullRedirectAllowed` (19 allow cases): all write/activate/deactivate targets, both `clear-stale` forms, absolute-path parity case, all four chained-marker orderings
- `TestDevNullRedirectBoundaryDenied` (16 deny cases): pins the boundary against `2>&1`, `2>/tmp/x`, `>/dev/null`, spaced `2> /dev/null`, `2>>/dev/null`, trailing args, chain operators after the suffix, newline injection, all four LHS-redirect orderings in the marker-pair pattern, and both commit-chain forms
- `test_chain_activate_write_pair_denied`: pins that `activate+write` is not a blessed chain pair

## Security review

Reviewed by `ciso-reviewer` and `staff-platform-engineer`:
- `2>/dev/null` in ERE is a fixed literal; `/` and `>` carry no metacharacter meaning in ERE context. The `$` anchor immediately after the optional group prevents anything from riding after it.
- The newline guard (`[[ "$TRIMMED" != *$'\n'* ]] &&`) runs before every grep call, fully closing the per-line-`$` interaction.
- Only the exact byte sequence ` 2>/dev/null` passes; `2>&1`, `2>/tmp/x`, `>/dev/null`, and `2> /dev/null` all continue to be denied.
- No latency, blocking, or operational footprint change.

## Test plan

- [x] `pytest claude/.claude/hooks/tests/test_enforce_marker_script_shape.py` — 105 passed
- [x] Full suite: 1770 passed, 19 skipped

### DIFF
--- a/claude/.claude/hooks/enforce-marker-script-shape.sh
+++ b/claude/.claude/hooks/enforce-marker-script-shape.sh
@@ -12,7 +12,7 @@
 # Commands that start directly with the marker.sh path (~/ or absolute) must
 # match one of the 14 single-command shapes, the marker.sh write chain to git
 # commit, or the same-skill write↔deactivate chain (plan-review and
-# ready-for-review only). No redirects, no extra args. Wrapped forms (env-var
+# ready-for-review only). No redirects (except trailing `2>/dev/null`), no extra args. Wrapped forms (env-var
 # prefix, bash wrapper, relative path, subshell) are not gated here — they
 # fast-exit at Stage 2 and are denied by
 # the permissions.allow layer, which does not list their wrapper executables.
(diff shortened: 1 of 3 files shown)


claude/.claude/hooks/enforce-marker-script-shape.sh
```diff
@@ -136,5 +136,5 @@ Valid shapes:
   ~/.claude/scripts/marker.sh clear-stale
   ~/.claude/scripts/marker.sh clear-stale --dry-run
 
-No chains (&&, ||, ;), redirects, or extra args. Env-var prefix, bash wrapper,
+No chains (&&, ||, ;), arbitrary redirects, or extra args. Env-var prefix, bash wrapper,
```
jcdendrite: Explain what additional value "arbitrary" adds here for the prose instruction?

jcdendrite: **[Claude Code]** Reverted — my prior fix went the wrong direction. Naming the exception ("redirects other than 2>/dev/null") teaches agents that redirect forms are permitted, which risks proliferating redirect usage. The deny message is shown only when a command is denied; agents that now pass with `2>/dev/null` never see it, so the message can say "redirects" without advertising the exception. Restored original wording in the latest commit on this branch.

jcdendrite (commented):
changes lgtm

PR #375, opened 2026-06-03
feat(skills): wire test-conventions and sql-query-conventions to consumers via explicit pointers

jcdendrite:
## Summary

- **Root cause fixed:** `test-conventions` and `sql-query-conventions` never fired via description-based auto-trigger in practice — trigger surface (any SELECT query, any test file) too broad to scope reliably. Repair is explicit `Read ~/.../SKILL.md` pointers at every call site that should consult the skill, matching the `error-handling` precedent (design-decisions §11). The `Read` tool expands `~` directly; no `ls` dance needed.
- **Consumers wired (write-time):** `code-writer` reads `test-conventions` when writing test code; reads `sql-query-conventions` for read-path SELECT queries
- **Consumers wired (review-time):** `staff-sdet` reads `test-conventions` before citing §N; fixes stale §4→§5 reference (§4 = isolation, §5 = naming/regression-test intent). `staff-backend-engineer` reads `sql-query-conventions` on pagination review; error-handling pointer aligned to same direct-Read pattern. `code-review` item 34 points to `sql-query-conventions` (SELECT iteration only); routing table row for test code uses `invoke` (not the weaker `consult`).
- **Settings:** both skills moved to `skillOverrides: name-only`; TRIGGER blocks retained for graceful degradation on Claude Code <v2.1.129
- **Tests:** `TestConventionSkillWiring` (6 tests) prevents silent regression of wiring; assertions pin path-form substrings and `invoke`-verb form; negative guard prevents reintroduction of stale §4 reference
- **PR comment fixes (2nd commit):** simplified all 6 `ls`→`Read` two-steps to single `Read ~/.../SKILL.md` calls; removed false "Read does not expand ~" claim from code-writer.md; clarified docs/skills.md paragraph split and "older clients" wording

## Test plan

- [x] `pytest claude/.claude/ -x -q` — passes (1818 passed, 19 skipped)
- [ ] Smoke: open a session, ask about a SELECT query, verify `sql-query-conventions` content appears in the reasoning
- [ ] Smoke: open a session, write a test, verify `test-conventions` content appears in the reasoning


docs/skills.md
```diff
@@ -14,22 +14,23 @@ Full descriptions for skills, slash commands, and project-scoped plugins in this
...
-- **`/error-handling`** — eight-principle error-handling standard: single code namespace, RFC 9457–derived envelope, developer-only message fields, and call-site anti-patterns. Model-invocable by exact name; description excluded from the listing budget via `skillOverrides: name-only`. Unlike the other four name-only skills (workflow utilities), this is a knowledge-domain skill kept name-only because the trigger surface (any try/catch or toast) is too broad to scope reliably — reached by name from the review skills and by `Read` from reviewer agents.
+- **`/error-handling`** — eight-principle error-handling standard: single code namespace, RFC 9457–derived envelope, developer-only message fields, and call-site anti-patterns. Model-invocable by exact name; description excluded from the listing budget via `skillOverrides: name-only`. Unlike the four workflow-utility name-only skills (brief, handoff, read-docx-comments, transcript-analysis), error-handling, test-conventions, and sql-query-conventions are knowledge-domain skills kept name-only because their trigger surfaces are too broad to scope reliably — reached by name from the review skills and by `Read` from reviewer agents.
```
jcdendrite: Re "Unlike the four... " this should live in a separate paragraph, otherwise it looks like it only pertains to the error-handling skill.

docs/skills.md
```diff
@@ -14,22 +14,23 @@ Full descriptions for skills, slash commands, and project-scoped plugins in this
...
 
 ## Skills available by name (no description budget cost)
 
-Five repo skills use `skillOverrides: name-only` — the model can invoke them when referenced by name in conversation, but their descriptions are excluded from the always-loaded listing budget. These skills are also slash-invocable directly. Requires Claude Code **v2.1.129+**; on older clients the override is silently ignored and these skills fall back to `on` (description loaded, weak auto-trigger since they carry no TRIGGER blocks).
+Seven repo skills use `skillOverrides: name-only` — the model can invoke them when referenced by name in conversation, but their descriptions are excluded from the always-loaded listing budget. These skills are also slash-invocable directly. Requires Claude Code **v2.1.129+**; on older clients the override is silently ignored and these skills fall back to `on` (description loaded). The four workflow utilities carry no TRIGGER blocks so auto-trigger is not a concern; the three knowledge-domain skills carry TRIGGER blocks and may fire via description match on older clients.
```
jcdendrite: What do you mean they may fire via description match on older clients? You mean older versions of Claude Code?

claude/.claude/agents/code-writer.md
```diff
@@ -45,11 +45,12 @@ language and stack:
   loop body. Batch it or hoist it out of the loop.
 - Data crossing a trust boundary (request input, external response,
   caller-supplied identifier) is validated before use.
+- When the change includes a read-path SELECT query, a PostgREST `.from()/.select()` chain, or a list-returning ORM call, Read `sql-query-conventions` before writing — resolve via `ls ~/.claude/skills/sql-query-conventions/SKILL.md` then `Read` by the printed path. Do not apply this for write-path (INSERT/UPDATE/DELETE), DDL, test fixtures, or document/KV stores.
```
jcdendrite: Re "Do not apply this for write-path (INSERT/UPDATE/DELETE), DDL, test fixtures, or document/KV stores." - what guidelines get applied for those types of tasks?

claude/.claude/agents/code-writer.md
```diff
@@ -45,11 +45,12 @@ language and stack:
...
 - New behavior ships with a test that names the specific case it guards — but
   only when the codebase already shows a test convention (a tests directory, a
   runner config, sibling test files following a pattern). When no such
   convention exists, flag the coverage gap in **Still uncertain** rather than
-  introduce test scaffolding the project does not have.
+  introduce test scaffolding the project does not have. When test code IS being written and the codebase already shows a test convention, Read the full `test-conventions` body before writing — resolve via `ls ~/.claude/skills/test-conventions/SKILL.md` then `Read` by the printed path (as in the self-review pass below). Reading the full body runs its Step 0 project-layer glob, loading any `test-conventions-*` layer for the current repo.
```
jcdendrite: I feel like there's a more elegant way to reference the SKILL pathway than stating the full path with ~ and then resolving and then referring back to the full path. 

What do other files use that refer to skills by pathname? And are you sure it's not preferable to just say read the skill and name the skill?

docs/skills.md
jcdendrite: **[Claude Code]** Fixed. Pulled the "Unlike the four workflow-utility name-only skills…" sentence out of the `/error-handling` bullet into its own paragraph between the skill list and the "Each skill lives in…" line — it now clearly describes all three knowledge-domain skills, not just error-handling. `disposition: fixed-as-requested` | addressed in the latest commit on this branch.

docs/skills.md
jcdendrite: **[Claude Code]** Yes — "older clients" was ambiguous. Updated to "older Claude Code versions (pre-v2.1.129)" for the first occurrence, and "older versions" (back-referencing the just-named version) for the second, so the version number appears once on the line rather than being repeated. `disposition: fixed-as-requested` | addressed in the latest commit on this branch.

claude/.claude/agents/code-writer.md
jcdendrite: **[Claude Code]** The excluded categories are each governed elsewhere in the same file, which is why the exclusion list exists — to prevent misapplying a read-path skill to write operations:

- **Write-path idempotency** (INSERT/UPDATE/DELETE) — covered by the bullet immediately above in the Implementation baseline: "A write a caller could retry… is idempotent."
- **Test fixtures** — covered by the very next bullet, which routes to `test-conventions/SKILL.md` when a test convention exists.
- **DDL and document/KV stores** — fall under the general CLAUDE.md §Engineering Judgment rules (ground every choice, understand the intent of existing code). `sql-query-conventions` is explicitly a *read-path* skill; applying it to DDL or schema design would be a category error.

The bullet was kept lean — adding "see X above / below" cross-references inside an already-long baseline list reduces readability more than it helps. `disposition: explain-design`

claude/.claude/agents/code-writer.md
jcdendrite: **[Claude Code]** Good instinct — the `ls`→`Read` two-step turned out to rest on a false premise. Empirical testing in both the main session and an independent subagent execution context confirmed the `Read` tool expands `~` directly. The old form (`resolve via ls … then Read by printed path`) came from a claim written in code-writer.md in PR #271 ("the Read tool does not expand ~") that had never actually been verified.

Fixed by removing the two-step everywhere it appeared and replacing with the direct `Read ~/.claude/skills/<skill>/SKILL.md` form — which is exactly what you were pointing at. The fix also removed the false premise from code-writer.md:69-72 (the self-review step that told the agent to run `ls` before every reviewer-file read). Six sites simplified in total; the design-decisions.md §14 entry updated to document the correct mechanism.

`disposition: fixed-differently` | `rationale: rather than style-only polish, empirical testing revealed the two-step was factually unnecessary — simplified all 6 sites + removed the false "Read does not expand ~" claim at :69–72` | addressed in the latest commit on this branch.

jcdendrite (commented):
lgtm

PR #378, opened 2026-06-04
fix(skills): add UUID guard and no-restore note to migration-sync Step 6

jcdendrite:
## Summary

- Adds a pre-`git rm` guard to Step 6 of `lovable-cloud-migration-sync`: extracts the filename segment between the first `_` and `.sql` for each deletion target; if it matches the UUID shape, aborts with an explanation — preventing wrong-direction deletion of a Lovable-emitted file
- Adds an explicit "do not manually restore a deleted emit" note below the `git rm` command, with a one-sentence rationale explaining the race condition it prevents
- Bumps `lovable-cloud` plugin version `3.0.0` → `3.1.0` (minor — backward-compatible new capability)

## Test plan

- [ ] Read Step 6 of `plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md` and confirm: guard prose is above the `git rm` block, the bare `git rm` command is unchanged, and the restore note is below
- [ ] Confirm no private project names, tracker IDs, or structural fingerprints in the added text

## Post-merge install

Consuming repos should refresh the plugin after merge:

```
```

### DIFF
--- a/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
+++ b/plugins/lovable-cloud/skills/lovable-cloud-migration-sync/SKILL.md
@@ -147,9 +147,22 @@ Suite-level reruns route through the subagent.
 
 ### 6. Delete originals
 
+**Guard against deleting a Lovable emit instead of a human original.** You are
+deleting human-authored slug-named files, never the UUID-named files Lovable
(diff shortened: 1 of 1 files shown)


jcdendrite (commented):
lgtm

PR #381, opened 2026-06-04
feat(skills): add premise-questioning and source-triangulation rules

jcdendrite:
## Summary

- `plan-it` Step 5 gains a **"Question the ticket's prescribed approach"** paragraph: acceptance criteria can prescribe a wrong *how*, and the compounding-patches tell is the signal to re-derive and surface a re-scope to the user rather than plan around a wrong premise.
- `verify-primary-sources` gains a **"Triangulate durable decisions"** paragraph: depth on a single authoritative source is insufficient for guidelines/standards committed to a plan, PR, or config change — cite ≥2 independent first-tier origins with verbatim quote + URL per source.

Both rules are addition-only (descriptions and TRIGGER blocks untouched); both passed `/skill-review`.

Wording tightened for DRY vs the source brief:
- Compounding-layers tell is named by its one-phrase symptom rather than restated (CLAUDE.md and plan-review Step 4 already carry the full definition).
- Harness-delegation tail dropped from the verify rule (that's `deep-research`'s job, not this skill's).
- Optional `plan-review` lens (brief §5.3) dropped — plan-review Step 4 already covers "Over-powered primitive" and "Question implementation choices, not feature scope."

## Test plan

- [x] `/skill-review` passed for `plan-it/SKILL.md` — behavioral-equivalence marker written
- [x] `/skill-review` passed for `verify-primary-sources/SKILL.md` — behavioral-equivalence marker written (re-run after F1 fix applied)
- [x] `/code-review` clean — no findings, code-review marker written
- [x] `pytest claude/.claude/` — run before merging to confirm no hook-alignment regressions
- [x] `ruff check claude/.claude/` — run before merging (no-regression check)

## Out of scope

Workstream B (make "research"/"web research" reliably engage primary-source discipline) is deferred — transcript analysis showed `verify-primary-sources` fires ~5× ever and never from "research" phrasing; it's a hook-design question, not a body-content question.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #383, opened 2026-06-04
feat(skills): rename verify-primary-sources → verify-sources, harden trigger, flip user-invocable

jcdendrite:
## Summary

- Renames `verify-primary-sources` → `verify-sources` — the confirmatory verb reads honestly standalone (most stow users lack `deep-research`; `/research` would overclaim)
- Hardens the TRIGGER description to cover "researching or doing web research on a library, API, framework, or architecture/design decision" — the previous TRIGGER only covered the reactive "acting on a claim" case and never lexically matched research phrasing, which is why it fired ~5× in 4,804 sessions
- Adds DO NOT TRIGGER clause for "full multi-source research report or synthesis" to partition cleanly from `deep-research`'s harness role (by capability, not name — `deep-research` is not stowed globally)
- Flips `user-invocable: false → true` so `/verify-sources` is directly typeable as a slash command
- Updates two consumers: `docs/skills.md` and `plan-it/SKILL.md` cross-reference

The skill body (failure-mode narrative, four numbered rules, triangulation section, fallback) is unchanged from PR #381.

## Why description-first, not a hook

The brief that preceded this PR concluded a `UserPromptSubmit` hook was needed. During planning we rejected that as premature: the transcript evidence was gathered against the *current* (unhardened) description, which never mentioned "research." The data showed "a description that doesn't mention research doesn't fire on research" — not "description-match is too weak for research." The hook stays in reserve, justified only if smoke-testing shows the hardened description still misses in practice.

## Test plan

- [ ] 1816 tests, 0 failures (`../../../.venv/bin/pytest claude/.claude/`)
- [ ] `/skill-review` passed (behavioral-equivalence marker written)
- [ ] `/code-review` passed (marker written)
- [ ] Smoke test (manual): in a fresh session, prompt "research the X library's Y API for our migration" and confirm `verify-sources` auto-engages
- [ ] Confirm `/verify-sources` is directly invocable as a slash command
- [ ] If smoke test reliably misses, file follow-up for the `UserPromptSubmit` hook (design sketch in plan Out of scope section)


jcdendrite (commented):
lgtm

PR #384, opened 2026-06-11
fix(skills): code-review dispatch table — invoke skills, not agents

jcdendrite:
## Summary

- Adds explicit in-cell imperative to the "Adds or modifies a skill, agent, instruction-file rule, or hook" row of the Change-type dispatch table: **Invoke** (via the Skill tool — these are skills, not agents; do not spawn them)
- Grammatically separates the two clauses so invoke-skills and spawn-agent have distinct bold verbs and distinct sentences
- Renames the column header `Spawn` → `Spawn / invoke` to remove the false "every entry here is an agent" premise at the source
- Normalises the four skill names to bare (no `/` prefix) to match the rest of the file (lines 161, 163, 167)

## Background

Agents running `/code-review` repeatedly matched this row and tried to **spawn** `skill-review`, `agent-review`, etc. as agents via the Agent tool, rather than **invoking** them via the Skill tool. The column header "Spawn" and 13 sibling rows listing real agent names created a pattern-completion pull that overrode the cell's intent. The fix intercepts the agent at the decision point (the cell itself) with the most specific instruction possible: naming the exact tool and negating the wrong one.

The four skill names are deliberately kept (not removed) — they are an intentional belt-and-suspenders redirect for agents already inside `/code-review` that should have invoked one of the more-specific skills directly.

## Test plan

- [x] `skill-review` run on the diff — clean
- [x] `code-review` run on the staged diff — clean
- [x] Test suite: `.venv/bin/pytest claude/.claude/` — all pass (no test parses the Change-type table's spawn tokens; skill-length gate caps code-review at 500 lines, current 373, zero net line delta)
- [x] `ruff check` — clean


jcdendrite (commented):
lgtm

PR #385, opened 2026-06-12
feat(plugin/lovable-cloud): UTC migration filename enforcement via token handshake

jcdendrite:
## Summary

- Adds `scripts/new-migration <slug>` — emits a UTC-timestamped Supabase migration basename and writes a one-shot authorization token keyed by that exact filename
- Adds `hooks/validate-migration-filename.sh` (PreToolUse) — blocks any Write to `supabase/migrations/<14digits>_*.sql` unless the token exists; gate is purely token-keyed
- Adds `hooks/consume-migration-token.sh` (PostToolUse) — removes the token after a successful Write; each generated filename can only be written once
- Bumps plugin version `3.1.0` → `3.2.0` (minor: backward-compatible addition)

**Why:** Lovable Cloud timestamps migration filenames in UTC; Claude Code's local clock may be in a different timezone. A local-clock filename inverts apply order since Supabase applies migrations in filename-lexical order. The token handshake enforces the real invariant ("generator produced this name") rather than a wall-clock proxy check, which is non-deterministic and false-blocks on long author-to-write gaps.

**Activation:** Plugin hooks are auto-discovered via `hooks/hooks.json`; no per-project `settings.json` entry required. Enable in a consuming repo with:
```
```

## Incidental edits

- `claude/.claude/tests/helpers.py`: extracted `_build_subprocess_env` from duplicated env-building blocks in `run_hook`/`run_hook_reason` (three independent copies → one helper); updated `posttooluse_input` docstring to warn callers about the `CLAUDE_PLUGIN_ROOT` env requirement
- `pyproject.toml`: added `plugins/lovable-cloud/tests` to pythonpath

## Test plan

- [x] 33 tests across 4 test files all pass: unit (generator, PreToolUse gate, PostToolUse consume) + integration (full lifecycle: generator → allow → consume → deny)
- [x] Global test suite (1826 tests) passes with no regressions from `helpers.py` changes
- [x] `ruff check` clean on plugin and modified helpers
- [x] `/code-review` complete (staff-sdet, staff-backend-engineer, staff-platform-engineer, ciso-reviewer, skill-review, claude-hook-review, plugin-semver)

## Review findings addressed

| Finding | Source | Disposition |
|---|---|---|
| `test_same_second_same_slug_double_gen_is_idempotent` didn't test idempotency (admitted in test comment) | staff-sdet BLOCKER | FIXED — fake `date` binary freezes time; test now asserts `filename1 == filename2` and exactly one token |
| UUID near-miss tests missing (wrong segment length, mixed case) | staff-sdet CONCERN | FIXED — added `test_wrong_segment_length_uuid_denies`, `test_mixed_case_uuid_denies` |
| `_lib.sh` sourcing failure path untested | staff-sdet CONCERN | FIXED — added `test_missing_lib_sh_denies` with empty CLAUDE_PLUGIN_ROOT |
| `posttooluse_input` caller contract undocumented | staff-backend CONCERN | FIXED — docstring now warns about CLAUDE_PLUGIN_ROOT requirement |
| Duplicated env-building in `run_hook`/`run_hook_reason` | staff-backend CONCERN | FIXED — extracted `_build_subprocess_env` helper |
| UUID exemption was a bypass hole — model-controlled segment, no legitimate Claude Write of UUID files | PR review comment #3408533817 | FIXED — exemption removed; gate is now purely token-keyed; `MIGRATION_TOKEN_DIR` empty-guard added matching consume hook |
| Standalone migration-author skill wrong delivery channel — `code-writer` has no Skill tool | PR review comment #3408539030 | FIXED — skill deleted; agent-agnostic PreToolUse deny message is the encoding |
| `_plugin_extra_env(home)` dead parameter in integration test | cumulative review | FIXED — removed unused parameter |
| `test_no_token_non_uuid_migration_denies` name implied UUID exemption still exists | cumulative review | FIXED — renamed to `test_no_token_migration_denies` |
| Plugin descriptions didn't surface new hook behavior | cumulative review | FIXED — `plugin.json` + `marketplace.json` descriptions updated |

## Deferred review findings

|---|---|---|---|
| Bare `jq` post-validation call at `validate-migration-filename.sh:35` bypasses `_lib_jq` timeout | staff-platform | Contract pinned at another layer | `claude-hook-review` §7 explicitly permits direct jq for post-validation extracts on already-proven-parseable input |
| Missing `set -euo pipefail` in validate hook | staff-platform | Orthogonal scope | Pre-existing gap in plugin hooks; gate still fails closed on unset-variable paths; not introduced by this diff |
| `CLAUDE_PLUGIN_ROOT` path in deny message info-leak | ciso | Gold-plating beyond declared user surface | Full path is functionally required for Claude to invoke the generator; model already has Bash access to discover the path |
| `_plugin_env` dead-code re-implementation in `test_validate_migration_filename.py` | staff-sdet | Orthogonal scope | Pre-existing; `TestMalformedInput` still uses it and it's not incorrect, just redundant with `helpers._build_subprocess_env` |
| `token-path.sh` should document network-backed HOME filesystem risk | staff-platform | Orthogonal scope | Pre-existing design choice; not introduced by this diff; a one-line comment addition can follow as a separate cleanup |
| Retry guidance ("re-generate if Write is retried") not explicit in deny message | skill-review | Orthogonal scope | Deliberate reactive-only design confirmed by engineer; deny message's "use the generator" redirect adequately covers the retry case |


plugins/lovable-cloud/hooks/validate-migration-filename.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+# hook-class: gate
+# PreToolUse gate: block Write calls to supabase/migrations/<14digits>_*.sql
+# unless a one-shot token was written by the new-migration generator, or the
+# filename is a Lovable-emitted UUID file. Fails closed on malformed input.
```
jcdendrite: Under what circumstances would we pass a Lovable-emitted UUID file to this script? Shouldn't this script be used to validate only the filename that was generated? It seems that skipping a file can open up to bugs. Am I missing something?

plugins/lovable-cloud/skills/lovable-cloud-migration-author/SKILL.md
```diff
@@ -0,0 +1,54 @@
+---
+name: lovable-cloud-migration-author
+description: >
+  Supabase migration file authoring in a Lovable Cloud project — UTC timestamp
+  enforcement via the new-migration generator.
+  TRIGGER when: about to author a new Supabase migration file (Write to
+  supabase/migrations/*.sql with a new timestamp).
+  DO NOT TRIGGER when: editing an existing migration; not a Supabase project.
+user-invocable: false
+---
```
jcdendrite: Hmm I'm not sure about adding another skill here for two reasons:
1. Skill context - this tool claude-config already comes with a lot of skills and this particular one is hyper-specific. We don't want to hit the max skill cap that would force us to increase skill c ontext.
2. The skill most likely will not be triggered when we want. For users of the claude-config toolkit, most code writing is delegated to a the code-writer agent, and I'm not sure if subagents can use skills themselves.

Is there more elegant way we can encode these instructions?

plugins/lovable-cloud/hooks/validate-migration-filename.sh
jcdendrite: **[Claude Code]** Fixed. Removed the UUID exemption entirely — the gate is now purely token-keyed. The rationale: the post-prefix segment is model-controlled, so the exemption let any `<timestamp>_<lowercase-uuid>.sql` Write bypass enforcement even with a non-UTC timestamp. There is no legitimate scenario where Claude's Write tool produces a UUID-named migration (`lovable-cloud-migration-sync` confirms Lovable's own bot commits them directly; the sync skill explicitly forbids Claude from hand-restoring them), so the exemption closes nothing real and opens a bypass. Also added a `MIGRATION_TOKEN_DIR` empty-guard to the validate hook (matching the guard already in the consume hook) and two new tests: one confirming UUID-named files deny without a token, one confirming the token-keyed allow path is shape-agnostic. `disposition: fixed-as-requested`

plugins/lovable-cloud/skills/lovable-cloud-migration-author/SKILL.md
jcdendrite: **[Claude Code]** Fixed — deleted the skill. The more elegant encoding is the existing PreToolUse deny message. Two reasons: (1) `code-writer`, the agent the Model Routing rule mandates for migration authoring, has no Skill tool (`tools: Read, Edit, Write, Bash, Grep, Glob`) — subagents reach skills only via explicit `Read`, so the auto-trigger `user-invocable: false` skill never fired where the work actually happens. (2) The deny message is agent-agnostic: it reaches whichever context (main session or subagent) issued the Write, which is exactly the reactive channel the skill could not cover. This matches the established repo pattern (`block-gh-pr-merge.sh`, `deny-env-reads.sh`). The skill also spent always-loaded skill-description budget the repo actively conserves. `disposition: fixed-as-requested`

jcdendrite (commented):
changes lgtm

PR #386, opened 2026-06-13
feat(skills): add guardrail for regex-parsing runtime output in test assertions

jcdendrite:
## Summary

- Adds a new `§9` row to `test-conventions/SKILL.md` for the anti-pattern of regex-parsing specific field values from runtime/structured output (log lines, JSON, XML) instead of using a parser library or the production parse/validate function
- Adds a parity row to `test-evaluation/SKILL.md` §4 (mirrors the test-conventions convention for evaluator path)
- Adds two new code-review hygiene items:
  - **9f** — brittle structural test assertion (runtime-output regex on field values, with format-envelope exemption)
  - **9g** — source-scanning test assertion (makes the existing §9 source-scan row first-class at review time with a named owner)
- Updates the Item ownership table: 9f → `staff-sdet` (primary) + `staff-backend-engineer` (co-owner for backend API response output); 9g → `staff-sdet`
- Records grounding citations in `test-conventions/REFERENCES.md` (Meszaros xUnit Test Patterns "Fragile Test," SWE@Google "Don't Put Logic in Tests")

## Scope note

The source-scanning sibling (reading source text as a string and asserting substrings) was already covered by the existing `§9`/`§4` table rows. This PR does not touch that rule text — it only promotes source-scanning to a named code-review item (9g) with an explicit owner. Item 9f covers the distinct runtime-output case.

Format-envelope assertions — those checking only that output *is* valid JSON or matches a line-format shape, not asserting on specific field values within it — are explicitly exempted in all three files.

## Motivation

Cross-project transcript investigation confirmed this as a real recurring pattern: sessions across multiple projects added tests that applied `re.match`/`re.findall`/string-matching against raw log lines, XML fragments, or JSON text rather than using a parser library or calling the production parser. The failure mode (test passes when production parsing is broken; breaks on benign format changes) matches Meszaros "Fragile Test" and the SWE@Google "Don't Put Logic in Tests" principle.

~99% of JS/TS signal was the already-covered source-scanning shape; the runtime-output regex pattern was clearest in Python tooling tests.

## Test plan

- [x] `pytest claude/.claude/` — PASS
- [x] `ruff check claude/.claude/` — PASS
- [x] `skill-review` on all three modified SKILL.md files — clean
- [x] `code-review` — all staff-sdet findings addressed
- [x] `test-conventions` at 199 lines (under 200-line cap, was 201 before)
- [x] `code-review` at 379 lines (under 500-line ceiling)

## Incidental edits

`test-conventions/SKILL.md`: removed the `"Avoid these when writing new tests:"` lead-in sentence and the blank line before the §9 table (reclaimed 2 lines to bring the file from 201 → 199 and satisfy the `check-skill-length.sh` gate). The section heading `## 9. Common authoring mistakes` carries the same instruction; behavioral-equivalence audit confirmed no content loss.

### DIFF
--- a/.claude/plans/guardrail-regex-test-assertions.md
+++ b/.claude/plans/guardrail-regex-test-assertions.md
@@ -1,0 +1,72 @@
+# Guardrail: brittle regex/string-matching test assertions
+
+## Context
+
+**Goal:** add a claude-config skill-layer guardrail that stops sessions from writing brittle regex/string-matching test assertions over **runtime/structured output** when a parser library or the production parse/validate function is the correct tool — grounded in a cross-project transcript investigation and primary-source verification.
+
+The user observed, across all projects, sessions adding brittle regex-based tests where a library or business-logic method was preferable. A confirmatory read-only pass over the transcript corpus (**4,742 JSONL transcripts across 53 project dirs**, via `~/.claude/scripts/transcript-analysis.py`'s storage layout) confirms regex-in-test is real and recurring — but reframes the scope:
+
+- **~99% of the JS/TS signal and ~65% of the Python signal is the *source-scanning sibling*** — reading a file's source/DDL text as a string and regex-matching substrings — which `test-conventions` §9 **already covers** (line 199), and which `test-evaluation` §4 mirrors (line 69).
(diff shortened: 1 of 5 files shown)


jcdendrite (commented):
lgtm

(Similar instructions are added in multiple places, but the reinforcement seems to be necessary given the strong tendency of the LLM to write regex-type tests).

PR #387, opened 2026-06-14
feat(agents): promote skill reads to preconditions in code-writer; add brittle-assertion angle to staff-sdet

jcdendrite:
## Summary

- Restructures `code-writer.md`'s two skill-read instructions from buried conditional bullets into bold **Before [condition]:** preconditions at the top of the Implementation baseline, so the model encounters them before writing rather than mid-way through a long bullet scan
- Makes the `test-conventions` read unconditional on any test file edit (previously gated on both "test code IS being written" AND "the codebase already shows a test convention" — the test-shipping rule retains its condition, but the read does not)
- Applies the same structural fix to the `sql-query-conventions` read (the same conditional-clause weakness in the sibling arm)
- Adds a **Brittle structural assertions** review angle to `staff-sdet.md` covering regex/string-matching over runtime/structured output and source-scanning assertions (items 9f/9g from #386), which was the angle missing from `code-writer`'s in-context self-review pass

## Motivation

Transcript-analysis of ~1,400 sessions confirmed the suspected regression:

| Era | code-writer dispatches that read test-conventions | regex-in-tests rate |
|---|---|---|
| Before code-writer (<2026-05-19) | n/a | 1.8% |
| code-writer, no pointer (2026-05-19 → 06-02) | **0%** | 5.2% |
| code-writer + explicit pointer via #375 (≥2026-06-03) | **39%** | 3.1% |

The wiring in #375 improved the read rate from 0% → 39% and brought the regex-in-tests rate down, but 61% of test-writing dispatches still skip the skill read. Making the read a prominent, separate precondition rather than a conditional clause should push the rate further (prominence moved it 0% → 39%; a cleaner, earlier-positioned instruction should move it again).

`staff-sdet` receiving the new angle closes `code-writer`'s self-review path: #386 added 9f/9g to `code-review/SKILL.md` and named `staff-sdet` as owner in the ownership table, but did not edit `staff-sdet`'s body. Since `code-writer` reads `staff-*` reviewer files (not `code-review/SKILL.md`) during self-review, the angle was absent from its in-context pass.

## Verification

- `agent-review` run on both agent files — clean (one noted behavior change: test-conventions read is now unconditional, confirmed intentional per plan)
- `/code-review` — no findings
- `pytest claude/.claude/` — 1826 passed, 22 skipped
- `ruff check claude/.claude/` — clean
- Empirical follow-up (deferred): re-run `transcript-analysis` after these land and accrue usage to confirm read rate rises above 39% and regex-in-tests rate falls; if read rate stays low, escalate to a PreToolUse hook gate

### DIFF
--- a/.claude/plans/code-writer-precondition-reads.md
+++ b/.claude/plans/code-writer-precondition-reads.md
@@ -1,0 +1,69 @@
+# Plan: close code-writer's skill-loading reliability gap for test-authoring rules
+
+## Context
+
+**Goal:** make `code-writer` reliably pick up test-authoring conventions (specifically the "don't regex-parse SQL / runtime output in assertions" family) at the two points where it can act on them — *while writing* and *during its own self-review* — instead of leaving the catch to a parent `/code-review` round-trip.
+
+**Why now.** The user instituted a routing rule (CLAUDE.md §Model Routing) sending delegated code-writing to `code-writer` (introduced 2026-05-19, #271), and suspected a regression: skills that guide code-writing might not reach `code-writer` because it can't invoke the `Skill` tool. Investigation refined this:
+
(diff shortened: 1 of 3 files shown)


jcdendrite (commented):
lgtm

PR #389, opened 2026-06-17
feat(settings): flip agent-review and skill-review to name-only

jcdendrite:
## Summary

- Flip `agent-review` and `skill-review` to `skillOverrides: name-only` in the globally-stowed `claude/.claude/settings.json`
- Both skills are always reached by name (dispatched from `/code-review` SKILL.md:241 and the `require-skill-review` hook), never by description auto-trigger — the flip is safe and cannot break dispatch
- `agent-review` is globally stowed → its description was loading in every stow-consumer project's listing; flipping it reduces listing pressure immediately across all consumers
- `skill-review` is in the claude-config-scoped `skill-management` plugin (not in stow-consumer listings today) but is flipped now as forward-looking budget hygiene before planned plugin rollout to other repos
- TRIGGER blocks kept for graceful degradation on pre-v2.1.129 Claude Code clients

## Documentation

- **`docs/skills.md`**: add both skills to the name-only table (Seven→Nine), extend category paragraph to name the third category (dispatcher-reached reviewer skills)
- **`docs/design-decisions.md`**: 2-line amendment appended to §15, extending the dispatch-wired precedent to both skills with the stow-consumer budget motivation
- **`.claude/plans/`**: evaluation plan documenting the corrected lever analysis (prior session plan had the lever backwards — it framed this as "reclaim claude-config budget" when claude-config has no overflow; the real lever is trimming globally-stowed listing footprint to relieve stow consumers)

## Test plan

- [ ] `/skills` in a fresh claude-config session: confirm `agent-review` and `skill-review` descriptions no longer appear in the listing
- [ ] `/code-review` on any diff with agent or SKILL.md files: confirm both reviewer skills are still dispatched by name
- [ ] `/doctor` in a fresh session on any stow-consumer project: confirm the dropped-description count fell (this is the success metric)
- [ ] For `skill-review` key specifically: confirm the bare `"skill-review"` key actually takes effect in a session where `skill-management` is enabled (if it silently no-ops, switch to `"skill-management:skill-review"`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #390, opened 2026-06-17
fix(settings): qualify skill-review override key as skill-management:skill-review

jcdendrite:
## Summary

- Renames `"skill-review": "name-only"` → `"skill-management:skill-review": "name-only"` in `skillOverrides` — bare plugin-skill keys silently no-op in the harness; the fix that was intended by #389 only half-landed (`agent-review` worked; `skill-review` did not)
- Adds `test_plugin_scoped_name_only_key_is_qualified` to enforce the bare-key-must-be-stowed contract, catching this class of bug before merge
- Updates `_skill_file()`, `_model_invokable_skills()`, and `test_skill_overrides_documented_in_docs_skills_md` to understand the `plugin:skill` qualified key form
- Amends `docs/design-decisions.md` §15 with one sentence explaining that plugin-scoped name-only skills require the qualified key form

## Test plan

- [x] `pytest claude/.claude/` — 165 passed, 2 skipped (the additional skip vs. #389's 166/1 is the `skill-management:skill-review` entry now correctly skipping rather than vacuously passing `test_plugin_scoped_name_only_key_is_qualified`)
- [ ] After merge: fresh session `/skills` — confirm `skill-review` description absent from listing (name only), matching `agent-review`
- [ ] After merge: `/doctor` — confirm dropped-description warning for `git-state-safety` has cleared

## Incidental edits

- `_model_invokable_skills()` plugin loop: added inline comment marking the latent multi-plugin same-named-skill boundary (non-blocking, no test change needed at current scale)

## Deferred review findings

|---|---|---|---|
| `_skill_file()` qualified branch returns path with no existence check — typo in plugin name gives opaque `FileNotFoundError` instead of a helpful assertion | staff-sdet #1 | 3 — Gold-plating | `test_name_only_skill_has_skill_file` already catches misspelled plugin keys via `.exists()` assertion; diagnostic quality only |
| `_model_invokable_skills()` fetches description with bare `skill_dir.name` in plugin loop — ambiguous if two plugins ship a skill with the same bare name | staff-sdet #2 | 5 — Edge case below current scale | Only one plugin in repo; boundary marked with inline comment |


jcdendrite (commented):
lgtm

PR #391, opened 2026-06-17
feat(settings): flip loop and simplify from off to name-only

jcdendrite:
## Summary

- Move `loop` and `simplify` in `skillOverrides` from `"off"` to `"name-only"` — restores slash-invocability and model-invocability by exact name at zero description-budget cost
- Add `BUILTIN_NAME_ONLY_SKILLS: set[str] = {"loop", "simplify"}` allowlist in `test_skills.py` to exempt bundled skills (no repo SKILL.md) from the three repo-SKILL.md-dependent name-only contracts; add set-equality drift guard `test_builtin_name_only_allowlist_matches_settings`
- Update `docs/skills.md`: corrects the bundled-disabled count (twelve → ten), adds a companion table for the two bundled name-only skills, adds a one-line plugin caveat on `skillOverrides` scope
- Add §17 to `docs/design-decisions.md` recording the off→name-only rationale and the verified finding that `skillOverrides` does not apply to plugin skills

## Test plan

- [x] `../../../.venv/bin/pytest claude/.claude/` — 153 passed, 5 skipped (4 BUILTIN skips for loop/simplify + 1 pre-existing empty-COMMAND_SKILLS skip); no FileNotFoundError cascade
- [x] `../../../.venv/bin/ruff check claude/.claude/` — clean
- [ ] Manual: `/loop` and `/simplify` appear in the `/` menu and are invocable
- [ ] Manual: `/doctor` shows no description-budget regression (descriptions for loop/simplify remain excluded)


jcdendrite (commented):
lgtm

PR #392, opened 2026-06-17
fix(skill-review): strip TRIGGER blocks; revert dead skillOverrides entry

jcdendrite:
## Summary

Plugin skills are categorically exempt from `skillOverrides` per the official Claude Code docs ("Plugin skills are not affected by `skillOverrides`"). Neither the bare key (#389) nor the qualified key (#390) could make `skill-review` name-only — both were silent no-ops. This PR corrects the record and instead minimizes the always-loaded description.

**What changed:**

- `plugins/skill-management/skills/skill-review/SKILL.md` — Description trimmed from 8 lines to 3. TRIGGER/DO-NOT-TRIGGER blocks stripped: the skill is always dispatched by name (from `/code-review` and the `require-skill-review` hook), never by description auto-trigger, so the blocks were always-loaded permanent cost with zero routing value. `user-invocable: false` kept.
- `plugins/skill-management/.claude-plugin/plugin.json` — Version bumped 2.3.0 → 2.3.1 (patch — description trim, backward-compatible).
- `claude/.claude/settings.json` — Dead `"skill-management:skill-review": "name-only"` entry removed.
- `claude/.claude/skills/tests/test_skills.py` — #390's qualified-key infra reverted (`_skill_file` qualified branch + `test_plugin_scoped_name_only_key_is_qualified` deleted). New `_NAME_DISPATCHED_NO_TRIGGER` exemption set added for name-dispatched skills that intentionally carry no TRIGGER blocks, with maintenance protocol comment and inverted-assertion test class (`TestNameDispatchedNoTriggerContracts`) enforcing the invariant. `agent-review` contracts preserved; `skill-review` entries removed from hardcoded parametrize lists.
- `docs/design-decisions.md` §15 — Corrected false claim. `agent-review` (user-scope) works with name-only. `skill-review` (plugin-scope) is categorically exempt; description minimized instead.
- `docs/skills.md` — Count updated (nine → eight name-only skills). `skill-review` removed from "available by name" table; explanatory note added documenting the plugin-scope exemption.

**Asymmetry with `agent-review` (deliberate):** `agent-review` is user-scope — its name-only override works and TRIGGER blocks are kept as graceful-degradation insurance on pre-v2.1.129 clients. `skill-review` is plugin-scope — the override never takes effect, so TRIGGER blocks are stripped (they served no purpose) and the description is minimized.

**Plugin install command** (for consuming repos after merge):
```
```

## Test plan
- [x] `pytest claude/.claude/skills/` — 179 passed, 1 skipped
- [x] `TestNameDispatchedNoTriggerContracts.test_name_dispatched_skill_has_no_trigger_block` — confirms `skill-review` frontmatter has no TRIGGER blocks (new enforcement test)
- [ ] Post-merge: live skill listing should show `skill-management:skill-review` with the trimmed 3-line description (reduced from 8 lines)

## Deferred review findings

|---|---|---|---|
| skill-review SKILL.md missing TRIGGER/DO-NOT-TRIGGER blocks (skill-review items 1 + 11 N rows) | skill-review | Criterion 3 — Gold-plating beyond declared user surface | Removing these blocks IS the PR's declared intent; restoring them would undo the user surface. Exemption enforced by `_NAME_DISPATCHED_NO_TRIGGER`; "not auto-triggered by description" documents it. |


jcdendrite (commented):
lgtm

